PR #11, opened 2019-11-08
Dynamically access context and activity from registrar

Sese-Schneider:
Previously `PurchasesFlutterPlugin.java` accessed context and activity once on initialisation and saved it to a variable.
However the activity could be attached to the engine after the registrar has been called.
Therefore it is necessary that this class gets the activity and context dynamically on call from the registrar.

@vegaro I hope you can publish this change with the update from [#89](https://github.com/RevenueCat/purchases-android/issues/89) from your repo `purchases-android`


android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
```diff
@@ -34,7 +34,6 @@
 import io.flutter.view.FlutterNativeView;
 import kotlin.UninitializedPropertyAccessException;
 
-import static com.revenuecat.purchases_flutter.Mappers.map;
```
vegaro: oopsies, thanks for cleaning that

vegaro:
I can actually release this as a small release before the other one comes out :)

Sese-Schneider:
Thanks!

PR #13, opened 2019-11-26
Update to 3.0.0

vegaro:


IchordeDionysos:
This would also fix #14.

@vegaro Any info when this might be released? I understand that you can't commit to anything.

PR #45, opened 2020-05-12
Exposes finishTransactions to flutter code

loregr:
I've implemented a method to be able to use `setFinishTransactions` from Dart.

### DIFF
--- a/ios/Classes/PurchasesFlutterPlugin.m
+++ b/ios/Classes/PurchasesFlutterPlugin.m
@@ -123,7 +123,7 @@ NSString *RNPurchasesPurchaserInfoUpdatedEvent = @"Purchases-PurchaserInfoUpdate
 }
 
 - (void)setFinishTransactions:(BOOL)finishTransactions
-                             result:(FlutterResult)result
+                       result:(FlutterResult)result
 {
     [RCCommonFunctionality setFinishTransactions:finishTransactions];
     result(nil);
--- a/ios/Classes/PurchasesFlutterPlugin.m
+++ b/ios/Classes/PurchasesFlutterPlugin.m
@@ -46,7 +46,9 @@ NSString *RNPurchasesPurchaserInfoUpdatedEvent = @"Purchases-PurchaserInfoUpdate
         [self setupPurchases:apiKey appUserID:appUserID observerMode:observerMode result:result];
     } else if ([@"setAllowSharingStoreAccount" isEqualToString:call.method]) {
         [self setAllowSharingStoreAccount:[arguments[@"allowSharing"] boolValue] result:result];
-    } else if ([@"addAttributionData" isEqualToString:call.method]) {
+    } else if ([@"setFinishTransactions" isEqualToString:call.method]) {
+              [self setFinishTransactions:[arguments[@"finishTransactions"] boolValue] result:result];
+          }else if ([@"addAttributionData" isEqualToString:call.method]) {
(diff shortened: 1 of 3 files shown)


vegaro:
Thanks for opening this @loregr 

I have one question though. Are you trying to set the SDK in observer mode? Why do you need to set finishTransactions?

There's already a parameter in the setup call https://github.com/RevenueCat/purchases-flutter/blob/master/lib/purchases_flutter.dart#L34 that will let you set the SDK in observer mode in case you already have other iap system in place and you don't want our SDK to finish/consume any transaction. SetFinishTransactions is kinda the legacy way of setting the SDK in observer mode.

loregr:
@vegaro thanks for your feedback. In my understanding (and I may be wrong) observer mode is to be used when I want only to get access to the advanced charting, webhooks, and attribution while I manage purchases and restores in a native way (or without going through revenueCat SDK). 
In may case, I still handle the whole in-app purchase process using RevenueCat SDK, the only thing that I needed (and it was missing) was the possibility to don't consume the purchase because I want the user to be able to restore the purchase (no-subscription) if he uninstalls and reinstalls the app, or simply changes his device. So setting `finishTransactions` to `false` was exactly what I needed. Also, it looks like that if I want to use observer mode I must pass an `appUserId`, otherwise the value will be ignored, by setting `finishTransactions` to `false` instead, I don't need any additional info.

vegaro:
I see, that makes sense. Can't you get that functionality by using the information inside `PurchaserInfo`? Even if that transaction has been consumed it should appear after a restore in the `allPurchasedProductIdentifiers` or in the `entitlements` if you're using the entitlements system.

(sorry it took so long to reply 🤦)

ios/Classes/PurchasesFlutterPlugin.m
```diff
@@ -46,7 +46,9 @@ - (void)handleMethodCall:(FlutterMethodCall *)call result:(FlutterResult)result
         [self setupPurchases:apiKey appUserID:appUserID observerMode:observerMode result:result];
     } else if ([@"setAllowSharingStoreAccount" isEqualToString:call.method]) {
         [self setAllowSharingStoreAccount:[arguments[@"allowSharing"] boolValue] result:result];
-    } else if ([@"addAttributionData" isEqualToString:call.method]) {
+    } else if ([@"setFinishTransactions" isEqualToString:call.method]) {
+              [self setFinishTransactions:[arguments[@"finishTransactions"] boolValue] result:result];
+          }else if ([@"addAttributionData" isEqualToString:call.method]) {
```
aboedo, line 51: Indentation is off here:

```suggestion
    } else if ([@"setFinishTransactions" isEqualToString:call.method]) {
              [self setFinishTransactions:[arguments[@"finishTransactions"] boolValue] result:result];
    } else if ([@"addAttributionData" isEqualToString:call.method]) {
```

ios/Classes/PurchasesFlutterPlugin.m
```diff
@@ -120,6 +122,13 @@ - (void)setAllowSharingStoreAccount:(BOOL)allowSharingStoreAccount
     result(nil);
 }
 
+- (void)setFinishTransactions:(BOOL)finishTransactions
+                             result:(FlutterResult)result
```
aboedo, line 126: indentation is off here as well: 
```suggestion
- (void)setFinishTransactions:(BOOL)finishTransactions
                       result:(FlutterResult)result
```

PR #65, opened 2020-06-15
Use includeBuilds for better local development

vegaro:
Inspired by https://proandroiddev.com/saying-goodbye-to-snapshots-with-gradles-composite-builds-bc98751392f6 and https://galex.co.il/2017/09/12/The_Best_of_Two_worlds_with_Gradle_Composite_Builds.html.

This adds two new gradle tasks `enableLocalBuild` and `disableLocalBuild`. By running `enableLocalBuild` gradle will use the local copy of `purchases-hybrid-common` instead of the one dowloaded from maven.

Depends on github.com/RevenueCat/purchases-hybrid-common/pull/44

### DIFF
--- a/example/android/settings.gradle
+++ b/example/android/settings.gradle
@@ -13,3 +13,9 @@ plugins.each { name, path ->
     include ":$name"
     project(":$name").projectDir = pluginDirectory
 }
+
+// Run enableLocalBuild task to enable building purchases-hybrid-common from your local copy
+if (file(".composite-enable").exists()) {
+    includeBuild ('../../../../purchases-hybrid-common/android')
+}
+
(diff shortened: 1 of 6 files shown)


example/android/settings.gradle
```diff
@@ -13,3 +13,9 @@ plugins.each { name, path ->
     include ":$name"
     project(":$name").projectDir = pluginDirectory
 }
+
+// Run enableLocalBuild task to enable building purchases-hybrid-common from your local copy
+if (file(".composite-enable").exists()) {
+    includeBuild ('../../../../purchases-hybrid-common/android')
```
aboedo, line 19: is this line assuming the local path of purchases-hybrid-common?

vegaro: yes, what could we do to make this easier? maybe an environment variable?

aboedo: env variable would be the absolute ideal

vegaro: let me try that

vegaro:
@aboedo I changed so the task now accepts a parameter with the path:

`./gradlew enableLocalBuild -PcommonPath="$HOME/Development/repos/purchases-hybrid-common/android"`

aboedo:
awesome!!

Issue #51, opened 2020-05-19
FatalException *** -[__NSCFConstantString stringByAppendingString:]: nil argument -[NSPlaceholderString initWithCharactersNoCopy:length:freeWhenDone:]


owenkealey:
Hi,

I have an app with about 800 users and the library works for 99% of people, but I saw this in Crashlytics today.
<img width="951" alt="Screen Shot 2020-05-19 at 7 34 56 PM" src="https://user-images.githubusercontent.com/26207512/82388705-3a7e6800-9a08-11ea-8bfe-a2e8ba37d473.png">

Any thoughts? I can provide more info if needed.

aboedo:
Is your app deleting `NSUserDefaults`? If so, that would explain the crash.
If that is the case, there are a couple of workarounds, like only deleting the keys your app uses, or configuring RevenueCat with a separate instance of NSUserDefaults. Let me know if this is your case and I can elaborate more.

owenkealey:
I have no idea honestly, I just setup the purchases library as was
instructed in the Flutter documentation for Revenue Cat.

aboedo:
I see. My thinking is that there might be a plugin or something (or the app itself) that's deleting user defaults. 
I'm working on https://github.com/RevenueCat/purchases-flutter/issues/49, which will allow for a quick workaround for this issue on flutter. I'll keep you posted. 
In the meantime, are you using any plugins like [shared_preferences](https://flutter.dev/docs/cookbook/persistence/key-value) that might be causing this?

owenkealey:
I am using flutter_secure_storage

aboedo:
thanks for the info. That one _should_ be safe, since it internally uses the keychain for storage (from their docs). 
I'm not sure what might be causing NSUserDefaults getting cleared in your case, but I'll let you know as soon as I get the workaround out. 
in #49 there's another workaround that might help, although that one involves a bit more work on the dev side

Issue #49, opened 2020-05-16
Expose `userDefaults` in `Purchases.configure` just like iOS repo


om-ha:
### Problem
The original problem that requires this change, is that RevenueCat uses NSUserDefaults as a data layer. The actual problem is using the default instance and not a separate standalone instance specially instantiated for RevenueCat.

This means when NSUserDefaults is reset/reloaded/cleared, then RevenueCat will lose track of all persisted data.

This might happen automatically for some reason, abnormalities with NSUserDefaults shared instance that puts RevenueCat in an undefined state.

### Fix
With this introduction for the problem, the solution is simply taking in an optional string ID as a parameter within `Purchases.Configure` that would create the standalone NSUserDefaults instance automatically for us. Like [this solution in an issue](https://github.com/RevenueCat/purchases-ios/issues/151#issuecomment-583076972t) from iOS repo. Or at least just make it so that RevenueCat uses a hardcoded NSUserDefaults domain ID.

om-ha:
For reference, I actually made a platform channel to fix this. In case if anybody is interested:

Inside `AppDelegate.swift` -> `didFinishLaunchingWithOptions` -> create a new `FlutterMethodChannel` (for our example "myapp_platform_channel") and inside `setMethodCallHandler` callback parameter `FlutterMethodCallHandler` add the following:

``` swift
// MARK: `setupRevenueCatPurchasesCupertino` method
if call.method == "setupRevenueCatPurchasesCupertino" {
    let argumentsDictionary = call.arguments as? Dictionary<String, Any?>
    let apiKey = argumentsDictionary?["apiKey"] as? String
    self.setupRevenueCatPurchasesCupertino(result: result, apiKey: apiKey)
    return
}
```

Then within `AppDelegate` class, create this private instance method:
``` swift
// MARK:- RevenueCat Method Channel
private func setupRevenueCatPurchasesCupertino(result: FlutterResult, apiKey: String?)  {
    let userDefaultsDomain = "com.yourdomain.yourapp.revenuecat"
    let revenueCatAPIKey = apiKey ?? "hard_code_your_api_key_if_you_dare"
    
    // create UserDefaults instance/domain for RevenueCat so our app wouldn't overwrite it
    guard let revenueCatUserDefaults = UserDefaults.init(suiteName: userDefaultsDomain) else {
        let errorMessage = "RevenueCat Error: can't create user defaults for RevenueCat domain"
        print(errorMessage)
        result(false)
        return
    }
[...]

PR #70, opened 2020-06-18
User defaults Suite Name option in setup

aboedo:
Added `userDefaultsSuiteName` option to setup. 

Addresses #51 and #49 .

### DIFF
--- a/android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
+++ b/android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
@@ -71,6 +71,7 @@ public class PurchasesFlutterPlugin implements MethodCallHandler {
                 String apiKey = call.argument("apiKey");
                 String appUserId = call.argument("appUserId");
                 Boolean observerMode = call.argument("observerMode");
+                String userDefaultsSuiteName = call.argument("userDefaultsSuiteName"); // iOS-only, unused.
                 setupPurchases(apiKey, appUserId, observerMode, result);
                 break;
             case "setFinishTransactions":
--- a/ios/Classes/PurchasesFlutterPlugin.m
+++ b/ios/Classes/PurchasesFlutterPlugin.m
@@ -41,7 +41,8 @@ NSString *RNPurchasesPurchaserInfoUpdatedEvent = @"Purchases-PurchaserInfoUpdate
         NSString *apiKey = arguments[@"apiKey"];
         NSString *appUserID = arguments[@"appUserId"];
         BOOL observerMode = [arguments[@"observerMode"] boolValue];
-        [self setupPurchases:apiKey appUserID:appUserID observerMode:observerMode result:result];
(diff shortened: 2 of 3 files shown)


lib/purchases_flutter.dart
```diff
@@ -32,17 +32,19 @@ class Purchases {
...
+  /// 
+  /// [userDefaultsSuiteName] iOS-only, will be ignored for Android. 
+  /// Set this if you would like the RevenueCat SDK to store its preferences in a different
+  /// NSUserDefaults suite, otherwise it will use standardUserDefaults.
+  /// Default is null, which will make the SDK use standardUserDefaults.
   static Future<void> setup(String apiKey,
-      {String appUserId, bool observerMode = false}) async {
-    if (appUserId == null) {
-      return await _channel.invokeMethod('setupPurchases', {'apiKey': apiKey});
-    } else {
```
aboedo, line 39: removed this if else block, and replaced it with actually checking if the value is NSNull and translating to nil.

android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
```diff
@@ -71,6 +71,7 @@ public void onMethodCall(MethodCall call, Result result) {
                 String apiKey = call.argument("apiKey");
                 String appUserId = call.argument("appUserId");
                 Boolean observerMode = call.argument("observerMode");
+                String userDefaultsSuiteName = call.argument("userDefaultsSuiteName"); // iOS-only, unused.
```
aboedo, line 74: we could also just not read this value at all, but I figured having something explicit might be better to that future devs wouldn't be confused by the disparity between iOS and Android implementations

lib/purchases_flutter.dart
aboedo: note that there was a bug here - if you passed in `observerMode` but not `appUserId`, `observerMode` is ignored

vegaro: funny there was a similar error in react native

vegaro: Just wanted to double check that removing this if else doesn't break Android

ios/Classes/PurchasesFlutterPlugin.m
```diff
@@ -41,7 +41,8 @@ - (void)handleMethodCall:(FlutterMethodCall *)call result:(FlutterResult)result
         NSString *apiKey = arguments[@"apiKey"];
         NSString *appUserID = arguments[@"appUserId"];
         BOOL observerMode = [arguments[@"observerMode"] boolValue];
-        [self setupPurchases:apiKey appUserID:appUserID observerMode:observerMode result:result];
+        NSString * _Nullable userDefaultsSuiteName = arguments[@"userDefaultsSuiteName"];
```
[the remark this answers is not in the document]
aboedo: `_Nullable` ironically means it can be nil, not null. obj-c naming at its best. 🤷🏻‍♂️

lib/purchases_flutter.dart
aboedo: yep, it works correctly in Android

Issue #82, opened 2020-07-03
[Android] NullPointerException on sendEvent


atrope:
Hi,
We've updated to the latest version (1.2) and started to see some errors in Crashlytics.

We'd like to know if this could interfere in our results or  we could ignore it..
```

java.lang.NullPointerException: 
  at com.revenuecat.purchases_flutter.PurchasesFlutterPlugin.sendEvent (PurchasesFlutterPlugin.java:2)
  at com.revenuecat.purchases_flutter.PurchasesFlutterPlugin.access$000 (PurchasesFlutterPlugin.java)
  at com.revenuecat.purchases_flutter.PurchasesFlutterPlugin$2.onReceived (PurchasesFlutterPlugin.java:8)
  at com.revenuecat.purchases.Purchases$sendUpdatedPurchaserInfoToDelegateIfChanged$$inlined$let$lambda$1.invoke (Purchases.java:4)
  at com.revenuecat.purchases.Purchases$sendUpdatedPurchaserInfoToDelegateIfChanged$$inlined$let$lambda$1.invoke (Purchases.java)
  at com.revenuecat.purchases.Purchases$sam$java_lang_Runnable$0.run (Purchases.java:2)
  at android.os.Handler.handleCallback (Handler.java:873)
  at android.os.Handler.dispatchMessage (Handler.java:99)
  at android.os.Looper.loop (Looper.java:214)
  at android.app.ActivityThread.main (ActivityThread.java:7156)
  at java.lang.reflect.Method.invoke (Method.java)
  at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run (RuntimeInit.java:494)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:975)

```

Thanks

vegaro:
Hi @atrope 

Looks like it's a race condition caused when an updated `PurchaserInfo` arrives after `onDetachedFromEngine` has been called in the plugin, but before we finish cleaning the instance of the Android SDK. So it looks like it's a crash when 

How do the numbers look in Crashlytics? Is it widespread?

atrope:
We had 2  messages like that, and also 2 new subscribers in that app after the update.

vegaro:
Hi, just released https://github.com/RevenueCat/purchases-flutter/releases/tag/1.2.1 which should have the fix. Let us know if you face more issues. Sorry about that!

atrope:
[...]

PR #83, opened 2020-07-03
Fixes NullPointerException on updated PurchaserInfo

vegaro:
Fixes https://github.com/RevenueCat/purchases-flutter/issues/82

Looks like it's a race condition caused when an updated PurchaserInfo arrives after onDetachedFromEngine has been called in the plugin, but before we finish cleaning the instance of the Android SDK. So it looks like it's a crash when

### DIFF
--- a/android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
+++ b/android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
@@ -9,6 +9,7 @@ import androidx.annotation.Nullable;
 import com.revenuecat.purchases.PlatformInfo;
 import com.revenuecat.purchases.PurchaserInfo;
 import com.revenuecat.purchases.Purchases;
+import com.revenuecat.purchases.PurchasesErrorCode;
 import com.revenuecat.purchases.common.CommonKt;
 import com.revenuecat.purchases.common.ErrorContainer;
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
```diff
@@ -235,31 +240,53 @@ public void onMethodCall(MethodCall call, Result result) {
...
+            result.success(null);
+        } else {
+            result.error(
+                    String.valueOf(PurchasesErrorCode.UnknownError.ordinal()),
+                    "Purchases can't be setup. There is no Application context",
+                    null);
+        }
+    }
 
+    private void setupUpdatedPurchaserInfoListener() {
```
vegaro, line 257: Moved this to a method

android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
```diff
@@ -235,31 +240,53 @@ public void onMethodCall(MethodCall call, Result result) {
         }
     }
 
-    private void sendEvent(String eventName, @Nullable Map<String, Object> params) {
```
vegaro, line 238: this was only used by the updatedPurchaserInfo listener, so I moved it in the new method `setupUpdatedPurchaserInfoListener` and added a null check

PR #84, opened 2020-07-03
Release 1.2.1

vegaro:
Going to release this to fix the bug since the fix is straightforward.


vegaro:
I made the release using git flow, and this got closed 🤦

PR #96, opened 2020-09-19
release 1.3.0

aboedo:
### 1.3.0

- Attribution V2:
    - Deprecated `addAttribution` in favor of `setAdjustId`, `setAppsflyerId`, `setFbAnonymousId`, `setMparticleId`.
    - Added support for OneSignal via `setOnesignalId`
    - Added `setMediaSource`, `setCampaign`, `setAdGroup`, `setAd`, `setKeyword`, `setCreative`, and `collectDeviceIdentifiers`
         https://github.com/RevenueCat/purchases-android/pull/184
- Bumped `purchases-hybrid-common` to 1.4.2 [Changelog here](https://github.com/RevenueCat/purchases-hybrid-common/releases)
- Bumped `purchases-ios` to 3.7.1 [Changelog here](https://github.com/RevenueCat/purchases-ios/releases)
- Bumped `purchases-android` to 3.5.0 [Changelog here](https://github.com/RevenueCat/purchases-android/releases)


vegaro (approved):
LGTM

vegaro (approved):
Sorry I didn't notice about `nonSubscriptionTransactions` on the previous review

lib/transaction.dart
```diff
@@ -0,0 +1,24 @@
+class Transaction {
+  /// RevenueCat Id associated to the transaction.
+  final String revenuecatId;
```
vegaro, line 3: I just realized the c is not capitalized in the Android and iOS class, and that the d is lowercase in `Id` 🤦

lib/purchases_flutter.dart
```diff
@@ -374,10 +376,106 @@ class Purchases {
...
+  /// Subscriber attribute associated with the install ad creative for the user
+  ///
+  /// [creative] Empty String or null will delete the subscriber attribute.
+  static Future<void> setCreative(String creative) async {
+    await _channel.invokeMethod('setCreative', {'creative': creative});
+  }
+
+  ///
+  /// Automatically collect subscriber attributes associated with the device identifiers
+  /// $idfa, $idfv, $ip
```
vegaro, line 475: This should be something like

```
   * $idfa, $idfv, $ip on iOS
   * $gpsAdId, $androidId, $ip on Android
```

vegaro (changes requested):

lib/transaction.dart
aboedo: I'm capitalizing the C, but the d in `ID` I'm not updating right now because it should match the original iOS code - once we update it in `purchases-ios`, we can update it here

aboedo: actually, it's a little more problematic - it's capitalized here: 
https://github.com/RevenueCat/purchases-hybrid-common/blob/develop/ios/PurchasesHybridCommon/PurchasesHybridCommon/RCTransaction%2BHybridAdditions.m#L17
and not here: 
https://github.com/RevenueCat/purchases-hybrid-common/blob/develop/android/src/main/java/com/revenuecat/purchases/common/mappers/TransactionMapper.kt#L7
so we'll need another patch release to fix it, otherwise mappers won't work. 
I'll take care of it

aboedo: okay, tested and working with common 1.4.2 https://github.com/RevenueCat/purchases-hybrid-common/pull/55

This PR will need to bump the version number once the hybrid-common release is out, though. adding a hold until then.

aboedo:
@vegaro this is ready for another pass

lib/purchaser_info_wrapper.dart
```diff
@@ -19,6 +20,9 @@ class PurchaserInfo {
   /// Set of purchased skus, active and inactive
   final List<String> allPurchasedProductIdentifiers;
 
+  /// Set of purchased skus, active and inactive
```
vegaro, line 23: I believe this docs are copy pasted from `allPurchasedProductIdentifiers`

vegaro (approved):
Feel free to merge when those docs are fixed

lib/purchaser_info_wrapper.dart
aboedo: lol good catch, fixing

Issue #99, opened 2020-09-23
iOS failed to Compile after  1.3.0 


atrope:
Hello,
After updating to 1.3.0 iOS stopped building for with the following error:

```
Ld /Users/trope/Library/Developer/Xcode/DerivedData/Runner-ewvwfgyxollwiacgjqgpvwzsvjkh/Build/Intermediates.noindex/Pods.build/Release-iphoneos/purchases_flutter.build/Objects-normal/arm64/Binary/purchases_flutter normal arm64 (in target 'purchases_flutter' from project 'Pods')
    cd /Users/trope/suamusica/OneApp/ios/Pods
[...]

PR #100, opened 2020-09-24
Fix: build issues

aboedo:
Addresses https://github.com/RevenueCat/purchases-flutter/issues/99

- Added `DEFINES_MODULE` build setting for cocoapods projects
- added swift version
- made framework compile statically
- replaced `#import` with `@import`

### Requirements: 
- [x] purchases-ios release that includes https://github.com/RevenueCat/purchases-ios/pull/353
- [x] purchases-hybrid-common release that includes https://github.com/RevenueCat/purchases-hybrid-common/pull/56

### DIFF
--- a/ios/purchases_flutter.podspec
+++ b/ios/purchases_flutter.podspec
@@ -16,6 +16,10 @@ Pod::Spec.new do |s|
   s.public_header_files = 'Classes/**/*.h'
   s.dependency 'Flutter'
   s.dependency 'PurchasesHybridCommon', '1.4.2'
+  s.pod_target_xcconfig   = { 'DEFINES_MODULE' => 'YES' }
   s.ios.deployment_target = '9.0'
+  s.static_framework      = true
+  s.swift_version         = '5.0'
+
 end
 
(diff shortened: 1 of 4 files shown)


ios/purchases_flutter.podspec
```diff
@@ -16,6 +16,10 @@ Pod::Spec.new do |s|
   s.public_header_files = 'Classes/**/*.h'
   s.dependency 'Flutter'
   s.dependency 'PurchasesHybridCommon', '1.4.2'
+  s.pod_target_xcconfig   = { 'DEFINES_MODULE' => 'YES' }
```
vegaro, line 19: what does this do?

vegaro: I created a new Flutter plugin and they do this by default:
```
 # Flutter.framework does not contain a i386 slice. Only x86_64 simulators are supported.
  s.pod_target_xcconfig = { 'DEFINES_MODULE' => 'YES', 'VALID_ARCHS[sdk=iphonesimulator*]' => 'x86_64' }
```

vegaro (approved):

ios/purchases_flutter.podspec
aboedo: it enables stricter header search paths, meaning that hopefully importing errors would surface before being able to publish the pod. 
it also defines a Module, which allows for the semantic import syntax, i.e.: `@import Purchases`. 
I'll add the VALID_ARCHS thing 👍

vegaro (approved):

PR #103, opened 2020-09-25
MacOS support

vegaro:
Adds macOS support and other files automatically added in the other platforms.


example/android/app/src/main/res/values-night/styles.xml
```diff
@@ -0,0 +1,18 @@
...
+    <!-- Theme applied to the Android Window as soon as the process has started.
+         This theme determines the color of the Android Window while your
+         Flutter UI initializes, as well as behind your Flutter UI while its
+         running.
+         
+         This Theme is only used starting with V2 of Flutter's Android embedding. -->
+    <style name="NormalTheme" parent="@android:style/Theme.Black.NoTitleBar">
+        <item name="android:windowBackground">?android:colorBackground</item>
+    </style>
+</resources>
```
vegaro, line 18: These Android files were automatically added when creating the macOS platform

ios/Classes/PurchasesFlutterPlugin.h
```diff
@@ -1,4 +1,8 @@
+#if TARGET_OS_OSX
+#import <FlutterMacOS/FlutterMacOS.h>
+#else
 #import <Flutter/Flutter.h>
+#endif
```
vegaro, line 5: This is the main change

vegaro:
@aboedo bump

macos/Classes/PurchasesFlutterPlugin.h
```diff
@@ -0,0 +1 @@
+../../ios/Classes/PurchasesFlutterPlugin.h
```
aboedo, line 1: were these links done automatically?

macos/purchases_flutter.podspec
```diff
@@ -0,0 +1,24 @@
+#
+# To learn more about a Podspec see http://guides.cocoapods.org/syntax/podspec.html.
+# Run `pod lib lint purchases_flutter.podspec' to validate before publishing.
+#
+Pod::Spec.new do |s|
+  s.name             = 'purchases_flutter'
```
aboedo, line 6: I suppose it doesn't matter, since these wouldn't be published anyway, but any concerns about this having the same name as the iOS pods? or is it intentional so that the import is the same?

aboedo (approved):

aboedo:
nice work on this one! I'm psyched we have macOS support!

macos/Classes/PurchasesFlutterPlugin.h
vegaro: I made this link to avoid having to copy paste these files every time we update the iOS part of the plugin

macos/purchases_flutter.podspec
vegaro: this one was created automatically by Flutter and I didn't have any problems when building so I think we are fine

AndroidNils:
Yeah! Many thanks for the work! 💙

Issue #25, opened 2020-03-07
Any plans to support ios subscription offers in the flutter sdk?


ericmartineau:
I read through the docs here: https://www.revenuecat.com/blog/signing-ios-subscription-offers but couldn't find the necessary calls in the flutter sdk. 

Do you guys have plans to implement this?

aboedo:
Hi Eric! Thanks for writing in. We do have plans to include subscription offers in flutter, we will likely roll it out within the next month!

nemoryoliver:
> Hi Eric! Thanks for writing in. We do have plans to include subscription offers in flutter, we will likely roll it out within the next month!

Any update on this?

gabrielfgularte:
Waiting for an update too.

leonardocustodio:
No updates on this yet? I have to use another package just to make this feature work. I would love to do everything through revenuecat package instead.

vegaro:
I am working on this atm and also adding MacOS support @leonardocustodio Sorry for taking so long to make these changes

leonardocustodio:
Nice, thank you very much for the great work.

vegaro:
This will go out in 1.4.0 along with the MacOS support

PR #104, opened 2020-09-26
Adds subscription offers support

vegaro:
- Adds getPaymentDiscount to get a signed discount.
- Creates a purchaseDiscountedProduct and purchaseDiscountedPackage. Since there is no overloads in dart, and upgrade info and type parameters in the regular purchase calls are only relevant to Android I figured it was cleaner to create new functions.

https://github.com/RevenueCat/purchases-flutter/issues/25


lib/product_wrapper.dart
```diff
@@ -74,3 +82,61 @@ class IntroductoryPrice {
     return 'IntroductoryPrice{introPrice: $introPrice, introPriceString: $introPriceString, introPricePeriod: $introPricePeriod, introPricePeriodUnit: $introPricePeriodUnit, introPricePeriodNumberOfUnits: $introPricePeriodNumberOfUnits, introPriceCycles: $introPriceCycles}';
   }
 }
+
+/// The details of a discount offer for a subscription product.
+class Discount {
```
aboedo, line 87: NABD, but should this be its own file?

aboedo (approved):
looks great!

lib/product_wrapper.dart
vegaro: sure, I will extract it

aboedo (approved):

PR #118, opened 2020-10-21
Adds tests and CircleCI

vegaro:
- Adds CircleCI config file which runs the following jobs:
  - lint
  - test (to run unit tests)
  - android-integration-test: runs a simple test on an emulator to verify the SDK can be configured, an appUserID is assigned and purchaserInfo can be fetched.
  - ios-integration-test: same as the Android test but on an iOS simulator

- The api key is passed via an env variable configured in CircleCI.
- The UpsellScreen in the example now takes care of fetching offerings itself and this is not done when loading the App anymore.
- Removes the `e2e` package and the test we had using it.


example/pubspec.yaml
```diff
@@ -21,9 +21,10 @@ dev_dependencies:
     path: ../
 
   pedantic: ^1.8.0
-  e2e: ^0.2.1
```
vegaro, line 24: This has been deprecated and replaced with [integration_test](https://github.com/flutter/plugins/tree/master/packages/integration_test), but I actually don't think we need it.

vegaro:
@aboedo don't review yet

vegaro:
The tests are running out of space and the Android emulator is flaky. I am going to try with another service or with Firebase Test Lab because this is not reliable

vegaro:
@aboedo ready

.circleci/config.yml
```diff
@@ -0,0 +1,243 @@
...
+      - test:
+          requires:
+            - checkout_code
+  android-integration-test:
+    jobs:
+      - checkout_code
+      - android-integration-test:
+          requires:
+            - checkout_code
+  ios-integration-test:
```
vegaro, line 238: split in two workflows because I was running out of space in the partition. It looks like this helps

.circleci/config.yml
```diff
@@ -0,0 +1,243 @@
...
+    description: "setup Flutter"
+    steps:
+      - run:
+          name: Setup environment variables
+          command: echo 'export PATH="$PATH:`pwd`/flutter/bin"'  >> $BASH_ENV
+      - restore_cache:
+          key: flutter_cache_1
+      - run:
+          name: Download flutter SDK
+          command: if ! test -f "flutter_sdk.zip"; then curl -o flutter_sdk.zip https://storage.googleapis.com/flutter_infra/releases/stable/macos/flutter_macos_1.22.2-stable.zip; fi
```
aboedo, line 113: maybe we could make this one multiline for clarity? i.e.: 
```yaml
command: |
    if ! test -f "flutter_sdk.zip"; then 
    curl -o flutter_sdk.zip 
    https://storage.googleapis.com/flutter_infra/releases/stable/macos/flutter_macos_1.22.2-stable.zip; 
    fi
```
not sure if that works off the bat or you need `\` for line breaks or something

.circleci/config.yml
```diff
@@ -0,0 +1,243 @@
...
+  test:
+    description: "Run tests for Flutter"
+    docker:
+      - image: cirrusci/flutter:stable
+    steps:
+      - attach_workspace:
+          at: .
+      - run:
+          name: Install Dependencies
+          command: flutter pub get
```
aboedo, line 177: we're doing this everywhere, is there a way to reuse it?

.circleci/config.yml
aboedo: we'll also get a separate check per workflow, right? And they can potentially run in parallel?

example/integration_test/app_test.dart
```diff
@@ -0,0 +1,23 @@
...
+  IntegrationTestWidgetsFlutterBinding.ensureInitialized();
+
+    setUpAll(() async {
+      print("setupAll");
+      const apiKey = String.fromEnvironment("API_KEY", defaultValue: null);
+      await Purchases.setup(apiKey, appUserId: "test-user");
+    });
+
+  test("Configures successfully", () async {
+    expect(Purchases.appUserID != null, true);
```
aboedo, line 16: other than checking that it won't crash, does this actually confirm that it's a successful config? i.e.: if the communication with the backend fails, wouldn't this still work?

example/integration_test/app_test.dart
```diff
@@ -0,0 +1,23 @@
...
+  });
+
+  test("Purchaser info fetched successfully", () async {
+    PurchaserInfo purchaserInfo = await Purchases.getPurchaserInfo();
```
aboedo, line 20: what I wouldn't give to be able to use `await` everywhere.

example/integration_test/app_test.dart
aboedo: the purchaserInfo test should catch most issues correctly, though. 
so maybe we just need to rename this to `Configures without crashing` or something to that effect

example/lib/main.dart
```diff
@@ -29,17 +28,17 @@ class _MyAppState extends State<InitialScreen> {
   // Platform messages are asynchronous, so we initialize in an async method.
   Future<void> initPlatformState() async {
     await Purchases.setDebugLogsEnabled(true);
-    await Purchases.setup("api_key");
+    const apiKey = String.fromEnvironment("API_KEY", defaultValue: null);
```
aboedo, line 31: I wish getting env variables was this easy everywhere.

aboedo: does this mean that from now on, for testing, we need to replace the default value here?

scripts/install-android-tools.sh
```diff
@@ -0,0 +1,28 @@
...
+curl https://dl.google.com/android/repository/sdk-tools-darwin-4333796.zip -o sdk-tools.zip
+
+unzip sdk-tools.zip
+
+mkdir -p "$ANDROID_SDK_ROOT/licenses"
+
+echo "24333f8a63b6825ea9c5514f83c2829b004d1fee" > "$ANDROID_SDK_ROOT/licenses/android-sdk-license"
+echo "84831b9409646a918e30573bab4c9c91346d8abd" > "$ANDROID_SDK_ROOT/licenses/android-sdk-preview-license"
+echo "d975f751698a77b662f1254ddbeed3901e976f5a" > "$ANDROID_SDK_ROOT/licenses/intel-android-extra-license"
+
```
aboedo, line 18: should these be env variables?

example/lib/main.dart
vegaro: Well, you know that it only works if it's a const? haha 

Also, yes, the defaultValue should be replace for testing. Should I add a comment?

.circleci/config.yml
vegaro: Yes, I am not sure about the parallel, I think we might need to enable that, but I think potentially yes. I honestly like the two checks

scripts/install-android-tools.sh
vegaro: nah, this is all over the internet, just as an example https://github.com/material-components/material-components-android/blob/master/.travis.yml

.circleci/config.yml
```diff
@@ -0,0 +1,242 @@
+version: 2.1
+
+orbs:
+  android: circleci/android@0.2.1
+  react-native: react-native-community/react-native@4.4.2
```
aboedo, line 5: react native?

.circleci/config.yml
```diff
@@ -0,0 +1,242 @@
...
+
+commands:
+
+  build-flutter-project:
+    description: "Install dependencies and build"
+    parameters:
+      working_directory:
+        default: .
```
aboedo, line 13: is this parameter automatically used?

.circleci/config.yml
```diff
@@ -0,0 +1,242 @@
...
+    steps:
+      - run:
+          name: Run tests
+          command: |
+            cd example
+            flutter drive --dart-define=API_KEY=$API_KEY --target=integration_test/app_test.dart \
+            --driver=test_driver/integration_test.dart
+
+  install-and-build-flutter:
+    description: "Install and build Flutter"
```
aboedo, line 34: this command seems fundamentally the same as `build-flutter-project`, but with a `cd example`. perhaps they can be unified?

.circleci/config.yml
```diff
@@ -0,0 +1,242 @@
...
+      - run:
+          name: install android tools
+          command: |
+            sh scripts/install-android-tools.sh
+            echo 'export PATH=$ANDROID_SDK_ROOT/tools/bin:$PATH'  >> $BASH_ENV
+            echo 'export PATH=$ANDROID_SDK_ROOT/tools:$PATH'  >> $BASH_ENV
+            echo 'export PATH=$ANDROID_SDK_ROOT/platform-tools:$PATH'  >> $BASH_ENV
+            echo 'export PATH=$ANDROID_SDK_ROOT/emulator:$PATH'  >> $BASH_ENV
+            source $BASH_ENV
+            sdkmanager --list
```
aboedo, line 95: does `sdkmanager --list` have a side-effect or is it for debug purposes?

.circleci/config.yml
```diff
@@ -0,0 +1,242 @@
...
+      - save_cache:
+          key: android=tools-v1-{{ checksum "scripts/install-android-tools.sh" }}-{{ arch }}
+          paths:
+            - /Users/distiller/android-tools
```
aboedo, line 99: let's maybe make this `~/android-tools` in case the `/Users/distiller/` path changes in the future

.circleci/config.yml
```diff
@@ -0,0 +1,242 @@
...
+
+  create-launch-android-emulator:
+    description: "create and launch android emulators"
+    steps:
+      - run:
+          name: create AVD
+          command: echo "no" | avdmanager --verbose create avd --force --name "Pixel_3a_API_29" --package "system-images;android-29;google_apis_playstore;x86_64"
```
aboedo, line 106: maybe we can add some line breaks for readability?

.circleci/config.yml
```diff
@@ -0,0 +1,242 @@
...
+          name: Setup environment variables
+          command: echo 'export PATH="$PATH:`pwd`/flutter/bin"'  >> $BASH_ENV
+      - restore_cache:
+          key: flutter_cache_1
+      - run:
+          name: Download flutter SDK
+          command: |
+            if ! test -f "flutter_sdk.zip"; then 
+            curl -o flutter_sdk.zip 
+            https://storage.googleapis.com/flutter_infra/releases/stable/macos/flutter_macos_1.22.2-stable.zip;
```
aboedo, line 130: this link is already out of date 😬 
maybe instead of downloading from here, we should 
```
git clone -b stable https://github.com/flutter/flutter.git
```
that also saves us from the zip step

example/lib/main.dart
aboedo: 🤔 it might lead to confusing situations if customers already have `API_KEY` set to something. 
maybe `REVENUECAT_API_KEY`?

vegaro:
@aboedo what do you think the trigger for this tests should be? tagging a version?

aboedo:
yeah, I think that would work. If they're not super long, we could have tagged versions + `release/*` branches

.circleci/config.yml
vegaro: Oh I was using some stuff from this orb but not anymore I will remove it

.circleci/config.yml
vegaro: I cleaned it up a little bit since there were two commands doing the same. I don't think this parameter was being automatically used btw

.circleci/config.yml
vegaro: yes, good catch, I realized that and changed it in 6dc9ba5995f5e4d1a225fbb42582aa2cb32576ba

.circleci/config.yml
vegaro: I think I found it somewhere else. I believe it's just for debugging purposes since it will display all the installed images. Do you think we should remove it?

aboedo: I don't mind it, but let's maybe add a comment

.circleci/config.yml
```diff
@@ -0,0 +1,191 @@
...
+    steps:
+      - checkout
+      - setup-flutter
+      - restore_cache:
+          name: Restore cocoapods specs and pods
+          key: v1-cocoapods-{{ checksum "./example/ios/Podfile.lock" }}-{{ arch }}
+      - simulator-dependencies
+      - run:
+          name: Open simulator
+          command: xcrun simctl boot "iPhone 11"
```
aboedo, line 172: is this needed? 
Also, we might wanna pre-start it https://circleci.com/docs/2.0/testing-ios/#reducing-job-time-and-best-practises

DEVELOPMENT.md
```diff
@@ -1,3 +1,9 @@
+- Edit the `defaultValue` in `example/lib/main.dart` for the apiKey with the sample app to test:
```
aboedo, line 1: The env variable only works in CI, right? i.e.: if you have it in your `.bash_profile` or `.zshrc` it won't get picked up?

example/lib/main.dart
```diff
@@ -57,23 +56,45 @@ class _MyAppState extends State<InitialScreen> {
...
+  }
+
+  Future<void> fetchData() async {
+    Offerings offerings;
+    try {
+      offerings = await Purchases.getOfferings();
+    } on PlatformException catch (e) {
+      print(e);
+    }
+
```
aboedo, line 86: does this need the sim to be logged into to app store / play store?

DEVELOPMENT.md
vegaro: so I am actually thinking about using `sed` like in the other repos and reemove this whole env variable thing

aboedo: yeah, that probably works best. I ended up doing it for iOS integration tests as well, since env variables are not exactly easy to work with on iOS

example/lib/main.dart
vegaro: We had it in the initial screen and I moved it to the UpsellScreen. I am not sure about Flutter lifecycle's, but I don't want people to think they have to fetch offerings as soon as the app starts. 

This change is independent from the integration test tho. I was doing UI tests originally, so I actually needed this to not crash the app in the sim. I thought it would be nice to keep it. I can revert it if you think we don't need to do this change now :)

vegaro:
@aboedo I am trying to figure out why the tests are actually failing but being reported as succeeds 😆

.circleci/config.yml
```diff
@@ -0,0 +1,198 @@
+version: 2.1
+
+orbs:
+  android: circleci/android@0.2.1
+  gcp-cli: circleci/gcp-cli@2.1.0
+
+commands:
+
+  prepare-tests:
+    description: "prepares tests to run"
```
aboedo, line 10: I think this should be `Replace API_KEY` for consistency with https://github.com/RevenueCat/purchases-android/pull/215/files#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47R166, and it's also more accurate

aboedo: note that unifying also helps if we actually want to build an orb to reuse these things

.circleci/config.yml
```diff
@@ -0,0 +1,198 @@
...
+
+  android-integration-test-build:
+    description: "Build Android Flutter integration tests"
+    docker:
+      - image: cirrusci/flutter:stable
+    steps:
+      - checkout
+      - run:
+          name: Change api key
+          command: sed -i s/api_key/$API_KEY/ example/integration_test/app_test.dart
```
aboedo, line 106: this should reuse the command defined at the beginning of the file

.circleci/config.yml
```diff
@@ -0,0 +1,198 @@
...
+          gcloud-service-key: GCLOUD_SERVICE_KEY
+          google-compute-zone: GOOGLE_COMPUTE_ZONE
+          google-project-id: GOOGLE_PROJECT_ID
+      - run:
+          name: Test with Firebase Test Lab
+          command: >
+            gcloud firebase test android run --type instrumentation \
+              --app example/build/app/outputs/apk/debug/app-debug.apk \
+              --test example/build/app/outputs/apk/androidTest/debug/app-debug-androidTest.apk \
+              --timeout 2m \
```
aboedo, line 137: I have the same concerns about reusing the example app for integration tests as I mentioned in the cordova PR, they will evolve in different ways with different responsibilities

.circleci/config.yml
```diff
@@ -0,0 +1,198 @@
...
+          name: Open simulator
+          command: xcrun instruments -w "iPhone 11 Pro (13.3) [" || true
+      - setup-flutter
+      - restore_cache:
+          name: Restore cocoapods specs and pods
+          key: v1-cocoapods-{{ checksum "./example/ios/Podfile.lock" }}-{{ arch }}
+      - prepare-tests
+      - build-flutter-project:
+          project: example
+      - run-flutter-integration-tests
```
aboedo, line 165: is this guaranteed to use the `iPhone 11 Pro (13.3)` or is it the device default for the executor and xcode version?

DEVELOPMENT.md
```diff
@@ -1,3 +1,9 @@
...
+
+```
+    const apiKey = String.fromEnvironment("API_KEY", defaultValue: null);
```
aboedo, line 4: does this need to be updated?

.circleci/config.yml
vegaro: It looks like it runs in the booted device. I upgraded xcode to 12.2 and it still uses the iPhone 11 Pro that is started couple steps before

.circleci/config.yml
vegaro: Ok, so I just remembered why I had this different. This runs in a linux machine and it gives issues when I run `-i .bck` with the `.bck` extension. It looks like the syntax is different in macos and linux. I will see if I can come up with something compatible for both

aboedo (approved):
🎉

PR #151, opened 2021-02-02
Release 2.0.3

aboedo:
Addresses #150


aboedo:
@vegaro this would have been caught by an integration test on macOS, I think it's worth adding an extra test. But let's do it as a separate PR so we can push the fix now

aboedo:
it might require some work as we'd have to use the dev channel. easy to do, though. I'll add it to CH

vegaro (approved):

PR #154, opened 2021-02-23
Removes all analyze warnings

vegaro:
I started migrating to null safety and one of the steps is to run `dart analyze`. I fixed all issues. Some of them are self explanatory docs, so happy to get suggestions on those.

PR #155, opened 2021-02-23
Migrates Flutter to null safety

vegaro:
I followed the steps in https://dart.dev/null-safety/migration-guide#step3-analyze and this PR is the result of the automatic migration.

We can publish this as version 3.0.0 of the SDK and it will show as a preview until dart 2.12.0 is out of beta.

Depends on https://github.com/RevenueCat/purchases-flutter/pull/154

### DIFF
--- a/lib/entitlement_info_wrapper.dart
+++ b/lib/entitlement_info_wrapper.dart
@@ -23,7 +23,7 @@ class EntitlementInfo {
 
   /// The expiration date for the entitlement, can be `null` for lifetime access.
   /// If the `periodType` is `trial`, this is the trial expiration date.
-  final String expirationDate;
+  final String? expirationDate;
 
   /// The store where this entitlement was unlocked from
   /// Either: appStore, macAppStore, playStore, stripe, promotional, unknownStore
@@ -38,13 +38,13 @@ class EntitlementInfo {
   /// The date an unsubscribe was detected. Can be `null`.
   /// @note: Entitlement may still be active even if user has unsubscribed.
   /// Check the `isActive` property.
-  final String unsubscribeDetectedAt;
+  final String? unsubscribeDetectedAt;
 
   /// The date a billing issue was detected. Can be `nil` if there is no
   /// billing issue or an issue has been resolved.
   /// @note: Entitlement may still be active even if there is a billing issue.
   /// Check the `isActive` property.
-  final String billingIssueDetectedAt;
+  final String? billingIssueDetectedAt;
 
   /// Construct an EntitlementInfo
   EntitlementInfo(
@@ -63,8 +63,8 @@ class EntitlementInfo {
 
   /// Constructs an EntitlementInfo from a JSON object
   factory EntitlementInfo.fromJson(Map<dynamic, dynamic> json) {
-    var periodType;
-    switch (json["periodType"] as String) {
+    late var periodType;
+    switch (json["periodType"] as String?) {
       case "INTRO":
         periodType = PeriodType.intro;
         break;
@@ -75,8 +75,8 @@ class EntitlementInfo {
         periodType = PeriodType.trial;
         break;
     }
-    var store;
-    switch (json["store"] as String) {
+    late var store;
+    switch (json["store"] as String?) {
       case "APP_STORE":
         store = Store.appStore;
         break;
@@ -103,12 +103,12 @@ class EntitlementInfo {
         periodType,
         json["latestPurchaseDate"] as String,
         json["originalPurchaseDate"] as String,
-        json["expirationDate"] as String,
+        json["expirationDate"] as String?,
         store,
         json["productIdentifier"] as String,
         json["isSandbox"] as bool,
-        json["unsubscribeDetectedAt"] as String,
-        json["billingIssueDetectedAt"] as String);
+        json["unsubscribeDetectedAt"] as String?,
+        json["billingIssueDetectedAt"] as String?);
   }
 
   @override
--- a/lib/entitlement_info_wrapper.dart
+++ b/lib/entitlement_info_wrapper.dart
@@ -35,12 +35,13 @@ class EntitlementInfo {
   /// False if this entitlement is unlocked via a production purchase
   final bool isSandbox;
 
-  /// The date an unsubscribe was detected. Can be `null`.
+  /// The date an unsubscribe was detected. Can be `null` if it's still
+  /// subscribed or product is not a subscription.
   /// @note: Entitlement may still be active even if user has unsubscribed.
   /// Check the `isActive` property.
   final String? unsubscribeDetectedAt;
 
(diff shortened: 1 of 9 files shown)


aboedo:
it looks like the lint error will go away once this gets rebased from #154, so we should be good

lib/entitlement_info_wrapper.dart
```diff
@@ -38,13 +38,13 @@ class EntitlementInfo {
   /// The date an unsubscribe was detected. Can be `null`.
   /// @note: Entitlement may still be active even if user has unsubscribed.
   /// Check the `isActive` property.
-  final String unsubscribeDetectedAt;
+  final String? unsubscribeDetectedAt;
 
   /// The date a billing issue was detected. Can be `nil` if there is no
```
aboedo, line 43: we should probably update the end of this line to "can be `null` if there is no"

lib/entitlement_info_wrapper.dart
```diff
@@ -61,8 +61,8 @@ class EntitlementInfo {
       this.billingIssueDetectedAt);
 
   factory EntitlementInfo.fromJson(Map<dynamic, dynamic> json) {
-    var periodType;
-    switch (json["periodType"] as String) {
+    late var periodType;
+    switch (json["periodType"] as String?) {
```
aboedo, line 65: what does this switch do if periodType is null? do we need to add a case to handle it?

lib/entitlement_info_wrapper.dart
```diff
@@ -73,8 +73,8 @@ class EntitlementInfo {
         periodType = PeriodType.trial;
         break;
     }
-    var store;
-    switch (json["store"] as String) {
+    late var store;
+    switch (json["store"] as String?) {
```
aboedo, line 77: same question regarding switching on null

lib/purchases_flutter.dart
```diff
@@ -175,13 +175,13 @@ class Purchases {
   /// [upgradeInfo] Android only. Optional UpgradeInfo you wish to upgrade from
   /// containing the oldSKU and the optional prorationMode.
   static Future<PurchaserInfo> purchasePackage(Package packageToPurchase,
-      {UpgradeInfo upgradeInfo}) async {
+      {UpgradeInfo? upgradeInfo}) async {
     var response = await _channel.invokeMethod('purchasePackage', {
       'packageIdentifier': packageToPurchase.identifier,
       'offeringIdentifier': packageToPurchase.offeringIdentifier,
       'oldSKU': upgradeInfo != null ? upgradeInfo.oldSKU : null,
       'prorationMode': upgradeInfo != null && upgradeInfo.prorationMode != null
```
aboedo, line 183: do we need to be doing `upgradeInfo!.prorationMode` here?

lib/purchases_flutter.dart
```diff
@@ -175,13 +175,13 @@ class Purchases {
...
-          ? upgradeInfo.prorationMode.index
+          ? upgradeInfo.prorationMode!.index
```
aboedo, line 184: same question here, does this need to also be unwrapping `upgradeInfo`?

aboedo (commented):
looking great! left a couple of questiosn

lib/purchases_flutter.dart
IchordeDionysos: It's not needed here as `upgradeInfo` is a function parameter and thus the analyzer can statically make sure that it's never `null`.

Unwrapping would cause a compilation error.

lib/purchases_flutter.dart
IchordeDionysos: Here we do need unwrapping as `prorationMode` could be implemented like this:

```dart
String? get prorationMode {
  if (random() < 0.5) return null;
  return 'RevenueCat';
}
```

And the analyzer can't be sure that every time `upgradeInfo.prorationMode` is called the value is always null.

What I would suggest to do instead is this:

```dart
final prorationMode = upgradeInfo?.prorationMode;

prorationMode != null ? prorationMode.index : null
```

With that we don't need to assert that it's not null and the analyzer knows it itself.

lib/purchases_flutter.dart
IchordeDionysos: More information about this one:
https://github.com/dart-lang/sdk/issues/42626

lib/entitlement_info_wrapper.dart
IchordeDionysos: If `periodType` is `null` then no case will match and periodType will still NOT be initialized.

And when it‘s then passed to `EntitlementInfo()` constructor as it‘s expected to be non-null it will crash.

There are three options:
- Leave it as it is and let it crash.
- Add a `case null: ...` to handle the null case.
- Make `periodType` nullable `String?` and be explicit that it may be null and handle it apprioriately.

I personally would prefer the last. So doing this:

```dart
String? periodType;
```

And the later:
```
EntitlementInfo(
  ...,
  periodType!,
  ...
  // Maybe this for the store (?!)
  store ?? Store.unknownStore,
);
```

I think you are better suited to judge what consequences would be acceptable.
Completely, preventing any errors while parsing is probably never possible.

vegaro:
Added a couple tests for the store and purchaseType null when creating entitlement info, and fixed some casts that were changed by the migration tool, but that we actually don't need.

lib/purchases_flutter.dart
aboedo: that makes sense 👍

lib/purchases_flutter.dart
aboedo: agreed. @vegaro we should do the updates here, even if we're safe for now since we know the current implementation of upgradeInfo won't change values for `prorationMode` in between calls

aboedo (approved):
@vegaro looks great! good to go once we improve safety on the changed calls to `proprationMode`. Thanks so much @IchordeDionysos for the detailed responses to my questions!!!

vegaro:
Sorry I missed that one @aboedo, I actually simplified oldSKU too.

aboedo (approved):

vegaro:
Changed the base. This is meant to be a pre release until the final Dart version is released.

IchordeDionysos:
@vegaro the new Dart version with null-safety was just released 😍 
https://medium.com/dartlang/announcing-dart-2-12-499a6e689c87

So we can probably now release this 💪

vegaro:
I saw that! I am actually working on the new release now :)

vegaro:
@IchordeDionysos Just released 3.0.0

IchordeDionysos:
Very cool 😍  Thanks for keeping the RevenueCat Flutter SDK always so up-to-date 💪

vegaro:
Let us know if you face any issue or have any suggestion. And thanks so much for reviewing this. We really appreciate it

PR #158, opened 2021-03-03
Adds scripts to make faster releases

vegaro:
I added fastlane because I could copy the scripts we are already using in another platforms.


aboedo (approved):
I'd love to find a way to share the fastlane stuff between projects eventually

Issue #163, opened 2021-03-08
getProducts(), type 'List<dynamic>' is not a subtype of type 'FutureOr<List<Product>>'


marchacio:
Hi everyone.
In my app, i have added the method "Purchases.getProducts(...)" but this throws a TypeError Exception.
VScode shows me that the corrupted file is "purchases_flutter.dart".

Today I upgraded the package to 3.0.0; before today, i used 2.0.2 

**My code:**
```
Future<List<Product>> caricaDonazioni() async {
    List<Product> prodotti = await Purchases.getProducts([
      'XXXXX',
      'XXXXX',
      'XXXXX',
      'XXXXX',
      'XXXXX',
    ], type: PurchaseType.inapp);

    prodotti.sort((prod1, prod2) => prod1.price.compareTo(prod2.price));

    return prodotti;
  }
```

vegaro:
Can you share the exception? I would like to look at the stacktrace. Thanks!

davidmartos96:
@vegaro I think that the problem is that the typing of invokeMethod has been removed, so that turns into dynamic.

Dynamic doesn't statically check the .map, but fails to infer the output type as List<Product>.

Probably bring the List<dynamic> back would fix it. Maybe it could help to deactivate the implicit dynamic lint to better spot/avoid these kind lf issues. 
dynamic is too magical 😄

vegaro:
As you say, bringing back `List<dynamic>` as the type fixes it, but also adding `map<Product>` and keeping the `final result`. What you are saying makes sense. For future reference, what's basically happening is:

1. Replacing `List<dynamic>` with `final` is making the result to be of type `dynamic`.
1. The static analyzer doesn't check the type of `map` because result is `dynamic`. We can specifically tell that `map` should be `map<Product>`, or just make sure that `result` is not `dynamic`

I was actually just looking at how to disable the `implicit-dynamic` lint because these issues look insanely hard to spot unless you have a ton of experience with the language, which I don't.

vegaro:
I am preparing a release 3.0.1 that should include the fix. Thanks for reporting and sorry about the trouble

nohli:
> As you say, bringing back `List<dynamic>` as the type fixes it, but also adding `map<Product>` and keeping the `final result`. What you are saying makes sense. For future reference, what's basically happening is:
> 
> 1. Replacing `List<dynamic>` with `final` is making the result to be of type `dynamic`.
[...]

PR #164, opened 2021-03-09
Fix getproducts and checkTrialOrIntroductoryPriceEligibility

vegaro:
Should fix https://github.com/RevenueCat/purchases-flutter/issues/163

I removed some explicit types when preparing the release 3.0.0. These changes broke `getProducts` and `checkTrialOrIntroductoryPriceEligibility`.

Followup tasks on this PR should be:
1. Write tests for all functions, this PR only adds tests for the broken ones.
1. Add rule to linter to disable implicit dynamic. 

I have created Clubhouse tasks for both

### DIFF
--- a/lib/purchases_flutter.dart
+++ b/lib/purchases_flutter.dart
@@ -120,8 +120,7 @@ class Purchases {
   ///
   /// Time is money.
   static Future<Offerings> getOfferings() async {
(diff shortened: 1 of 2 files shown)


vegaro:
@aboedo going to merge this without review because the current release is broken

PR #170, opened 2021-03-11
Adds new Magic Weather sample app

codykerns:
This PR adds the new 'MagicWeather' sample app to a new /Examples folder (for platform consistency).

Also leaves the existing example folder in place.


aboedo (commented):
Looks great! Leaving a couple of questions: 

- do we actually need to have a `lib/main.dart`? I'm not sure if running straight from terminal works without it
- would having it in `Examples` actually be more confusing for developers? we now have to folders: `example` and `Examples`

vegaro:
Sorry I missed this. Looking now

Examples/MagicWeather/lib/src/components/native_dialog.dart
```diff
@@ -0,0 +1,55 @@
...
+    if (!Platform.isIOS) {
+      return AlertDialog(
+        title: new Text(
+          title,
+        ),
+        content: new Text(
+          this.content,
+        ),
+        actions: <Widget>[
+          new FlatButton(
```
vegaro, line 23: There's a warning that says `FlatButton` is deprecated and `TextButton` should be used

Examples/MagicWeather/android/app/src/debug/AndroidManifest.xml
```diff
@@ -0,0 +1,7 @@
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+    package="m4trixdevlimited.com.flutter_magic_weather_app">
```
vegaro, line 2: should we update this package name? in other examples we have `com.revenuecat.sample` if I am not wrong

Examples/MagicWeather/android/app/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,51 @@
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+    package="m4trixdevlimited.com.flutter_magic_weather_app">
```
vegaro, line 2: same question here about the package name

vegaro (commented):
Besides the package name it looks good to me. I am not sure why the lint is complaining. I am going to look into it. It only complains when running `flutter analyze` in the root which is strange

vegaro:
Just fixed the lint job, it was missing the `pub get` in the example folder. Apparently running `pub get` in the root only does pub get in the package and in the current example, but not in Examples/MagicWeather. Anyhow, it shows the only real issue now

aboedo (approved):

PR #184, opened 2021-04-08
Release/3.2.0

aboedo:
## 3.2.0
- Added new Magic Weather sample app, available in `/Examples/`
    https://github.com/RevenueCat/purchases-flutter/pull/170


Examples/MagicWeather/android/app/build.gradle
```diff
@@ -44,7 +44,7 @@ android {
 
     defaultConfig {
         // TODO: Specify your own unique Application ID (https://developer.android.com/studio/build/application-id.html).
-        applicationId "m4trixdevlimited.com.flutter_magic_weather_app"
+        applicationId "com.revenuecat.sampleapp"
```
aboedo, line 47: I missed a couple of these when fixing the package names in #170 🤦 
I tested that it works correctly for both android and ios

aboedo:
I did, `Package has 0 warnings.`. I think it helps that they're in different folders: `example` and `Examples`

aboedo:
but I could see that being confusing, though.

aboedo:
double-checked with `pana` to get the score. it's `110/120` on the command line (weirdly the web version shows a max of 130, where docs account for 20, and in the command line they account for 10). 
In any case, it seems like the score is preserved before and after this PR.

vegaro:
Thanks for checking with pana and for the updates. I am reviewing again!

aboedo:
okay, this underwent some changes:

- moved examples: they're now at `revenuecat_examples`, one is `MagicWeather` and the other one is `purchase_tester`
- updated package names
- updated CI for those changes
- added a readme in `examples` that explains the structure of those examples and the differences between them.

example/README.md
```diff
@@ -1,2 +1,10 @@
-This project is a starting point for a Flutter application with a RevenueCat integration. For help 
-getting started with, view our [online documentation](https://docs.revenuecat.com).
+# Examples
+Our examples are located at `revenuecat_examples/`.
+
+### MagicWeather
+
+This project is the best showcase for how to use `purchases-flutter` for a subscriptions app.
```
vegaro, line 6: I think the right name is `purchases_flutter`

vegaro (approved):
LGTM!

example/README.md
aboedo: good catch, the repo is with `-` but the package is `_`

Issue #162, opened 2021-03-06
Web support


Albert-Jan:
Are there any plans regarding adding web support with a Stripe implementation? This might be extra interesting now Flutter 2.0 promotes Web to the stable channel :P Keep it up, guys!

kulveersingh:
Web support is the most important feature because businesses are pissed off by their 30% cut in payments.

aboedo:
hi! 👋 sorry for the delayed response, I somehow missed this ticket when it was first opened. 

The product team has this in their thoughts. We don't have plans of adding Stripe support to our SDKs in the near future though as we are focusing on Google and Apple IAPs for mobile.

However, using Stripe with RevenueCat on mobile is still possible. You can use our Stripe integration through our REST API ([documented here](https://docs.revenuecat.com/docs/stripe)). 

You would have to set up a server to handle the web payments, and then instead of calling `purchasePackage`, you'd send the Stripe tokens to RevenueCat [as described here](https://docs.revenuecat.com/docs/stripe#5-send-stripe-tokens-to-revenuecat).

Hope that helps. 

We reserve GitHub Issues for bug reports only, so I'm closing this out since it's a feature request.

Albert-Jan:
Hi! I think you misunderstood my original question. You can now build your flutter code to the web. So I was not talking about using stripe on mobile, but accessing a users' entitlements on the web version of my flutter application. I completely understand if adding stripe support is not a top priority on the web, but reading a users' entitlements on a web app would be really really useful. Cheers 😎

aboedo:
@Albert-Jan I see, thanks for the clarification! 

Most of what I said does still apply, though - you can use RevenueCat with Stripe, and it's done through our REST API. If you only need to get user's entitlements for items they buy through a mobile app, then all you need to do is [get the subscriber info](https://docs.revenuecat.com/reference#subscribers). 

If you want to also allow for them to make in-app purchases, then you'd set up a server to handle the web payments and send over the Stripe tokens to RevenueCat after checkout, then consume the REST API to get the updated subscriber information. 

This would also allow users to purchase a subscription through your web app and use it on the mobile versions of your app. 

We're happy to help you get set up with this if you'd like. Just head to our Help Center at https://support.revenuecat.com/ to get started.

PR #191, opened 2021-04-30
Update pubspec.yaml

CoreyRab:
Small tweak to the description on pub.dev.

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -1,5 +1,5 @@
 name: purchases_flutter
-description: A Flutter plugin that makes it simple to build in-app purchases and subscriptions – receipt validation, status tracking, and more. Supports iOS and Android.
+description: A Flutter plugin that makes it simple to add and manage in-app purchases (IAP) and subscriptions. Supports Android, iOS, macOS, iPadOS, and watchOS.
 version: 3.2.2
 homepage: https://www.revenuecat.com/
 repository: https://github.com/RevenueCat/purchases-flutter


aboedo (commented):
we support macOS for flutter as well, right @vegaro?

pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: purchases_flutter
-description: A Flutter plugin that makes it simple to build in-app purchases and subscriptions – receipt validation, status tracking, and more. Supports iOS and Android.
+description: A Flutter plugin that makes it simple to build in-app purchases (IAP) and subscriptions. Supports iOS, Android, and Stripe.
```
vegaro, line 2: Would this lead to developers to think that we support the Web platform? We've had several requests for it https://github.com/RevenueCat/purchases-flutter/issues/162 . @RevenueCat/sdk what do you think?

vegaro:
we do support MacOS @aboedo , we should probably add it yes, good call

pubspec.yaml
aboedo: It might also lead to some confusion since on our pub.dev, Web is marked as not supported, right?

CoreyRab: Why is marked as not supported?

aboedo:
I didn't realize that Flutter had watchOS support, they didn't list it in their platforms https://flutter.dev/docs/development/tools/sdk/release-notes/supported-platforms but they do have articles for it. 
I think we should list it, yes. 
As for iPadOS, happy to go either way - today it may be clear to everyone that everything supported by iOS is supported by iPadOS (since they're mostly the same), but if they start branching out more in the future then it may become less obvious over time and worth mentioning explicitly

vegaro (approved):

PR #193, opened 2021-05-18
Adds canMakePayments method

beylmk:
Will need to update to remove the includeBuilds and point to new purchases-hybrid-common

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -9,7 +9,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.5.2'
+        classpath 'com.android.tools.build:gradle:4.0.1'
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
     }
 }
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -9,7 +9,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:4.0.1'
+        classpath 'com.android.tools.build:gradle:3.5.2'
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
     }
 }
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -38,5 +38,5 @@ android {
 
 dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
-    implementation "com.revenuecat.purchases:purchases-hybrid-common:1.6.2"
+    implementation "com.revenuecat.purchases:purchases-hybrid-common:1.7.0"
 }
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -39,22 +39,4 @@ android {
 dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
     implementation "com.revenuecat.purchases:purchases-hybrid-common:1.6.2"
-
-    implementation('com.revenuecat.purchases:purchases') {
-        version {
-            branch = 'can-make-payments'
-        }
-    }
-
-    implementation('com.revenuecat.purchases:common') {
-        version {
-            branch = 'can-make-payments'
-        }
-    }
-
-    implementation('com.revenuecat.purchases:utils') {
-        version {
-            branch = 'can-make-payments'
-        }
-    }
 }
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -39,4 +39,22 @@ android {
 dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
     implementation "com.revenuecat.purchases:purchases-hybrid-common:1.6.2"
+
+    implementation('com.revenuecat.purchases:purchases') {
+        version {
+            branch = 'can-make-payments'
+        }
+    }
+
+    implementation('com.revenuecat.purchases:common') {
+        version {
+            branch = 'can-make-payments'
+        }
+    }
+
+    implementation('com.revenuecat.purchases:utils') {
+        version {
+            branch = 'can-make-payments'
+        }
+    }
 }
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -38,5 +38,5 @@ android {
 
 dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
-    implementation "com.revenuecat.purchases:purchases-hybrid-common:1.7.0"
+    implementation "com.revenuecat.purchases:purchases-hybrid-common:1.7.1"
 }
--- a/android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
+++ b/android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
@@ -9,14 +9,15 @@ import androidx.annotation.Nullable;
 import com.revenuecat.purchases.PurchaserInfo;
 import com.revenuecat.purchases.Purchases;
 import com.revenuecat.purchases.PurchasesErrorCode;
-import com.revenuecat.purchases.common.CommonKt;
-import com.revenuecat.purchases.common.ErrorContainer;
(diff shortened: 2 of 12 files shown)


android/build.gradle
```diff
@@ -9,7 +9,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.5.2'
+        classpath 'com.android.tools.build:gradle:4.0.1'
```
vegaro, line 12: Can you extract this update, this one https://github.com/RevenueCat/purchases-flutter/pull/193/files#diff-90156c7b6ee416a97411a15877e2e1eac7b5c7c630ead570c421caf5153362d4R8 and https://github.com/RevenueCat/purchases-flutter/pull/193/files#diff-38561c91b2badbeee265859a611c25856b5e46a13aaf19e30d2adb6f859b9745R6 to a separate PR? That way changes are more isolated since this is not really part of the `canMakePayments` support and if they cause issues, it's easier to revert and spot what's causing problems.

vegaro (changes requested):

android/build.gradle
aboedo: I believe this part is needed for tests to pass locally, though, right? 
Agreed with separating it as a PR, but I definitely love that we're finally taking care of this. I've been updating locally and then not committing the changes for a while, which is not a great way to do things.

android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
```diff
@@ -537,6 +542,22 @@ private void collectDeviceIdentifiers(final Result result) {
...
+        CommonKt.canMakePayments(applicationContext,
+                features,
+                new OnResultAny<Boolean>() {
+                    @Override
+                    public void onReceived(Boolean received) {
+                        result.success(received);
+                    }
+
+                    @Override
+                    public void onError(@org.jetbrains.annotations.Nullable ErrorContainer errorContainer) {
```
aboedo, line 555: we're already importing `@Nullable` in line 7, right? can we just use `@Nullable` here? I believe we do that in other places

revenuecat_examples/purchase_tester/android/settings.gradle
```diff
@@ -20,3 +20,8 @@ if (file(".composite-enable").exists()) {
     includeBuild (path)
 }
 
+includeBuild('/Users/madelinebeyl/git/revcat/purchases-hybrid-common/android')
```
aboedo, line 23: oops

android/build.gradle
vegaro: I also confirmed yesterday and it looks like this is the default when creating a new Flutter plugin. So yes, thanks for updating these 🙌🏻

revenuecat_examples/purchase_tester/android/settings.gradle
beylmk: yeah `hybrid-common` needs to be updated, i left this here for now in case yall wanted to test?

ios/Classes/PurchasesFlutterPlugin.m
```diff
@@ -146,6 +146,9 @@ - (void)handleMethodCall:(FlutterMethodCall *)call
         [self setCreative:creative result:result];
     } else if ([@"collectDeviceIdentifiers" isEqualToString:call.method]) {
         [self collectDeviceIdentifiersWithResult:result];
+    } else if ([@"canMakePayments" isEqualToString:call.method]) {
+          NSArray *features = arguments[@"features"];
```
aboedo, line 150: I didn't catch it when we were doing the `purchases-hybrid-common`, but I think best practice here would have been to define the type as `NSArray <NSString *>*`. 
It doesn't really matter for now, though, since the array isn't necessary in the first place.

lib/purchases_flutter.dart
```diff
@@ -584,6 +591,15 @@ enum PurchaseType {
   subs
 }
 
+/// Billing Feature types
+enum BillingFeature {
+  subscriptions,
+  subscriptions_update,
+  in_app_items_on_vr,
+  subscriptions_on_vr,
+  price_change_confirmation
```
aboedo, line 600: convention [for enum values is lowerCamelCase](https://dart.dev/guides/language/effective-dart/style#prefer-using-lowercamelcase-for-constant-names)

revenuecat_examples/purchase_tester/android/settings.gradle
aboedo: oh, ok, that's fine, then. 

I'd usually recommend leaving a `// TODO: remove before merging` or a comment in the PR to make the intent clear

test/purchases_flutter_test.dart
```diff
@@ -82,4 +82,22 @@ void main() {
         ['monthly_intro_pricing_one_week']);
     expect(list.length, 1);
   });
+
+  test('canMakePayments with no params calls successfully', () async {
+    response = true;
+    await Purchases.canMakePayments();
+    expect(log, <Matcher>[
+      isMethodCall('canMakePayments', arguments: {'features': []})
+    ]);
```
aboedo, line 91: 🤔 
I'd also check that the value is passed along correctly, and have the response be random so that if for some reason we make a change that makes the code only compatible with `true`, it gets caught: 

```dart
response = randomGenerator.nextBool();
var receivedCanMakePayments = await Purchases.canMakePayments();
expect(receivedCanMakePayments, response);
```
that way if we later broke [this line](https://github.com/RevenueCat/purchases-flutter/pull/193/files#diff-50113335b625b38d77e98819f82d0437574295a701bcabd0450f59d5f7f12507R522
) tests would fail

aboedo (commented):
looks great so far, left a couple of minor comments

test/purchases_flutter_test.dart
beylmk: i passed a parameter in the second test, were you thinking of something else?

good call on the random 👍

ios/Classes/PurchasesFlutterPlugin.m
beylmk: well that one isn't merged yet either :)

beylmk:
@aboedo @vegaro i think this is ready now! updated to `purchases-hybrid-common` 1.7.0 and removed local build stuff

ios/Classes/PurchasesFlutterPlugin.m
```diff
@@ -435,6 +438,11 @@ - (void)setCreative:(nullable NSString *)creative result:(FlutterResult)result
     result(nil);
 }
 
+- (void)canMakePayments:(NSArray <NSNumber *>*)features result:(FlutterResult)result
```
aboedo, line 441: we should name this one `canMakePaymentsWithFeatures:result:` to follow the obj-c practice of referencing the first param in the method name

aboedo (approved):
looks great! good to go once `purchases-hybrid-common` 1.7.1 is out and we update

android/build.gradle
```diff
@@ -38,5 +38,5 @@ android {
 
 dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
-    implementation "com.revenuecat.purchases:purchases-hybrid-common:1.6.2"
+    implementation "com.revenuecat.purchases:purchases-hybrid-common:1.7.0"
```
aboedo, line 41: we can bump this to 1.7.1 now

aboedo (approved):
looks good to go! just a minor naming update and bump to 1.7.1

beylmk:
@vegaro when you get a chance to re-review :)

revenuecat_examples/purchase_tester/lib/main.dart
```diff
@@ -29,9 +29,14 @@ class _MyAppState extends State<InitialScreen> {
   // Platform messages are asynchronous, so we initialize in an async method.
   Future<void> initPlatformState() async {
     await Purchases.setDebugLogsEnabled(true);
-    await Purchases.setup("api_key");
+    await Purchases.setup("waBUsUKyxopyEZrRZqSLUwAdrnnJWmRv");
```
vegaro, line 32: There's some stuff to be reverted in this file 😃

vegaro (changes requested):

vegaro (approved):

PR #194, opened 2021-05-18
Exclude labels from locking issue

tiinanguyen:
Exclude issues with "enhancement" label from being locked. Introduces "do-not-lock" label if there are any other issues that engineering wants to keep open.


tiinanguyen:
Left as draft for now, want to test if `issue-exclude-labels` means do not lock issues with all the labels or if an issue has one of them

aboedo (approved):

Issue #220, opened 2021-07-22
PlatformException(22, There is an issue with your configuration....


urbanMar:
Affter success transaction with GP I get:
`[   +6 ms] I/flutter (10430): PlatformException(22, There is an issue with your configuration. Check the underlying error for more details., {code: 22, message: There is an issue with your configuration. Check the underlying error for more details., readableErrorCode: ConfigurationError, readable_error_code: ConfigurationError, underlyingErrorMessage: Invalid package name., userCancelled: false}, null)
`

LOGS:
`[+53934 ms] D/[Purchases] - DEBUG(10430): ℹ️ Vending PurchaserInfo from cache.
[   +7 ms] D/[Purchases] - DEBUG(10430): ℹ️ Checking if cache is stale AppInBackground false
[        ] D/[Purchases] - DEBUG(10430): ℹ️ Vending Offerings from cache
[   +4 ms] D/[Purchases] - DEBUG(10430): ℹ️ Checking if cache is stale AppInBackground false
[  +35 ms] I/flutter (10430): 2021-07-22T13:25:27.188315	I	purchas product:	 Product{identifier: ci1yintroa, description: Intro offer for the start of our new app. One year all functionality. , title: camping.info 1 year - Intro Offer (Camping Info Test), price: 25.99, priceString: 25,99 zł, currencyCode: PLN, introductoryPrice: IntroductoryPrice{introPrice: 0.0, introPriceString: 0,00 zł, introPricePeriod: P1W, introPricePeriodUnit: DAY, introPricePeriodNumberOfUnits: 7, introPriceCycles: 1}}
[        ] I/flutter (10430): 2021-07-22T13:25:27.189017	I	purchaserInfo2:	 {}
[+19454 ms] D/[Purchases] - DEBUG(10430): ℹ️ Requesting products from the store with identifiers: ci1yintroa
[  +45 ms] D/[Purchases] - DEBUG(10430): ℹ️ Products request finished for ci1yintroa
[...]

PR #223, opened 2021-07-23
Fix: error codes mismap

aboedo:
adds a couple of error codes that were missing + tests. 
I don't think this fully addresses the issue in #220, since the error displayed there _is_ the correct one for code `19`, so the bug is probably in the assignment of the error code, not the translation of it.

### DIFF
--- a/test/purchases_flutter_test.dart
+++ b/test/purchases_flutter_test.dart
@@ -175,4 +175,32 @@ void main() {
       fail("there was an exception " + e.toString());
     }
   });
+
+  test('errors are mapped correctly', () { 
+    assert(PurchasesErrorCode.unknownError.index == 0);
+    assert(PurchasesErrorCode.purchaseCancelledError.index == 1);
+    assert(PurchasesErrorCode.storeProblemError.index == 2);
+    assert(PurchasesErrorCode.purchaseNotAllowedError.index == 3);
+    assert(PurchasesErrorCode.purchaseInvalidError.index == 4);
+    assert(PurchasesErrorCode.productNotAvailableForPurchaseError.index == 5);
+    assert(PurchasesErrorCode.productAlreadyPurchasedError.index == 6);
+    assert(PurchasesErrorCode.receiptAlreadyInUseError.index == 7);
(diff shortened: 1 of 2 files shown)


test/purchases_flutter_test.dart
```diff
@@ -175,4 +175,60 @@ void main() {
...
+    expect(PurchasesErrorCode.configurationError.index, 23);
+    expect(PurchasesErrorCode.unsupportedError.index, 24);
+  });
+
+  test('PurchasesErrorHelper maps errors correctly', () {
+    expect(PurchasesErrorHelper.getErrorCode(PlatformException(code: '0')), PurchasesErrorCode.unknownError);
+    expect(PurchasesErrorHelper.getErrorCode(PlatformException(code: '1')), PurchasesErrorCode.purchaseCancelledError);
+    expect(PurchasesErrorHelper.getErrorCode(PlatformException(code: '2')), PurchasesErrorCode.storeProblemError);
+    expect(PurchasesErrorHelper.getErrorCode(PlatformException(code: '3')), PurchasesErrorCode.purchaseNotAllowedError);
+    expect(PurchasesErrorHelper.getErrorCode(PlatformException(code: '4')), PurchasesErrorCode.purchaseInvalidError);
```
beylmk, line 212: was wondering if there's any way for us to check that the length of PurchasesErrorCode is equal to the length in the native SDKs, but i don't think we expose those anywhere? just throwing it out there in case you have ideas. could help us catch missing error codes in the future.

beylmk (approved):
⚡

test/purchases_flutter_test.dart
aboedo: I'm really not sure how we could do that, since all calls to the native stuff go through `channel`, and the channel for tests isn't really tied to native code.

I think the closest we might be able to get is to create a test suite for android and one for iOS, then have a test that does an exhaustive switch for all existing cases in the enum, no default. 
if a new case gets added to the enum, the compiler should complain. 

I think that's outside the scope of this PR, but I'll throw something on the backlog for it

Issue #222, opened 2021-07-22
Android: PurchasesErrorHelper.getErrorCode(e) returning wrong error for code 19 (paymentPendingError)


WieFel:
**Describe the bug**
When purchasing a package in my flutter app, i want to handle the case for a pending purchase. Therefore, in the `try`-`catch` block that surrounds my `Purchases.purchasePackage(package)`, I want to check whether the returned error is a `paymentPendingError`. Complete code:
```dart
_purchase(Package package) async {
    try {
      PurchaserInfo purchaserInfo = await Purchases.purchasePackage(package);
      var isPurchased = purchaserInfo
          .entitlements.all[<MY_PRODUCT_ID>].isActive;
      if (isPurchased) {
        // Unlock that great "pro" content
      }
    } on PlatformException catch (e) {
      print(
          "PENDING ERROR: ${e.code} ${e.message} ${PurchasesErrorHelper.getErrorCode(e)}");
      // }
    }
  }
```

When completing a purchase with "Slow test card. Approves after a few minutes" in my Android app, I get the following error from RevenueCat debug log:
```
E/[Purchases] - ERROR(17202): 🤖‼️ The payment is pending.
```

The `print` statement I have added in the `catch` block yields the following:
```
I/flutter (17202): PENDING ERROR: 19 The payment is pending. PurchasesErrorCode.insufficientPermissionsError
```
It yields error code `19` and message `The payment is pending.`, which is correct. However, `PurchasesErrorHelper.getErrorCode(e)` yields the wrong error code `PurchasesErrorCode.insufficientPermissionsError` instead of the expected `PurchasesErrorCode.paymentPendingError`.

I think the mapping from number->error code is probably out of date. When counting through the `PurchasesErrorCode` enum, I get to index `20` for `paymentPendingError`, not `19`...

beylmk:
Hi @WieFel, great catch! You're totally right, our mappings are off by 1 after a certain point due to a previous refactor. Thanks for letting us know...expect a fix very soon.

WieFel:
Thanks @beylmk. Please let me know when the fix is published ;)

aboedo:
Hi @WieFel!! 👋  Thanks for reporting! 
I'm digging in and I made [this pr](#223), but I still haven't been able to reproduce this. 

Could you tell us SDK version you're using?

WieFel:
[...]

PR #225, opened 2021-07-29
Fix: error codes mismap

aboedo:
Addresses #222 

Bumps `purchases-hybrid-common` to `1.8.1` to fix a bug in error mapping in Android.

### DIFF
--- a/android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
+++ b/android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
@@ -314,7 +314,7 @@ public class PurchasesFlutterPlugin implements FlutterPlugin, MethodCallHandler,
             result.success(null);
         } else {
             result.error(
-                    String.valueOf(PurchasesErrorCode.UnknownError.ordinal()),
+                    String.valueOf(PurchasesErrorCode.UnknownError.code),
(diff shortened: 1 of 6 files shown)


android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
```diff
@@ -314,7 +314,7 @@ private void setupPurchases(String apiKey, String appUserID, @Nullable Boolean o
             result.success(null);
         } else {
             result.error(
-                    String.valueOf(PurchasesErrorCode.UnknownError.ordinal()),
+                    String.valueOf(PurchasesErrorCode.UnknownError.code),
```
aboedo, line 317: This particular usage worked correctly because it was index `0`, but it's still a good idea to update.

vegaro: Thanks for checking for usages

vegaro (approved):

android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
aboedo: just tested, the method is actually supposed to be `getCode()` since it gets synthetized for java. updated the usage and tested.

Issue #213, opened 2021-07-02
Recompile with -Xlint:deprecation for details. Warning


Jayvd:
- [ ] I have updated Purchases SDK to the latest version
- [ ] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-flutter/blob/develop/CONTRIBUTING.md)
- [ ] I have read the [Help Center](https://support.revenuecat.com/hc/en-us)
- [ ] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [ ] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-flutter/issues)

I get this message when compiling, any ideas why?

Thank you

```
Note: /Users/jayvd/.pub-cache/hosted/pub.dartlang.org/purchases_flutter-3.3.1/android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java uses or overrides a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
```

aboedo:
Hi @Jayvd 👋 Thanks for reporting! 

You're correct! There are a few deprecated things we're using. 
We'll update those.
<img width="560" alt="image" src="https://user-images.githubusercontent.com/3922667/124312929-87f36100-db46-11eb-9763-203f19c2cf28.png">

Thanks again!

daveshirman:
Is there any rough timeline on when this will be amended please?

The fewer deprecated APIs that are accessed, the less the chance of the app not building in the near future. 

Thanks.

aboedo:
@daveshirman I don't have an exact ETA, however, for this particular case, removing the deprecated API usage isn't really an option:

The deprecations that get referenced by the warning are all related to flutter's v1 plugin format. Since 1.12, Flutter's had a new format for plugins to use, which introduces better handling of the app's lifecycle. 

We [upgraded to the new format back in June 2020](https://github.com/RevenueCat/purchases-flutter/commit/da374dae3c19a843c21cb77c5296d1c48e969655), however, Flutter recommends that we [keep the old methods around](https://flutter.dev/docs/development/packages-and-plugins/plugin-api-migration#upgrade-steps) for apps that use the v1 method of embedding. 

So the best path to follow for this particular case is to suppress the deprecation warnings, and remove the methods once v1 support is removed, which might take a long time. 

[...]

PR #232, opened 2021-08-12
suppress deprecation warnings for types from plugin versions pre-1.12

aboedo:
Flutter released a new format for plugins in version 1.12. 

We [upgraded to the new format back in June 2020](https://github.com/RevenueCat/purchases-flutter/commit/da374dae3c19a843c21cb77c5296d1c48e969655), however, Flutter recommends that we [keep the old methods around](https://flutter.dev/docs/development/packages-and-plugins/plugin-api-migration#upgrade-steps) for apps that use the v1 method of embedding. 

Having those references around produces deprecation warnings, though. So this PR suppresses the deprecation warnings, along with a couple coming from implementations of our own methods that are deprecated. 

This is also what Flutter itself is doing for their plugins, [see this example](https://github.com/flutter/plugins/blob/4383bb15d91384efb22dbe409c3fd58e4552267c/packages/battery/battery/android/src/main/java/io/flutter/plugins/battery/BatteryPlugin.java#L34). 

Addresses #213

### DIFF
--- a/android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
+++ b/android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
@@ -47,6 +47,7 @@ public class PurchasesFlutterPlugin implements FlutterPlugin, MethodCallHandler,
     private static final String PURCHASER_INFO_UPDATED = "Purchases-PurchaserInfoUpdated";
 
     // Only set registrar for v1 embedder.
+    @SuppressWarnings("deprecation")
     private PluginRegistry.Registrar registrar;
     // Only set activity for v2 embedder. Always access activity from getActivity() method.
     @Nullable private Context applicationContext;
@@ -59,6 +60,7 @@ public class PurchasesFlutterPlugin implements FlutterPlugin, MethodCallHandler,
     /**
      * Plugin registration.
      */
+    @SuppressWarnings("deprecation")
     public static void registerWith(Registrar registrar) {
         PurchasesFlutterPlugin instance = new PurchasesFlutterPlugin();
         instance.onAttachedToEngine(registrar.messenger(), registrar.context());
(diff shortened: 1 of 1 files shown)


taquitos:
I think it might make sense to add some logging in order to warn folks that it's deprecated.

android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
```diff
@@ -416,10 +419,12 @@ private void restoreTransactions(final Result result) {
         CommonKt.restoreTransactions(getOnResult(result));
     }
 
+    @SuppressWarnings("deprecation")
     private void reset(final Result result) {
```
beylmk, line 423: should we also mark these methods as deprecated for flutter users? or does that bubble up from hybrid-common already?

aboedo: they're [already marked as deprecated](https://github.com/RevenueCat/purchases-flutter/blob/main/lib/purchases_flutter.dart#L294), this is just our own usage of the deprecated methods in the implementation

aboedo:
@taquitos for deprecations coming from pre-1.12 plugins, developers would already get warnings in their code, coming in from Flutter's deprecation itself, since they have to reference the old types (see [migration guide here](https://github.com/flutter/flutter/wiki/Upgrading-pre-1.12-Android-projects))

For deprecations coming from our own calls to our deprecated methods, folks shouldn't get a warning since that's not something they can actually change. What they _can_ do is to make sure they're not referencing our deprecated methods, which is done at the dart level, and is already taken care of with our deprecation warnings. 

We normally don't issue out explicit logs for these since the build system / IDE would take care of it. 

@taquitos @beylmk do you feel like it's worth adding explicit logs at runtime for deprecations?

taquitos:
I didn't realize the IDE would emit warnings even without these deprecations suppressed. As long as folks get the message that they should migrate when they can, that's all I'm really concerned about, and that sounds like it will happen.

beylmk (approved):
👍

PR #235, opened 2021-08-16
fix: more deprecation warnings

aboedo:
another stab at https://github.com/RevenueCat/purchases-flutter/issues/213

I checked with a local project and verified that we're still getting warnings, caused by the references to deprecated types in the imports. 

Following the example from [this project from Flutter,](https://github.com/flutter/plugins/blob/4383bb15d91384efb22dbe409c3fd58e4552267c/packages/battery/battery/android/src/main/java/io/flutter/plugins/battery/BatteryPlugin.java#L34) I removed the imports and used the full path to the types so they get covered by the `@SuppressWarnings("deprecation")`


beylmk (approved):
¯\_(ツ)_/¯

PR #259, opened 2021-10-10
Remove excluded architectures from simulator build

n8han:
The removed instruction was preventing me from building for the
simulator on an x86 mac.

```
    ld: warning: Could not find or use auto-linked framework 'PurchasesCoreSwift'
    Undefined symbols for architecture arm64:
      "_OBJC_CLASS_$_RCCommonFunctionality", referenced from:
          objc-class-ref in PurchasesFlutterPlugin.o
      "_OBJC_CLASS_$_RCPurchases", referenced from:
          objc-class-ref in PurchasesFlutterPlugin.o
    ld: symbol(s) not found for architecture arm64
```

My own Podfile requires a different and apparently conflicting
instruction:

```rb
post_install do |installer|
  installer.pods_project.build_configurations.each do |config|
    config.build_settings["EXCLUDED_ARCHS[sdk=iphonesimulator*]"] = "arm64"
  end
end
```

It's sad that we need these hacks at all, but it's better to keep the
hack outside the framework since your downstream users may need a
different hack depending on their system architecture.

### DIFF
--- a/ios/purchases_flutter.podspec
+++ b/ios/purchases_flutter.podspec
@@ -19,10 +19,8 @@ Pod::Spec.new do |s|
   s.ios.deployment_target = '9.0'
   s.swift_version         = '5.0'
 
-  # Flutter.framework does not contain a i386 slice.
-  s.pod_target_xcconfig = { 
+  s.pod_target_xcconfig = {
     'DEFINES_MODULE' => 'YES',
-    'EXCLUDED_ARCHS[sdk=iphonesimulator*]' => 'i386',
     'OTHER_LDFLAGS' => '-framework Purchases'
    }
 


NachoSoto:
I forget the exact context for this, but I seem to remember that this exclusion was to handle some older Xcode issue.
I think it's no longer really necessary to do this?

joshdholtz:
A little bit of research on this and found some history on this...

1. `EXCLUDED_ARCHS` added here 👉  https://github.com/flutter/flutter/pull/64504
2. Changed from `VALID_ARCHS` here 👉 https://github.com/flutter/flutter/pull/41828
3. Originally added because of a CocoaPods suggestion 👉  https://github.com/CocoaPods/CocoaPods/pull/8159

It seems like this was originally added in to "prevent tests from running on the i386 simulator" because `Flutter.framework` doesn't support i386. 

@n8han But... I want to figure out why this is failing for you and try to replicate it on my Mac if possible. So I have a few questions for you if you don't mine answering 😊 

1. Your Mac is an x86_64, correct?
2. Are you setting `Build Active Architecture Only` / `ONLY_ACTIVE_ARCH` to `YES` or `NO`?
3. Kind of related to ^ but is this happening when building for debug, test, and/or release?
4. Which iOS versions are you supporting?

Thanks for taking to answer! Just want to make sure this PR gets resolved 🙃

vegaro:
Something we can test is to create a new Flutter plugin from scratch and check how the tooling creates the Xcode project.

vegaro:
@n8han I know it has been a while, but can you share the Podfile of your app? Do you have the following in it:

```
post_install do |installer|
  installer.pods_project.targets.each do |target|
    flutter_additional_ios_build_settings(target)
  end
end
```

This is added by default when creating a new project in the latest flutter version. You can check what `flutter_additional_ios_build_settings` does [here](https://github.com/flutter/flutter/blob/a618ca27a76faa546fe7704215e49b2a871e56db/packages/flutter_tools/bin/podhelper.rb#L33)

aboedo:
Thanks for opening up this PR.
It’s been a while, and we haven’t had any reports of other users running into this issue.

There’s a strong chance that this has been fixed in an Xcode update, and we haven’t been able to reproduce this.
We’re tentatively closing this PR, but feel free to reopen if you’re still experiencing this issue and would like to discuss it again.

For the time being, it looks like the changes here aren’t needed for the SDK to build with the latest builds of Flutter and Xcode.
@n8han Thanks again for your contribution, and please let us know if you have any questions.

Issue #300, opened 2022-01-18
_InternalLinkedHashMap<Object?, Object?>' is not a subtype of type 'Map<String, dynamic> error while purchasePackage() called


abeyshiferaw0:
- [ ] I have updated Purchases SDK to the latest version
- [ ] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-flutter/blob/main/CONTRIBUTING.md)
- [ ] I have searched the [Community](https://community.revenuecat.com)
- [ ] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [ ] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-flutter/issues)

**Describe the bug**
calling purchasePackage() would throw this error after subscription purchase is complete

> _InternalLinkedHashMap<Object?, Object?>' is not a subtype of type 'Map<String, dynamic>

1. Environment
   1. Output of `flutter doctor` `[✓] Flutter (Channel stable, 2.8.0, on macOS 11.5.2 20G95 darwin-x64, locale
    en-ET)
[✓] Android toolchain - develop for Android devices (Android SDK version 31.0.0)
[!] Xcode - develop for iOS and macOS (Xcode 12.5.1)
    ! Flutter recommends a minimum Xcode version of 13.0.0.
      Download the latest version or update via the Mac App Store.
[✓] Chrome - develop for the web
[✓] Android Studio (version 2020.3)
[✓] VS Code (version 1.63.2)
[✓] Connected device (2 available)`
   2. How widespread is the issue. Percentage of devices affected
   > I only tested on Android.

2. [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue
> Issue is not caught by package but error thrown is  type '_InternalLinkedHashMap<Object?, Object?>' is not a subtype of type 'Map<String, dynamic>'

3. Steps to reproduce, with a description of expected vs. actual behavior
call purchasePackage method with subscription packages should reproduce the error
`    PurchaserInfo purchaserInfo = await Purchases.purchasePackage(
      offering.availablePackages.first,
    );`

4. **Other information** (e.g. stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc.)

`    final prorationMode = upgradeInfo?.prorationMode;
    final response = await _channel.invokeMethod('purchasePackage', {
      'packageIdentifier': packageToPurchase.identifier,
[...]

PR #270, opened 2021-11-05
Code refactoring + more safety with freezed

devasidmi:
- [x] Code improvements + more safety with hashcodes & json parsing with Freezed package

  - [x] No PR or issue number

  - [x] image

Single test failed because of wrong API key

PR #301, opened 2022-01-18
Fix wrong type purchase package purchase product

vegaro:
[sc-12529]

Fixes https://github.com/RevenueCat/purchases-flutter/issues/300

This line https://github.com/RevenueCat/purchases-flutter/pull/270/files#diff-2a88ebe6bbbad2ced268a39ea09b1e2d0553678f4e1e43bc8ed79a31c3b3921bL73 changed from `Map<dynamic, dynamic>` to `Map<String, dynamic>`.

It would have been caught by tests, but it's a shame we didn't have tests for such crucial functions. I added some tests for the fixed functions and will follow up with a PR making sure all functions are tested to prevent this from happening again. I want to release ASAP.


lib/models/discount.freezed.dart
```diff
@@ -1,5 +1,6 @@
 // coverage:ignore-file
 // GENERATED CODE - DO NOT MODIFY BY HAND
+// ignore_for_file: type=lint
```
vegaro, line 3: These `.freezed.dart` were not updated to the latest version. I am going to change CI to make sure these are updated everytime something gets merged to `main`.

lib/purchases_flutter.dart
```diff
@@ -190,13 +190,15 @@ class Purchases {
     PurchaseType type = PurchaseType.subs,
   }) async {
     final prorationMode = upgradeInfo?.prorationMode;
-    final response = await _channel.invokeMethod('purchaseProduct', {
+    final purchaserInfoJson =
+        await _invokeMethodReturningPurchaserInfoJsonFromMap(
```
vegaro, line 194: Better names accepted

lib/purchases_flutter.dart
```diff
@@ -628,6 +635,32 @@ class Purchases {
   /// Android only. Call close when you are done with this instance of Purchases to disconnect
   /// from the billing services and clean up resources
   static Future<void> close() => _channel.invokeMethod('close');
+
+  static Future<Map<String, dynamic>>
+      _invokeMethodReturningPurchaserInfoJsonFromMap(
```
vegaro, line 640: Better names accepted

lib/purchases_flutter.dart
```diff
@@ -628,6 +635,32 @@ class Purchases {
...
+    String method,
+    Map<String, Object?> arguments,
+  ) async {
+    final response = await _invokeMethodReturningMap(
+      method,
+      arguments,
+    );
+    final purchaserInfoJson =
+        Map<String, dynamic>.from(response['purchaserInfo']);
```
vegaro, line 649: We were missing this `Map<String, dynamic>.`

lib/purchases_flutter.dart
```diff
@@ -628,6 +636,30 @@ class Purchases {
   /// Android only. Call close when you are done with this instance of Purchases to disconnect
   /// from the billing services and clean up resources
   static Future<void> close() => _channel.invokeMethod('close');
+
+  static Future<Map<String, dynamic>>
+      _invokeMethodReturningPurchaserInfoJsonFromMap(String method,
+          // ignore: require_trailing_commas
+          [dynamic arguments]) async {
+    final response = await _invokeMethodReturningMap(
```
aboedo, line 644: perhaps `_invokeReturningMap` already conveys the same?

lib/models/discount.freezed.dart
```diff
@@ -275,37 +276,31 @@ class _$_Discount implements _Discount {
...
-                const DeepCollectionEquality().equals(other.period, period)) &&
-            (identical(other.periodUnit, periodUnit) ||
-                const DeepCollectionEquality()
-                    .equals(other.periodUnit, periodUnit)) &&
-            (identical(other.periodNumberOfUnits, periodNumberOfUnits) ||
-                const DeepCollectionEquality()
-                    .equals(other.periodNumberOfUnits, periodNumberOfUnits)));
+        (other.runtimeType == runtimeType &&
+            other is _Discount &&
+            const DeepCollectionEquality()
```
beylmk, line 281: was the idential() call no longer necessary?

lib/purchases_flutter.dart
```diff
@@ -628,6 +636,30 @@ class Purchases {
   /// Android only. Call close when you are done with this instance of Purchases to disconnect
   /// from the billing services and clean up resources
   static Future<void> close() => _channel.invokeMethod('close');
+
+  static Future<Map<String, dynamic>>
+      _invokeMethodReturningPurchaserInfoJsonFromMap(String method,
```
aboedo, line 641: perhaps `_invokeReturningPurchaserInfoJson` already conveys the same meaning?

lib/purchases_flutter.dart
```diff
@@ -239,11 +243,13 @@ class Purchases {
...
   ) async {
-    final response = await _channel.invokeMethod('purchaseProduct', {
+    final purchaserInfoJson =
+        await _invokeMethodReturningPurchaserInfoJsonFromMap(
+            'purchaseProduct', {
       'productIdentifier': product.identifier,
       'signedDiscountTimestamp': discount.timestamp.toString()
     });
-    return PurchaserInfo.fromJson(response['purchaserInfo']);
+    return PurchaserInfo.fromJson(purchaserInfoJson);
```
aboedo, line 252: ❤️

aboedo (approved):
looks great! left minor suggestions, they can go into a follow-up PR too

lib/models/discount.freezed.dart
vegaro: This code is autogenerated and we forgot to generate it when increasing the version of freezed in 3.9.1

aboedo: I guess not? the `.freezed` files are all auto-generated by freezed

lib/purchases_flutter.dart
```diff
@@ -628,6 +636,30 @@ class Purchases {
...
+  }
+
+  static Future<Map<String, dynamic>> _invokeMethodReturningMap(String method,
+      // ignore: require_trailing_commas
+      [dynamic arguments]) async {
+    final result = await _channel.invokeMethod<Map<dynamic, dynamic>>(
+      method,
+      arguments,
+    );
+    final response = Map<String, dynamic>.from(result!);
```
beylmk, line 660: maybe we can log if result is nil?

lib/purchases_flutter.dart
beylmk: could just be `invokeMethodReturningPurchaserInfoJson`. but maybe we can also just call `invokeMethodReturningMap` here, and then have a method `getPurchaserInfoJsonFromMap` or something, to separate out the two pieces.

lib/models/discount.freezed.dart
beylmk: 👍

lib/models/discount.freezed.dart
vegaro: https://github.com/RevenueCat/purchases-flutter/pull/302 😬

aboedo: nice!

lib/purchases_flutter.dart
vegaro: This will crash, and I think it should crash because something is very wrong if there's no result, so I don't think logging is necessary

lib/purchases_flutter.dart
vegaro: I liked the idea of extracting `_getPurchaserInfoJsonFromMap`. Checkout the latest commit

vegaro:
@aboedo @beylmk I renamed the functions and extracted the part where we get the Json from the Map. Let me know what you think

aboedo (approved):
❤️ it

beylmk (approved):
⚡

lib/purchases_flutter.dart
beylmk: yeah i agree it should crash, just seems like we sometimes don't get the most descriptive error messages. i suppose using the ! should give us a clear message though

Issue #309, opened 2022-01-27
IntroductoryPrice is no longer exposed by default


mrgnhnt96:
**Describe the bug**

In a previous version of [purcahses_flutter](https://github.com/RevenueCat/purchases-flutter/releases/tag/3.8.0) the `IntroductoryPrice` class was located with the product_wrapper file. Since then, it as been moved into its own file. The object_wrapper barrel file hasn't been updated to include the new introductory_price file. 

The only way to use the IntroductoryPrice object is by importing the implementation file.

beylmk:
@mrgnhnt96 i'm going to do a quick release with this fix, but the team is thinking through ways to automate/test this to prevent it from happening again.

beylmk:
@mrgnhnt96 this should be fixed via [3.9.3](https://github.com/RevenueCat/purchases-flutter/releases/tag/3.9.3), closing for now! feel free to re-open if there are any more problems

PR #310, opened 2022-01-27
feat: expose introductory price

mrgnhnt96:
The IntroductoryPrice object is no longer exported by the library. The only way to use the object is by importing the implementation file.

fixes issue #309


beylmk (approved):
thanks for catching this and opening a PR! 👏

Issue #318, opened 2022-02-10
Breaking change not treated as breaking change


Lootwig:
Somewhere between 3.8.0 and 3.9.3, the "IntroductoryPrice.periodUnit" field was re-typed from `String` to the `PeriodUnit` enum, and the entire class moved into a different location.

This is a **breaking change** which breaks builds without updating code and should be reflected by

a) [bumping the major version of the package](https://semver.org/) and
b) [highlighting this in the changelog](https://pub.dev/packages/purchases_flutter/changelog)

joshdholtz:
@Lootwig My apologies on this! It looks like this was missed when considering the version bump and changelog. We 100% should bumped the version or called it out in the changelog. 

It appears this happened in https://github.com/RevenueCat/purchases-flutter/pull/270 which was included in `3.9.1`. Since its too late to update the changelog, I'll make sure that we will update the changelog appropriately. 

I do want to through the diff to make sure there aren't any other types that we missed that also need calling out in the changelog.

I appreciate you letting us know and we will get this updated as soon as we can! Thank you 😊

beylmk:
Hey @Lootwig ignore the bot! We are in the process of releasing the fix. I'll post back here when that's done!

beylmk:
Hey @Lootwig all fields should be added back in https://github.com/RevenueCat/purchases-flutter/releases/tag/3.9.4! I'm going to close this ticket for now, but feel free to re-open if you notice anything else. Thanks again for raising this to our attention.

PR #319, opened 2022-02-11
Add back IntroductoryPrice.introPricePeriodUnit string

beylmk:
In https://github.com/RevenueCat/purchases-flutter/pull/270/files#diff-31ad10b1074cdc6e6c82c3cc2ab3221a117afbaaee017193b6b189d7f34bc769R40 we changed periodUnit from a string to a PeriodUnit enum. An [issue](https://github.com/RevenueCat/purchases-flutter/issues/318) was opened indicating we released this change without reflecting it in the changelog and did a minor instead of a major.

This fix adds back the string type and deprecates it. We can leave the PeriodUnit enum type since the properties have different names. 

Also:
* added unit tests to confirm the IntroductoryPrice values are set properly from the original JSON
*  added a little "api test" method to our current sample app which simply accesses each property, ensuring the types are correct. This issue highlights our need for a full API tester in purchases-flutter, but this is a quicker api test
* created a [follow-up ticket](https://app.shortcut.com/revenuecat/story/13290/remove-intropriceperiodunit) for eventually removing introPricePeriodUnit, since some of the changes required aren't obviously attributed to the property

We will release a 3.9.4 with this fix.

[sc-13246]

### DIFF
--- a/lib/models/introductory_price.dart
+++ b/lib/models/introductory_price.dart
@@ -42,7 +42,8 @@ class IntroductoryPrice with _$IntroductoryPrice {
     @JsonKey(name: 'periodNumberOfUnits') int periodNumberOfUnits,
   ) = _IntroductoryPrice;
 
-  /// Maps introPricePeriodUnit string to PeriodUnit enum type
+  /// Unit for the billing period of the introductory price, can be DAY, WEEK,
+  /// MONTH or YEAR.
   PeriodUnit get periodUnit {
     // ignore: deprecated_member_use_from_same_package
     switch (introPricePeriodUnit) {
--- a/lib/models/introductory_price.dart
+++ b/lib/models/introductory_price.dart
@@ -55,6 +55,29 @@ class IntroductoryPrice with _$IntroductoryPrice {
     }
   }
 
+  /// Introductory price of a subscription in the local currency.
+  @Deprecated('Use price instead.')
+  double get introPrice => price;
+
+  /// Formatted introductory price of a subscription, including
+  /// its currency sign, such as €3.99.
+  /// @Deprecated('Use priceString instead.')
+  String get introPriceString => priceString;
+
+  /// Billing period of the introductory price, specified in
+  /// ISO 8601 format.
+  @Deprecated('Use period instead.')
+  String get introPricePeriod => period;
+
+  /// Number of units for the billing period of the introductory price.
+  @Deprecated('Use periodNumberOfUnits instead.')
+  int get introPricePeriodNumberOfUnits => periodNumberOfUnits;
+
+  /// Number of subscription billing periods for which the
+  /// user will be given the introductory price, such as 3.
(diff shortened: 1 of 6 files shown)


lib/models/introductory_price.dart
```diff
@@ -19,6 +15,8 @@ enum PeriodUnit {
 
 /// Contains all the introductory information associated with a [Product]
 class IntroductoryPrice with _$IntroductoryPrice {
+  const IntroductoryPrice._();
```
beylmk, line 18: needed to add this to enable the custom getter of the PeriodUnit type. see https://github.com/rrousselGit/freezed#custom-getters-and-methods

lib/models/introductory_price.dart
```diff
@@ -35,15 +33,28 @@ class IntroductoryPrice with _$IntroductoryPrice {
...
     @JsonKey(name: 'cycles') int cycles,
 
-    /// Unit for the billing period of the introductory price, can be DAY, WEEK,
-    /// MONTH or YEAR.
-    @JsonKey(name: 'periodUnit', unknownEnumValue: PeriodUnit.unknown)
-        PeriodUnit periodUnit,
+    /// String representation of unit for the billing period of the introductory
+    /// price, can be DAY, WEEK, MONTH or YEAR.
+    @Deprecated('Use periodUnit property of type PeriodUnit instead.')
+    @JsonKey(name: 'periodUnit') String introPricePeriodUnit,
```
beylmk, line 39: couldn't have two properties with same JsonKey, so i pulled the enum out

revenuecat_examples/purchase_tester/lib/main.dart
```diff
@@ -127,6 +132,21 @@ class _UpsellScreenState extends State<UpsellScreen> {
       ),
     );
   }
+
+  void apiTestIntroductoryPrice(IntroductoryPrice introPrice) {
```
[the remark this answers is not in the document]
vegaro: I think we need to add all of them back since they are breaking changes right?

beylmk:
@vegaro ready for re-review

lib/models/introductory_price.dart
```diff
@@ -35,15 +33,51 @@ class IntroductoryPrice with _$IntroductoryPrice {
...
+    }
+  }
+
+  /// Introductory price of a subscription in the local currency.
+  @Deprecated('Use price instead.')
+  double get introPrice => price;
+
+  /// Formatted introductory price of a subscription, including
+  /// its currency sign, such as €3.99.
+  /// @Deprecated('Use priceString instead.')
```
vegaro, line 64:
```suggestion
  @Deprecated('Use priceString instead.')
```

vegaro (approved):

lib/models/introductory_price.dart
beylmk: good catch

PR #327, opened 2022-02-18
Fix changelogs

beylmk:
These changes are already made in the Github releases, just wanted to be sure they're updated everywhere

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 ## 3.9.4
 
 - Add back IntroductoryPrice fields that were accidentally removed without a major release (https://github.com/RevenueCat/purchases-flutter/pull/319)
+- Fields added back:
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -29,6 +35,12 @@
     https://github.com/RevenueCat/purchases-flutter/pull/285
 - Updated JSON serialization using Freezed
     https://github.com/RevenueCat/purchases-flutter/pull/270
+- Breaking changes:
```
aboedo, line 38: Perhaps we can add a note for full transparency here, explaining that this was unintended and fixed in 3.9.4? 
Just so it doesn't look like we _intended_ to make breaking changes in a hotfix

beylmk: sure, good call

beylmk: fixed in this pr and in the gh release..

aboedo (approved):

PR #336, opened 2022-03-09
Add Amazon consumer proguard rules to plugin and prepare alpha.3 release

vegaro:
[CF-236]

Adds consumer proguard rules so they get added automatically to apps. Also prepares alpha.3 version.

There have been several reports of installation problems in Flutter due to missing Proguard rules. We need to do a release on top of the latest `amazon.alpha.2` version that adds these rules automatically. 

When we are ready to update purchases-flutter to Android v5, these will get automatically added via purchases-android but until then, let's add them to our purchases-flutter plugin.

[CF-236]: https://revenuecats.atlassian.net/browse/CF-236?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


taquitos (approved):
:shipit:

PR #337, opened 2022-03-11
add missing parameter when calling setup with amazon

aboedo:
We had the parameter in the method signature but not at the callsite


vegaro:
🤦

vegaro (approved):

PR #344, opened 2022-04-12
[CF-459] Flutter deferred purchases

beylmk:
Adds functionality for promoted purchases 

I had trouble unit testing this, since we don't currently have a way of mocking the ios -> flutter method channel communication... and the method we could test in purchases_flutter is the _startPromotedProductPurchase, but that's private. Open to any ideas. We don't have any tests for the purchaserInfoUpdatedListener either, not sure if anyone has context on whether we tried for that?

To manually test in purchase tester, use:
itms-services://?action=purchaseIntent&bundleId=com.revenuecat.sampleapp&productIdentifier=com.revenuecat.monthly_4.99.1_week_intro

One thing I've noticed that might be broken/could be improved (though probably out of the scope of this PR): If you've already purchased the product, purchases-hybrid-common still returns a purchaserInfo and productIdentifier. There's no error or indication that the purchase didn't successfully go through. I am going to investigate whether that's information we can pass along somehow...

### DIFF
--- a/ios/Classes/PurchasesFlutterPlugin.m
+++ b/ios/Classes/PurchasesFlutterPlugin.m
@@ -511,7 +511,7 @@ signedDiscountTimestamp:(nullable NSString *)discountTimestamp
     [self.defermentBlocks addObject:makeDeferredPurchase];
     NSInteger position = [self.defermentBlocks count] - 1;
     [self.channel invokeMethod:PurchasesStartDeferredPurchaseEvent
-                     arguments:[NSNumber numberWithInteger:position]];
+                     arguments:@{@"callbackID": [NSNumber numberWithInteger:position]}];
 }
 
 #pragma mark -
--- a/ios/Classes/PurchasesFlutterPlugin.m
+++ b/ios/Classes/PurchasesFlutterPlugin.m
@@ -172,9 +172,9 @@ NSString *PurchasesReadyForPromotedProductPurchaseEvent = @"Purchases-ReadyForPr
         [self paymentDiscountForProductIdentifier:productIdentifier
                                discountIdentifier:discountIdentifier
                                            result:result];
-    } else if ([@"startDeferredPurchase" isEqualToString:call.method]) {
+    } else if ([@"startPromotedProductPurchase" isEqualToString:call.method]) {
         NSNumber *callbackID = arguments[@"callbackID"];
-        [self startDeferredPurchase:callbackID
+        [self startPromotedProductPurchase:callbackID
                             result:result];
     } else if ([@"close" isEqualToString:call.method]) {
         [self closeWithResult:result];
@@ -477,18 +477,11 @@ signedDiscountTimestamp:(nullable NSString *)discountTimestamp
                                                }];
 }
 
-- (void)startDeferredPurchase:(NSNumber *)callbackID
+- (void)startPromotedProductPurchase:(NSNumber *)callbackID
                       result:(FlutterResult)result {
     RCDeferredPromotionalPurchaseBlock makePurchaseBlock = [self.startPurchaseBlocks objectAtIndex:[callbackID integerValue]];
     [RCCommonFunctionality makeDeferredPurchase:makePurchaseBlock
-                                completionBlock:^(NSDictionary * _Nullable responseDictionary,
-                                                  RCErrorContainer * _Nullable error) {
-                                                    if (error) {
-                                                        [self rejectWithResult:result error:error];
-                                                    } else {
-                                                        result(responseDictionary);
-                                                    }
-                                                }];;
+                                completionBlock:[self getResponseCompletionBlock:result]];
 }
 
 - (void)closeWithResult:(FlutterResult)result {
--- a/ios/Classes/PurchasesFlutterPlugin.m
+++ b/ios/Classes/PurchasesFlutterPlugin.m
@@ -9,11 +9,13 @@
 
 @property (nonatomic, retain) FlutterMethodChannel *channel;
 @property (nonatomic, retain) NSObject <FlutterPluginRegistrar> *registrar;
+@property (nonatomic, retain) NSMutableArray<RCDeferredPromotionalPurchaseBlock> *defermentBlocks;
 
 @end
 
 
-NSString *RNPurchasesPurchaserInfoUpdatedEvent = @"Purchases-PurchaserInfoUpdated";
+NSString *PurchasesPurchaserInfoUpdatedEvent = @"Purchases-PurchaserInfoUpdated";
+NSString *PurchasesStartDeferredPurchaseEvent = @"Purchases-MakeDeferredPurchase";
 
 
 @implementation PurchasesFlutterPlugin
@@ -170,6 +172,10 @@ NSString *RNPurchasesPurchaserInfoUpdatedEvent = @"Purchases-PurchaserInfoUpdate
         [self paymentDiscountForProductIdentifier:productIdentifier
                                discountIdentifier:discountIdentifier
                                            result:result];
+    } else if ([@"makeDeferredPurchase" isEqualToString:call.method]) {
+        NSNumber *callbackID = arguments[@"callbackID"];
+        [self makeDeferredPurchase:callbackID
+                            result:result];
     } else if ([@"close" isEqualToString:call.method]) {
         [self closeWithResult:result];
     } else {
@@ -471,6 +477,20 @@ signedDiscountTimestamp:(nullable NSString *)discountTimestamp
                                                }];
 }
 
+- (void)makeDeferredPurchase:(NSNumber *)callbackID
+                      result:(FlutterResult)result {
+    RCDeferredPromotionalPurchaseBlock defermentBlock = [self.defermentBlocks objectAtIndex:[callbackID integerValue]];
+    [RCCommonFunctionality makeDeferredPurchase:defermentBlock
+                                completionBlock:^(NSDictionary * _Nullable responseDictionary,
+                                                  RCErrorContainer * _Nullable error) {
+                                                    if (error) {
+                                                        [self rejectWithResult:result error:error];
+                                                    } else {
+                                                        result(responseDictionary);
+                                                    }
+                                                }];;
+}
+
 - (void)closeWithResult:(FlutterResult)result {
     result(nil);
 }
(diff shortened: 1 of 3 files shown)


beylmk:
@vegaro check out the description, wondering if you have context on the unit testing and "already purchased" situations...

ios/Classes/PurchasesFlutterPlugin.m
```diff
@@ -170,6 +172,10 @@ - (void)handleMethodCall:(FlutterMethodCall *)call
         [self paymentDiscountForProductIdentifier:productIdentifier
                                discountIdentifier:discountIdentifier
                                            result:result];
+    } else if ([@"startPromotedProductPurchase" isEqualToString:call.method]) {
+        NSNumber *callbackID = arguments[@"callbackID"];
+        [self startPromotedProductPurchase:callbackID
+                            result:result];
```
aboedo, line 178: indentation is off here, the `:`s should match

ios/Classes/PurchasesFlutterPlugin.m
```diff
@@ -471,6 +477,13 @@ - (void)paymentDiscountForProductIdentifier:(NSString *)productIdentifier
                                                }];
 }
 
+- (void)startPromotedProductPurchase:(NSNumber *)callbackID
+                      result:(FlutterResult)result {
+    RCDeferredPromotionalPurchaseBlock makePurchaseBlock = [self.startPurchaseBlocks objectAtIndex:[callbackID integerValue]];
```
aboedo, line 482: we should add line breaks here for readability

ios/Classes/PurchasesFlutterPlugin.m
```diff
@@ -471,6 +477,13 @@ - (void)paymentDiscountForProductIdentifier:(NSString *)productIdentifier
                                                }];
 }
 
+- (void)startPromotedProductPurchase:(NSNumber *)callbackID
+                      result:(FlutterResult)result {
```
aboedo, line 481: `:`s should match

ios/Classes/PurchasesFlutterPlugin.m
```diff
@@ -479,10 +492,23 @@ - (void)closeWithResult:(FlutterResult)result {
 #pragma mark Delegate Methods
 
 - (void)purchases:(RCPurchases *)purchases didReceiveUpdatedPurchaserInfo:(RCPurchaserInfo *)purchaserInfo {
-    [self.channel invokeMethod:RNPurchasesPurchaserInfoUpdatedEvent
+    [self.channel invokeMethod:PurchasesPurchaserInfoUpdatedEvent
                      arguments:purchaserInfo.dictionary];
 }
 
+-(void) purchases:(RCPurchases *)purchases shouldPurchasePromoProduct:(SKProduct *)product defermentBlock:(RCDeferredPromotionalPurchaseBlock)makeDeferredPurchase {
```
aboedo, line 499: nits: 
- extra space between `(void)` and `purchases:`
- line breaks between params

ios/Classes/PurchasesFlutterPlugin.m
```diff
@@ -479,10 +492,23 @@ - (void)closeWithResult:(FlutterResult)result {
...
+    if (!self.startPurchaseBlocks) {
+            self.startPurchaseBlocks = [NSMutableArray array];
```
aboedo, line 501: indentation here is off as well

ios/Classes/PurchasesFlutterPlugin.m
```diff
@@ -479,10 +492,23 @@ - (void)closeWithResult:(FlutterResult)result {
...
+    }
+
+    [self.startPurchaseBlocks addObject:makeDeferredPurchase];
+    NSInteger position = [self.startPurchaseBlocks count] - 1;
+    [self.channel invokeMethod:PurchasesReadyForPromotedProductPurchaseEvent
+                     arguments:@{@"callbackID": [NSNumber numberWithInteger:position],
```
aboedo, line 507: I think you can replace the `[NSNumber numberWithInteger:position]` with simply `@(position)`

lib/purchases_flutter.dart
```diff
@@ -13,11 +13,36 @@ typedef PurchaserInfoUpdateListener = void Function(
...
+/// ``Purchases`` will not proceed with the promoted purchase.
+///
+/// This can be tested by opening a link like:
+/// itms-services://?action=purchaseIntent&bundleId=<BUNDLE_ID>&productIdentifier=<SKPRODUCT_ID>
+///
+/// - Parameter productIdentifier: the productIdentifier associated with the promoted purchase
+/// - Parameter startPurchase: call this block when the app is ready to handle the purchase
+///
+/// ### Related Articles:
+/// - [Apple Documentation](https://rev.cat/testing-promoted-in-app-purchases)
```
aboedo, line 32: ❤️

aboedo (commented):
do we still need a noop handler on the android side, so that if you call it accidentally on android it doesn't crash?

aboedo:
looking great! Left a few minor comments

ios/Classes/PurchasesFlutterPlugin.m
beylmk: 😱

beylmk:
> do we still need a noop handler on the android side, so that if you call it accidentally on android it doesn't crash?

the only public functions that were add were for adding and removing listeners, and those don't interact with the methodchannels, so i think we should be fine. i'll double-check running on an android device though

ios/Classes/PurchasesFlutterPlugin.m
```diff
@@ -479,10 +493,25 @@ - (void)closeWithResult:(FlutterResult)result {
...
 
 - (void)purchases:(RCPurchases *)purchases didReceiveUpdatedPurchaserInfo:(RCPurchaserInfo *)purchaserInfo {
-    [self.channel invokeMethod:RNPurchasesPurchaserInfoUpdatedEvent
+    [self.channel invokeMethod:PurchasesPurchaserInfoUpdatedEvent
                      arguments:purchaserInfo.dictionary];
 }
 
+- (void)purchases:(RCPurchases *)purchases
+shouldPurchasePromoProduct:(SKProduct *)product
+   defermentBlock:(RCDeferredPromotionalPurchaseBlock)makeDeferredPurchase {
```
vegaro, line 502: indentation is off in these three lines

ios/Classes/PurchasesFlutterPlugin.m
```diff
@@ -479,10 +493,25 @@ - (void)closeWithResult:(FlutterResult)result {
...
+    if (!self.startPurchaseBlocks) {
+        self.startPurchaseBlocks = [NSMutableArray array];
+    }
+
+    [self.startPurchaseBlocks addObject:makeDeferredPurchase];
+    NSInteger position = [self.startPurchaseBlocks count] - 1;
+    [self.channel invokeMethod:PurchasesReadyForPromotedProductPurchaseEvent
+                     arguments:@{@"callbackID": @(position),
+                                 @"productID": product.productIdentifier
```
vegaro, line 511: I think indentation is also off in this one? I am not sure tho

lib/purchases_flutter.dart
```diff
@@ -633,6 +686,27 @@ class Purchases {
...
+      },
+    );
+    final purchaserInfo = PurchaserInfo.fromJson(
+      Map<String, dynamic>.from(result['purchaserInfo']),
+    );
+    final productIdentifier = result['productIdentifier'];
+    return PromotedPurchaseResult(
+      productIdentifier: productIdentifier,
+      purchaserInfo: purchaserInfo,
+    );
```
vegaro, line 707: `_startPromotedProductPurchase` is private but did you try calling it from the tests? In dart, privacy is at library level (not class like in Java) so I think you might still be able to invoke it. I am not sure. 

It would actually be better if you find a way to invoke `'Purchases-ReadyForPromotedProductPurchase'` in the tests to fire the event and test that it does what it should. We don't have any example on how to do that but you have access to the channel in purchases_flutter_test.dart, so I am pretty sure it's probably something like `channel.invokeMethod('Purchases-ReadyForPromotedProductPurchase', arguments)`, same way you are doing in https://github.com/RevenueCat/purchases-flutter/pull/344/files#diff-a84fbe19c95708058ab36b565eb31eb4a3071f224431bdb07085001b0f6271bbR509. I haven't tested it though, so it's just an idea.

ios/Classes/PurchasesFlutterPlugin.m
beylmk: yeah i didn't know how this one was to be handled since the second parameter is longer. my xcode doesn't seem to have any formatting set up, is there a file we use or something?

ios/Classes/PurchasesFlutterPlugin.m
beylmk: the default from xcode seemed to want to align the @s...

lib/purchases_flutter.dart
beylmk: yeah, it wouldn't let me call it from tests. i COULD make it public and then use @visibleForTesting as described [here](https://stackoverflow.com/a/54219882), but that would only warn if someone tried to call it from outside our library.

and invoking that `'Purchases-ReadyForPromotedProductPurchase'` event also wasn't working, i thought it should too since it's the same channel. i'll do a little more research on whether we can set that up...

ios/Classes/PurchasesFlutterPlugin.m
vegaro, line 502: I think this can be:

```suggestion
- (void)         purchases:(RCPurchases *)purchases
shouldPurchasePromoProduct:(SKProduct *)product
            defermentBlock:(RCDeferredPromotionalPurchaseBlock)makeDeferredPurchase {
```

lib/purchases_flutter.dart
vegaro: Alright I see your point. I don't want to block this PR on that, since we are missing a bunch of other unit tests. I tried and I also couldn't get it to work 🤔

vegaro (approved):

lib/purchases_flutter.dart
beylmk: thanks for trying... i can make a sharpen the saw ticket for investigating further flutter unit tests

PR #328, opened 2022-02-18
Release version 4.0.0

joshdholtz:
- Update to iOS v4 and Android v5
- Added Amazon support

PR #346, opened 2022-04-20
Update to 4.0.0 (Part 1): Rename customer info to purchaser info

vegaro:
Extracted from https://github.com/RevenueCat/purchases-flutter/pull/328

Renames purchaser to customer 

Comparing against release/4.0.0 so we don't break main

### DIFF
--- a/lib/purchases_flutter.dart
+++ b/lib/purchases_flutter.dart
@@ -8,9 +8,9 @@ import 'object_wrappers.dart';
 export 'object_wrappers.dart';
 
 /// Used to handle async updates from [Purchases].
-/// Should be implemented to receive updates when the [PurchaserInfo] changes.
-typedef PurchaserInfoUpdateListener = void Function(
-  PurchaserInfo purchaserInfo,
+/// Should be implemented to receive updates when the [CustomerInfo] changes.
+typedef CustomerInfoUpdateListener = void Function(
+  CustomerInfo customerInfo,
 );
 
 /// iOS Only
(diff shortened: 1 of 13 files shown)


lib/purchases_flutter.dart
```diff
@@ -351,69 +347,69 @@ class Purchases {
...
   /// and a bool that can be used to know if a user has just been created for the first time.
   ///
   /// [newAppUserID] The appUserID that should be linked to the currently user
   static Future<LogInResult> logIn(String appUserID) async {
     final result =
         await _channel.invokeMethod('logIn', {'appUserID': appUserID});
-    final purchaserInfo = PurchaserInfo.fromJson(
-      Map<String, dynamic>.from(result['purchaserInfo']),
+    final customerInfo = CustomerInfo.fromJson(
+      Map<String, dynamic>.from(result['customerInfo']),
```
aboedo, line 369: **This** is a delicate one - purchasesHybridCommon needs to also send `customerInfo`. Currently it's sending `purchaserInfo` (at least on iOS, up to [this PR](https://github.com/RevenueCat/purchases-hybrid-common/pull/129)). The PR after that one will modify this string

aboedo: Since it's the only place where we depend on a value sent by hybrid common, perhaps we can have only this change on a separate PR?

aboedo: You should be able to test this out with MagicWeather, there's a login option

aboedo (approved):
Looks great, but maybe we should sync up about the dict used in logIn

lib/purchases_flutter.dart
vegaro: I reverted the change and added todos. I think it makes sense as a separate PR

PR #348, opened 2022-04-21
Update to 4.0.0 (Part 2): Rename product, transaction and discount

vegaro:
Renames StoreProduct, StoreTransaction and StoreProductDiscount


lib/models/store_product_wrapper.dart
```diff
@@ -1,16 +1,16 @@
 import 'package:freezed_annotation/freezed_annotation.dart';
 
-import 'discount.dart';
 import 'introductory_price.dart';
+import 'store_product_discount.dart';
 
-part 'product_wrapper.freezed.dart';
-part 'product_wrapper.g.dart';
```
aboedo, line 7: maybe we can get rid of the `wrapper` part now? we're definitely inconsistent about naming, with some stuff having a `wrapper` prefix and others not having it

aboedo: like, customerInfo and storeTransaction don't have the suffix

aboedo (approved):
looks great!! I left a comment regarding the `wrapper` prefix, but we can discuss it further and act on it in a separate PR if needed

lib/models/package_wrapper.freezed.dart
```diff
@@ -28,7 +28,7 @@ class _$PackageTearOff {
       @JsonKey(name: 'packageType', unknownEnumValue: PackageType.unknown)
           PackageType packageType,
       @JsonKey(name: 'product')
-          Product product,
+          StoreProduct product,
```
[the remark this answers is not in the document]
vegaro: this is an interesting one. I checked and in Android we still have `product` but in `iOS` we have `storeProduct`. I would be opposed to change it to reduce the amount of breaking changes but we should be consistent. So maybe we should do `storeProduct` since iOS is the more popular one

lib/models/store_product_wrapper.dart
vegaro: good call. I wonder if changing the name will break people's imports? If they are not importing object_wrappers.dart it will right?

lib/models/package_wrapper.freezed.dart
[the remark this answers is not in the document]
vegaro: ok so I changed it to storeProduct but keeping the key in the dictionary as product, since that's what iOS is doing right now. If it gets updated, I will update it

PR #349, opened 2022-04-21
Update to phc 3.0.0 (Part 3): Remove identity v2 and old attribution APIs

vegaro:
Removed `createAlias`, `identify`, `reset` and `addAttributionData` along with `PurchasesAttributionNetwork`


aboedo (commented):
my favorite kind of PR ❤️  
We should start compiling a migration guide

aboedo:
the title of the PR is copy-pasta 😂 I think it should be "Remove identity v2 and old attribution APIs"

aboedo (approved):

PR #350, opened 2022-04-21
Update to phc 3.0.0 (Part 4): Add useAmazon boolean to configure function

vegaro:
- Adds `useAmazon` boolean to configure call
- Adds automatic proguard file, although this should be brought automatically from purchases-android but it won't hurt


android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
```diff
@@ -293,10 +295,16 @@ public void onMethodCall(@NonNull MethodCall call, @NonNull Result result) {
...
-    private void setupPurchases(String apiKey, String appUserID, @Nullable Boolean observerMode, final Result result) {
+    private void setupPurchases(String apiKey, String appUserID, @Nullable Boolean observerMode,
+                                @Nullable Boolean useAmazon, final Result result) {
         if (this.applicationContext != null) {
             PlatformInfo platformInfo = new PlatformInfo(PLATFORM_NAME, PLUGIN_VERSION);
-            CommonKt.configure(this.applicationContext, apiKey, appUserID, observerMode, platformInfo);
+            Store store = Store.PLAY_STORE;
+            if (useAmazon != null && useAmazon) {
+                store = Store.AMAZON;
+            }
```
aboedo, line 305: does something like this work? 
```suggestion
            Store store = useAmazon == true? Store.AMAZON: Store.PLAY_STORE
```

ios/Classes/PurchasesFlutterPlugin.m
```diff
@@ -46,6 +46,7 @@ - (void)handleMethodCall:(FlutterMethodCall *)call
         [self setupPurchases:apiKey
                    appUserID:appUserID
                 observerMode:observerMode
+                   useAmazon:NO
```
aboedo, line 49: we don't even need to have the parameter here, though, right? since it will always be `NO`, or at least, until Amazon Store for iOS is a thing

aboedo: I "Amazon Store for iOS is a thing" and I scared myself.

revenuecat_examples/purchase_tester/lib/src/constant.dart
```diff
@@ -0,0 +1,5 @@
+//TO DO: add the Apple API key for your app from the RevenueCat dashboard: https://app.revenuecat.com
+const appleApiKey = 'appl_api_key';
+
+//TO DO: add the Google API key for your app from the RevenueCat dashboard: https://app.revenuecat.com
+const googleApiKey = 'googl_api_key';
```
aboedo, line 5: should we add an amazon key here as a placeholder?

revenuecat_examples/purchase_tester/lib/main.dart
```diff
@@ -31,7 +31,12 @@ class _MyAppState extends State<InitialScreen> {
   // Platform messages are asynchronous, so we initialize in an async method.
   Future<void> initPlatformState() async {
     await Purchases.setDebugLogsEnabled(true);
-    await Purchases.setup('api_key');
+
+    if (Platform.isAndroid) {
+      await Purchases.setup(googleApiKey);
+    } else {
+      await Purchases.setup(appleApiKey);
+    }
```
aboedo, line 39: should we add an amazon key here as a placeholder?

aboedo:
it looks like the linter isn't happy?

ios/Classes/PurchasesFlutterPlugin.m
vegaro: yeah, no we can remove this. it's not needed

android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
vegaro: In Kotlin yes, but not in Java. useAmazon is a Boolean, and in order to compare to true, it needs to be unboxed, which can cause a NullPointerException if it's null

aboedo (approved):

PR #355, opened 2022-04-29
Update to phc 3.0.0 (Part 6): Fixes for local development and purchase tester

vegaro:
- purchases-hybrid-common has two variants, but this library doesn't have any, trying to work with a local copy of purchases-hybrid-common fails to compile because Gradle doesn't know which variant to use. More info on this in https://developer.android.com/studio/build/build-variants#resolve_matching_errors
- purchases-hybrid-common uses Gradle 7.2 so that also required to be updated 
- purchase tester stopped working when upgrading the AGP version and also required some fixes

### DIFF
--- a/revenuecat_examples/purchase_tester/android/app/build.gradle
+++ b/revenuecat_examples/purchase_tester/android/app/build.gradle
@@ -38,6 +38,11 @@ android {
         versionCode flutterVersionCode.toInteger()
         versionName flutterVersionName
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
+        // purchases-hybrid-common has two variants, but this library doesn't have any
+        // Gradle doesn't know which variant of purchases-hybrid-common to use when using
+        // includeBuild and working with a local copy of purchases-hybrid-common
+        // This line also needs to be added to purchases_flutter's build.gradle
+        missingDimensionStrategy 'dependencyVersions', 'latestDependencies', 'unityIAP'
     }
 
     buildTypes {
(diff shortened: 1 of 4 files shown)


revenuecat_examples/purchase_tester/android/app/build.gradle
```diff
@@ -38,6 +38,13 @@ android {
         versionCode flutterVersionCode.toInteger()
         versionName flutterVersionName
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
+        // purchases-hybrid-common has two variants, but this library doesn't have any
+        // Gradle doesn't know which variant of purchases-hybrid-common to use when using
+        // includeBuild and working with a local copy of purchases-hybrid-common
+        // This line also needs to be added to purchases_flutter's build.gradle
+        // https://developer.android.com/studio/build/build-variants#resolve_matching_errors
+        missingDimensionStrategy 'dependencyVersions', 'latestDependencies', 'unityIAP'
```
aboedo, line 46: just to be clear: for customers, this won't be an issue, because `purchases_flutter`s gradle package provides the `missingDimensionStrategy`, right? 
Why is this needed in the sample app?

vegaro: This is only an issue when pointing to a local copy of purchases-hybrid-common, since that project has two variants. When we publish `purchases-hybrid-common` we publish the variants separately as different artifacts so this is not an issue for customers

aboedo: that makes sense, thanks!

aboedo (approved):

PR #357, opened 2022-04-29
Update to 4.0.0: iOS update. Blocked on purchases-hybrid-common 3.0.0 

vegaro:


ios/Classes/PurchasesFlutterPlugin.m
```diff
@@ -2,18 +2,22 @@
...
 @import StoreKit;
 
-#import <PurchasesHybridCommon/PurchasesHybridCommon.h>
+@import PurchasesHybridCommon;
 
+typedef void (^RCPurchaseCompletedBlock)(RCStoreTransaction * _Nullable,
+                                         RCCustomerInfo * _Nullable,
+                                         NSError * _Nullable,
+                                         BOOL userCancelled);
+typedef void (^RCStartPurchaseBlock)(RCPurchaseCompletedBlock);
```
vegaro, line 11: This is a typealias in RevenueCat so it's not accessible from objc so I just copied it. When we migrate the plugin to Swift this can be removed

PR #363, opened 2022-05-06
chore: Updated freezed dependencies to latest.

pattobrien:
purchases_flutter package was creating a dependency conflict due to the breaking changes within freezed v2.0.0. Updating dependencies is the only way forward without removing revenue cat entirely.

  -  Updated ```freezed``` and ```freezed_annotation``` to the latest versions
  - As freezed is no longer including package:collection, added an extension for firstWhereOrNull to resolve the breaking change

  - No functionality changed, no additional unit tests necessary
  - All existing unit tests passed

### DIFF
--- a/lib/models/offering_wrapper.dart
+++ b/lib/models/offering_wrapper.dart
@@ -51,3 +51,12 @@ class Offering with _$Offering {
   factory Offering.fromJson(Map<String, dynamic> json) =>
       _$OfferingFromJson(json);
 }
+
+extension PackageListX on List<Package> {
+  Package? firstWhereOrNull(bool Function(Package element) test) {
+    for (final element in this) {
+      if (test(element)) return element;
+    }
+    return null;
+  }
+}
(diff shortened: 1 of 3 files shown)


lib/models/offering_wrapper.dart
```diff
@@ -51,3 +51,12 @@ class Offering with _$Offering {
   factory Offering.fromJson(Map<String, dynamic> json) =>
       _$OfferingFromJson(json);
 }
+
```
taquitos, line 54: Let's add a comment about why this is needed, something along the lines of 
```
Extension needed because this was a deprecation from freezed 1.x that was removed in 2.x?
```

taquitos:
We'll need to do a lot of regression testing as we've had issues with `freezed` in the past. Our integration tests are pretty lacking, to, so it's a manual process for right now. We won't be able to merge this until we can do that. We're also kinda swamped with just getting back from a company offsite + Google I/O next week. So that's to say- thank you for looking into this and we're definitely working to test it.

BWhiteApps:
Any update on this?

vegaro (approved):

PR #368, opened 2022-05-18
 Update to phc 3.0.0 (Part 4 Second part): Add useAmazon boolean to configure function 

vegaro:
- Creates a `PurchasesConfiguration` and an `AmazonConfiguration`
- Deprecates `setup` in favor of `configure`
-  Extracts `Store` to its own file

On top of #350 

[[CF-654]]

[CF-654]: https://revenuecats.atlassian.net/browse/CF-654?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/purchases_flutter.dart
+++ b/lib/purchases_flutter.dart
@@ -96,7 +96,7 @@ class Purchases {
         ..appUserID = appUserId
         ..observerMode = observerMode
         ..userDefaultsSuiteName = userDefaultsSuiteName
-        ..useAmazon = useAmazon
+        ..store = useAmazon ? Store.amazon : null
     );
     return configure(configuration);
   }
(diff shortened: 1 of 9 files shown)


lib/purchases_flutter.dart
```diff
@@ -80,23 +80,43 @@ class Purchases {
...
-          'appUserId': appUserId,
-          'observerMode': observerMode,
-          'userDefaultsSuiteName': userDefaultsSuiteName,
-          'useAmazon': useAmazon
-        },
-      );
+  }) {
+    final configuration = (
+        PurchasesConfiguration(apiKey)
+        ..appUserID = appUserId
```
aboedo, line 96: what are the double dots?

aboedo: TIL about [cascade notation](https://www.dartlang.org/guides/language/language-tour#cascade-notation-). That's pretty cool

aboedo (approved):

lib/purchases_flutter.dart
vegaro: Yes, I didn't feel the need to create a Builder with this notation

PR #371, opened 2022-05-20
Update samples to 4.0.0

vegaro:
I created a FlavorConfig in both examples that can be used for determining if the app is being built for Google Play, Amazon Appstore or Apple Store. Each flavor contains an api key for each platform. 

This required creating three different entry points (`main_amazon_appstore.dart`, `main_google_play.dart` and `main_apple_store.dart`). I also created three run configurations for Android Studio.

Screen Shot 2022-05-19 at 6 56 43 PM

The app can also be ran via terminal:

```
flutter run -t lib/main_google_play.dart
flutter run -t lib/main_amazon_appstore.dart
flutter run -t lib/main_apple_store.dart
```

https://medium.com/flutter-community/flutter-ready-to-go-e59873f9d7de#90e6

### DIFF
--- a/revenuecat_examples/MagicWeather/.run/Amazon Appstore.run.xml
+++ b/revenuecat_examples/MagicWeather/.run/Amazon Appstore.run.xml
@@ -1,0 +1,6 @@
+<component name="ProjectRunConfigurationManager">
+  <configuration default="false" name="Amazon Appstore" type="FlutterRunConfigurationType" factoryName="Flutter">
+    <option name="filePath" value="$PROJECT_DIR$/lib/main_amazon_appstore.dart" />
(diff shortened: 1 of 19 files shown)


aboedo (approved):
Looks good! 
I was a little confused by the usage of `flavor` at first since this has nothing to do with gradle's flavors. 
Since this only determines which store is being used (unlike gradle's flavors, which will determine dependencies), I'd vote to name it something like `StoreConfig` or something to reflect that it's about stores. 
That way the values would look like `Store.amazon`

vegaro:
makes sense! I named it flavor because most examples would actually set Gradle flavors to change things like application id or app icon, which we are not doing

PR #374, opened 2022-05-26
fix: include amazon in store enum

fire-at-will:
This PR adds `Amazon` to the `Store` enum.

### Motivation
Resolves [CF-653](https://revenuecats.atlassian.net/browse/CF-653).

### Description of Changes
- Adds `amazon` to the `Store` enum and tests that deserialization from JSON works correctly.

### Tests
- Adds JSON deserialization tests for all possible `Store` values.


fire-at-will:
Question for the reviewer: Do we need to wait for [this PR](https://github.com/RevenueCat/purchases-hybrid-common/pull/142) to be merged before merging this PR into `main`?

aboedo:
@fire-at-will we try to keep `main` shippable at any point in time, so that if we want to release something, we can rest assured that `main` is the place to go. 
This will be even more important when adding release trains, since they'll just generate releases from main. 
So the answer is yes, we should hold on merging until the relevant changes have been shipped and we can bump our hybrid common dependency on this PR. 

That said, we can review it in the meantime! I'll take a look

test/entitlement_info_test.dart
```diff
@@ -50,6 +71,42 @@ void main() {
...
+    final entitlementInfo = EntitlementInfo.fromJson(entitlementInfoJson);
+    expect(entitlementInfo.store, Store.promotional);
+  });
+
+  test('amazon store gets parsed if present in json', () {
+    final entitlementInfoJson = generateEntitlementInfoJSON('AMAZON');
+    final entitlementInfo = EntitlementInfo.fromJson(entitlementInfoJson);
+    expect(entitlementInfo.store, Store.amazon);
+  });
+
```
aboedo, line 109: ❤️

aboedo (commented):
Looking great so far! I'll approve when the required update in hybrid common gets shipped

fire-at-will:
Thanks @aboedo! I'll update this PR to contain a bump in the `purchases-hybrid-common` version when it's ready and will let you know!

vegaro:
You can also compare it against `release/4.0.0` if you want since that branch actually needs this change

fire-at-will:
@vegaro I changed the base branch to release/4.0.0 and found that the AMAZON store case has already been added there, so this PR now just contains a bump to `purchases-hybrid-common` and a few tests. Should be ready for review!

PR #378, opened 2022-05-31
Fix integration tests and samples

vegaro:
The changes introduced by https://github.com/RevenueCat/purchases-flutter/pull/371 broke some stuff (including integration tests) because of the removal of `lib/main.dart`, which flutter looks for by default.

I decided to implement the changes in a different more simple way that relies on compile time variables instead. 

I also fixed integration tests which were broken because of an old version of xcode being used that doesn't support `async`.  Checkout the integration tests passing in [this commit ](https://github.com/RevenueCat/purchases-flutter/pull/378/commits/c64a97184b9ec359fbbc41ba48b5fa5083f8052a).


vegaro:
@aboedo bump

aboedo (approved):

PR #381, opened 2022-06-15
Fix Store transaction JSON keys in 4.0.0

vegaro:
Store transaction had wrong JSON keys

[CF-753]

[CF-753]: https://revenuecats.atlassian.net/browse/CF-753?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/models/store_transaction.freezed.dart
+++ b/lib/models/store_transaction.freezed.dart
@@ -21,11 +21,11 @@ StoreTransaction _$StoreTransactionFromJson(Map<String, dynamic> json) {
 /// @nodoc
 mixin _$StoreTransaction {
   /// RevenueCat Id associated to the transaction.
-  @JsonKey(name: 'revenueCatIdentifier')
+  @JsonKey(name: 'revenueCatId')
   String get revenueCatIdentifier => throw _privateConstructorUsedError;
 
   /// Product Id associated with the transaction.
-  @JsonKey(name: 'productIdentifier')
+  @JsonKey(name: 'productId')
   String get productIdentifier => throw _privateConstructorUsedError;
 
   /// Purchase date of the transaction in ISO 8601 format.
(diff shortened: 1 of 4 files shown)


lib/models/store_transaction.freezed.dart
```diff
@@ -193,25 +193,23 @@ class _$_StoreTransaction implements _StoreTransaction {
...
-      @JsonKey(name: 'revenueCatIdentifier')
-          final String revenueCatIdentifier,
-      @JsonKey(name: 'productIdentifier')
-          final String productIdentifier,
-      @JsonKey(name: 'purchaseDate')
-          final String purchaseDate) = _$_StoreTransaction;
+          @JsonKey(name: 'revenueCatId') final String revenueCatIdentifier,
+          @JsonKey(name: 'productId') final String productIdentifier,
+          @JsonKey(name: 'purchaseDate') final String purchaseDate) =
+      _$_StoreTransaction;
```
tonidero, line 199: Small thing but I'm curious. Are we setting a 80 char line length in flutter/dart? 

I think that's the standard, but personally I'm a fan of 120 characters. That said, not going to change it unless we want to

tonidero (commented):
Do we have tests for this? 

Other than that, looks good but will wait until someone who knows more of the project reviews it.

lib/models/store_transaction.freezed.dart
vegaro: Good question, we are following the default lint rules for flutter, I guess we could change it because it's super short

vegaro:
We don't have tests, which is unfortunate. I will try to add some, I don't think we have any test that cover deserialization, but I think it should be easy. The only thing that's annoying is that the JSON is created in purchases-hybrid-common, so we have to mock it what that library sends, which could not match reality. Meaning that we could change purchases-hybrid-common to send something different and the tests would still pass.

lib/models/store_transaction.freezed.dart
aboedo: +1 for 120, it's the char limit for GitHub not to start scrolling horizontally, and usually it's still thin enough to fit most screens

vegaro: https://github.com/dart-lang/dart_style/issues/833 that's a long issue lol

vegaro:
Added some simple tests @tonidero @aboedo

lib/models/store_transaction.freezed.dart
vegaro: I looked a bit into this and I don't think there's a way of change the config for both `flutter format` and the linter 🤔 I think there are only ways of changing it in your IDE and a way to pass it as a parameter to the linter

aboedo (approved):

PR #382, opened 2022-06-15
Fix customer info delegate in 4.0.0

vegaro:
I was implementing un inexistent delegate. Since the delegate is optional, xcode wasn't complaining

[CF-752]

[CF-752]: https://revenuecats.atlassian.net/browse/CF-752?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


aboedo:
FWIW we do have the method marked as obsolete, but I guess since it's an optional method in a delegate, Xcode has no way of determining that the method you implemented is definitely there for the delegate. That sucks

PR #388, opened 2022-06-24
Add API tests to purchases-flutter

tonidero:
https://revenuecats.atlassian.net/browse/CSDK-139

This adds a new project in the root folder called `api_tester`. This will call all public APIs of the flutter SDK to make sure there are no unintentional regressions on the API. This follows similar patterns we already have on the iOS and android sdks, where we detect issues at compile time.

* Note that I've based this PR on `release/4.0.0` since it seems it had a bunch of changes on the API.


api_tester/lib/models/customer_info_wrapper_api_test.dart
```diff
@@ -0,0 +1,75 @@
+import 'package:purchases_flutter/object_wrappers.dart';
+
+// ignore_for_file: unused_element
+// ignore_for_file: unused_local_variable
+class _CustomerInfoApiTest {
+  void _checkFromJsonFactory(Map<String, dynamic> json) {
```
tonidero, line 6: I added `fromJson` and `toJson` checks to all these `freeze` models... I don't think that's meant to be public, but it currently is (don't think there is an easy fix for that though), so I thought I would add them as part of the API tests... Lmk if you prefer me to remove them.

api_tester/lib/models/customer_info_wrapper_api_test.dart
```diff
@@ -0,0 +1,75 @@
...
+    CustomerInfo customerInfo = CustomerInfo.fromJson(json);
+  }
+
+  void _checkToJson(CustomerInfo customerInfo) {
+    Map<String, dynamic> json = customerInfo.toJson();
+  }
+
+  void _checkConstructor(
```
tonidero, line 14: Same as with the `fromJson` and `toJson` methods, don't think we want the constructor to be public, but it currently is so I added api tests for them.

api_tester/lib/models/customer_info_wrapper_api_test.dart
```diff
@@ -0,0 +1,75 @@
...
+        allPurchaseDates,
+        activeSubscriptions,
+        allPurchasedProductIdentifiers,
+        nonSubscriptionTransactions,
+        firstSeen,
+        originalAppUserId,
+        allExpirationDates,
+        requestDate
+    );
+    customerInfo = CustomerInfo(
```
tonidero, line 40: Note how I'm checking the constructor twice, once with the required parameters, and another one with the optional parameters. I thought it was a bit weird that the optional parameters are always named and the required parameters are positional... But I think it's not really a problem

api_tester/lib/errors_api_test.dart
```diff
@@ -0,0 +1,41 @@
...
+      case PurchasesErrorCode.operationAlreadyInProgressError:
+      case PurchasesErrorCode.unknownBackendError:
+      case PurchasesErrorCode.invalidAppleSubscriptionKeyError:
+      case PurchasesErrorCode.ineligibleError:
+      case PurchasesErrorCode.insufficientPermissionsError:
+      case PurchasesErrorCode.paymentPendingError:
+      case PurchasesErrorCode.invalidSubscriberAttributesError:
+      case PurchasesErrorCode.logOutWithAnonymousUserError:
+      case PurchasesErrorCode.configurationError:
+      case PurchasesErrorCode.unsupportedError:
```
tonidero, line 33: If there are missing `case` on these enum tests, it will be caught during the `flutter analyze` step

api_tester/lib/models/purchase_configuration_api_test.dart
```diff
@@ -0,0 +1,41 @@
...
+    bool observerMode = configuration.observerMode;
+    String? userDefaultsSuiteName = configuration.userDefaultsSuiteName;
+    Store? store = configuration.store;
+    configuration.appUserID = null;
+    configuration.appUserID = "fakeUserId";
+    configuration.observerMode = false;
+    configuration.userDefaultsSuiteName = null;
+    configuration.userDefaultsSuiteName = "fakeSuiteName";
+    configuration.store = null;
+    configuration.store = Store.playStore;
```
tonidero, line 21: `PurchaseConfiguration` actually allows to set these properties, which I found a bit weird... We might want to mimic something like Java's `Builder` pattern in the future?

api_tester/lib/purchases_flutter_api_test.dart
```diff
@@ -0,0 +1,420 @@
...
+
+  void _checkRemoveReadyForPromotedProductPurchaseListener() {
+    void listener(
+        String productIdentifier,
+        Future<PromotedPurchaseResult> Function() startPurchase
+        ) {}
+    Purchases.removeReadyForPromotedProductPurchaseListener(listener);
+  }
+
+  void _checkGetOfferings() async {
```
tonidero, line 64: Note how I'm marking some of these methods as `async`. This is so we can test `async` methods easily. Right now we have a mix, with some allowing `async` and some not.

On these tests, I'm testing both the `Future` value returned, and the `await` system, to make sure there are no regressions on that part. I thought we could only test the `await` system... But I wasn't sure, so I added both. Lmk if you think only with the `await` system would be enough.

api_tester/lib/purchases_flutter_api_test.dart
```diff
@@ -0,0 +1,420 @@
...
+  void _checkPurchaseProduct() async {
+    String productIdentifier = "fakeProductId";
+    UpgradeInfo? upgradeInfo;
+    PurchaseType purchaseType = PurchaseType.subs;
+    Future<CustomerInfo> customerInfoFuture = Purchases.purchaseProduct(
+      productIdentifier,
+      upgradeInfo: upgradeInfo,
+      type: purchaseType
+    );
+    customerInfoFuture = Purchases.purchaseProduct(
```
tonidero, line 93: When there are optional parameters, I'm calling the method multiple times to make sure the different options are valid

fastlane/Fastfile
```diff
@@ -93,4 +100,31 @@ def replace_in(previous_text, new_text, path)
   sed_regex = 's|' + previous_text + '|' + new_text + '|'
   backup_extension = '.bck'
   sh("sed", '-i', backup_extension, sed_regex, path)
-end
\ No newline at end of file
+end
+
+def check_api_tester_imports_up_to_date
```
tonidero, line 105: This was a slightly hacky approach I thought to make sure we don't forget adding api_tests to the `api_tests_import.dart` file. Any thoughts are welcome!

aboedo: it looks great to me. Small nitpick is that I'd probably have the actual api tests live in their own folder within `lib` if possible, so that we don't have to filter out the `main.dart` and `api_tests_import.dart` files and also so that everything in a folder looks very similar

aboedo (approved):
Looks amazing!!

api_tester/android/app/build.gradle
```diff
@@ -0,0 +1,71 @@
+def localProperties = new Properties()
```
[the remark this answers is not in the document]
tonidero: Well to test this, we actually have to build the project. Right now I'm building an `apk` for android, which will compile all these api tests, so at least 1 platform is needed to actually build. I first tried to remove some things but run into some issues, so I decided to leave those folders as they were.

api_tester/lib/models/purchase_configuration_api_test.dart
vegaro: I started doing a builder for the `PurchaseConfiguration` and then removed it since I found it clean to just use the .. notation

```
configuration = PurchasesConfiguration(StoreConfig.instance.apiKey)
         ..appUserID = null
         ..observerMode = false;
```

I hadn't thought about this case though, so I think it makes sense to go back to that builder

tonidero: Sounds good! Let me know when that's done and I can rebase my branch over your changes

PR #421, opened 2022-08-09
Add release automation

tonidero:
This PR will automate part of the release process for new versions of flutter. Now, after pushing a new tag, we will create a github release, publish to `pub.dev`, and create a new PR pointing to the next snapshot version automatically.


fastlane/Fastfile
```diff
@@ -86,3 +156,19 @@ def analyze_api_tester
...
+  android_build_gradle_contents = File.read(android_build_gradle_path)
+  android_phc_version = android_build_gradle_contents.match("ext.common_version = '(.*)'").captures[0]
+  UI.user_error!("Android PHC version not found.") if android_phc_version.nil?
+
+  ios_podspec_path = '../ios/purchases_flutter.podspec'
+  ios_podspec_contents = File.read(ios_podspec_path)
+  ios_phc_version = ios_podspec_contents.match("s.dependency 'PurchasesHybridCommon', '(.*)'").captures[0]
+  UI.user_error!("iOS PHC version not found.") if ios_phc_version.nil?
+
+  UI.user_error!("Android and iOS PHC versions don't match. Please make sure they match.") if android_phc_version != ios_phc_version
```
tonidero, line 171: This is a new validation that will happen during the `bump` lane. I think it's useful. That said, I'm not sure if we ever want to have different versions of PHC in iOS and Android. If so, we probably need to figure out what we want to put in the `VERSIONS.md` file.

.gitignore
```diff
@@ -37,3 +37,6 @@ fastlane/report.xml
 *.bck
 
 fastlane/.env
+
+# CI created folders
+vendor/
```
tonidero, line 42: When publishing a package, it seems everything in the repo that is not gitignored will be uploaded. This folder stores gem dependencies on CI jobs, so I had to gitignore it to make sure it only included the packages we wanted.

tonidero (commented):
The next step will be to actually create a release and tag and test that everything works.

.circleci/config.yml
```diff
@@ -1,12 +1,21 @@
-release-tags-and-branches: &release-tags-and-branches
-  filters:
-    tags:
-      ignore: /^.*-SNAPSHOT/
-    branches:
-      only: /^release\/.*/
+
 
 version: 2.1
```
vegaro, line 3: NADB, maybe clean up the white space above this line

.circleci/config.yml
```diff
@@ -227,6 +246,30 @@ jobs:
       - swissknife/fail_if_dirty:
           print-modified-files: true
 
+  prepare-next-version:
+    description: "Creates a PR with the new snapshot version"
+    docker:
+      - image: cirrusci/flutter
```
vegaro, line 252: Does this need to be `cirrusci/flutter` or can it be just a ruby image? I wonder which one is smaller, but I assume the plain ruby one

RELEASING.md
```diff
@@ -8,9 +8,5 @@
 1. Update to the latest SDK versions in ios/purchases_flutter.podspec, macos/purchases_flutter.podspec and android/build.gradle.
 1. Update versions in VERSIONS.md and push changes
 1. If purchases-hybrid-common was updated, run `pod update PurchasesHybridCommon` in both `MagicWeather` and `purchase_tester`
-1. `flutter pub pub publish --dry-run`
 1. Wait until PR approved and make sure local copy matches remote
-1. `git tag x.y.z`
-1. `git push origin --tags`
-1. Create a new release in github
-1. `flutter pub pub publish`
+1. Create a tag for the new release in the last commit of the branch and push the tag. The rest will be performed automatically by CircleCI. If the automation fails, you can revert to manually calling `bundle exec fastlane release`.
```
vegaro, line 12: I love this

RELEASING.md
```diff
@@ -8,9 +8,5 @@
 1. Update to the latest SDK versions in ios/purchases_flutter.podspec, macos/purchases_flutter.podspec and android/build.gradle.
 1. Update versions in VERSIONS.md and push changes
 1. If purchases-hybrid-common was updated, run `pod update PurchasesHybridCommon` in both `MagicWeather` and `purchase_tester`
-1. `flutter pub pub publish --dry-run`
```
vegaro, line 11: I think this is still good to run because it will sometimes point issues with formatting that will most likely fail the publish command

tonidero: It will be run at the beginning of the `release` lane, so errors would be catched before creating a github release or any other operations. Do you think we should still run it locally?

.circleci/config.yml
tonidero: Probably should work with a plain ruby image. Will try it out, we might have the same issue with the cache we had in iOS (or maybe not since it's also a linux image).

.circleci/config.yml
```diff
@@ -243,3 +286,8 @@ workflows:
   ios-integration-test:
     jobs:
       - ios-integration-test: *release-tags-and-branches
+
+  deploy:
+    jobs:
+      - make-release: *release-tags
+      - prepare-next-version: *release-tags
```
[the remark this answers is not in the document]
tonidero: Hmm this is setup as a single workflow in iOS and android. Should we update those as well? Also, I believe you can run failed jobs in a workflow: 
image
Does that have any issues that you know of?

fastlane/Fastfile
```diff
@@ -45,6 +49,45 @@ lane :bump do |options|
...
+    hybrid_common_version: phc_version
+  )
+  commit_current_changes
+  push_to_git_remote(set_upstream: true)
+end
+
+desc "Create release"
+lane :release do |options|
+  Dir.chdir('..') do
+    sh('flutter', 'pub', 'pub', 'publish', '--dry-run')
```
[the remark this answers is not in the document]
tonidero: Hmm well it seems to work fine with a single `pub`. I can try to remove it and see if it causes any issues on the next release.

About the dry run, I think I would still like to run it here to make sure we stop the release before doing anything, but I can add it to the releasing instructions to run it locally so we can figure out issues early

fastlane/Fastfile
```diff
@@ -45,6 +49,45 @@ lane :bump do |options|
...
+    github_release(version: current_version_number)
+    setup_credentials_file
+    sh('flutter', 'pub', 'pub', 'publish', '--force')
```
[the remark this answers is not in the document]
tonidero: So it doesn't ask for confirmation of the release (note that I haven't tested it with the flag yet. Plan to do that once we actually do a release)

RELEASING.md
taquitos: `show` ->`show.`

taquitos (approved):
This is really exciting! 🚢🐐

.circleci/config.yml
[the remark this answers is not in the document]
vegaro: Leave it like this. I still don't understand what "from Failed" means, since I believe sometimes I see it runs even succeeded jobs but I just tested it and it works as expected. Don't worry about it.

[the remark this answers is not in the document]
tonidero: My guess is that it would start executing failed jobs, and any jobs that depend on those... That said, I'm not sure either 🤷‍♂️

RELEASING.md
vegaro: I think it can save some time to run it locally, otherwise the job has to be rerun right? But I am not opposed. I am just saying because in a lot of the releases I've done, there was some formatting error that would prevent the release

vegaro (approved):

RELEASING.md
tonidero: Yeah I think that makes sense. I added it back to the `RELEASING.md` doc but I also run it at the beginning of the `release` lane just in case

PR #423, opened 2022-08-09
Update Hybrid Common 4.1.0 and add AdServices

joshdholtz:
- Updated Hybrid Common to `4.1.0`
- Adds new `enableAdServicesAttributionTokenCollection` method
  - Will essentially be a no-op on non-iOS platforms


api_tester/lib/api_tests/purchases_flutter_api_test.dart
```diff
@@ -144,6 +144,11 @@ class _PurchasesFlutterApiTest {
     future = Purchases.setAutomaticAppleSearchAdsAttributionCollection(enabled);
   }
 
+  void _checkEnableAdServicesAttributionTokenCollection() {
+    Future<void> future;
+    future = Purchases.enableAdServicesAttributionTokenCollection();
```
taquitos, line 149: Future<void> future = Purchases.enableAdServicesAttributionTokenCollection();

taquitos (approved):
Tiny nit, but otherwise 🚢🐐

api_tester/lib/api_tests/purchases_flutter_api_test.dart
tonidero: I think we want to specify the type of the future since this is part of the API tests. We can maybe make it in a single line... Not sure why it was separated in other methods

tonidero (approved):

PR #425, opened 2022-08-09
Update minimum XCode version & iOS target in README

britannio:
As per https://docs.revenuecat.com/docs/ios#requirements

I previously had `platform :ios, '10.0'` in `ios/Podfile` but pod couldn't satisfy dependencies.


aboedo (approved):
Thanks so much for calling this out and for making a PR to fix it!! We missed it in the readme notes

PR #436, opened 2022-08-12
Bump compileSdkVersion for flutter in android

tonidero:
This PR bumps the `compileSdkVersion` for the android library to 31, to match what we have in `purchases-android`.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -26,7 +26,7 @@ apply plugin: 'com.android.library'
 apply plugin: 'kotlin-android'
 
 android {
-    compileSdkVersion 30
+    compileSdkVersion 31
 
     compileOptions {
         sourceCompatibility JavaVersion.VERSION_1_8


android/build.gradle
```diff
@@ -26,7 +26,7 @@ apply plugin: 'com.android.library'
 apply plugin: 'kotlin-android'
 
 android {
-    compileSdkVersion 30
+    compileSdkVersion 31
```
tonidero, line 29: I was thinking about ways to make this the same across our android repos... It's a bit painful to have to update it in so many places currently, but it's difficult to share this information across repos. One idea I had was to validate that the project matches the version of the subproject as part of the tests, so hybrids would check PHC and PHC would check `purchases-android`... What do you think?

vegaro: How would you do that? We are thinking about moving to a monorepo so things like this can be easily shared across

vegaro (approved):

android/build.gradle
tonidero: Monorepo would definitely help. My idea was to do a basic search for the `compileSdkVersion` on the dependency (just searching for the number after `compileSdkVersion` or wherever it's stored), and see that it matches the version in the library itself. Very basic. I think this test would have to be done through fastlane. Not sure if I'm overengineering this too much though... Might not be worth it.

PR #440, opened 2022-08-16
Update fastlane plugin

tonidero:
### Description
This PR bumps the fastlane internal plugin to use the latest version. One of the changes is that we won't be validating the base branch automatically in the `bump` lane anymore. Instead, we will ask for confirmation from the user in a prompt: https://github.com/RevenueCat/fastlane-plugin-revenuecat_internal/pull/20

Issue #403, opened 2022-07-27
[4.0.0-rc.1] User cancelled crash


MinseokKang003:
After updating from 4.0.0-beta.1 to 4.0.0-rc.1, 
app crashes(unhandled) when user cancels payment process.

The issue is resolved when downgrading back to 4.0.0-beta.1.

Can you please check if the issue is reproducible?

samry:
Experiencing the same issue, except I'm on 4.0.0-beta.1.

The device (NOT a simulator) is an iPhone XR with iOS 15.5.

Logs below/:

`-[__SwiftValue _fastCStringContents:]: unrecognized selector sent to instance 0x283330cf0
*** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '-[__SwiftValue _fastCStringContents:]: unrecognized selector sent to instance 0x283330cf0'
*** First throw call stack:
(0x193901288 0x1ac5fb744 0x1939defc4 0x193895e98 0x193894f70 0x19852fd58 0x19852ff0c 0x1985300f0 0x197fc1d1c 0x10619096c 0x106190770 0x106184cd8 0x106184240 0x10685022c 0x1068b7454 0x1068b74b0 0x193566e6c 0x193568a30 0x193576f48 0x193576b98 0x1938b9800 0x193873704 0x193886bc8 0x1af9ba374 0x1961f6648 0x195f77d90 0x104d7c3a0 0x104fbdce4)`

MinseokKang003:
@samry 's logs are the same crash logs I encountered.

I'm on iPhone13 Pro(Real Device) with iOS 15.5.

beylmk:
Hey all,

Thanks for reporting... I'm looking into this today

beylmk:
We've identified the issue and are working on a fix...I'll post back here when it's out. Thank you for your patience!

beylmk:
Hey all, please see [4.0.0-rc.2](https://github.com/RevenueCat/purchases-flutter/releases/tag/4.0.0-rc.2) for the fix. Feel free to re-open this issue if needed.

NachoSoto:
Proper fix for this is https://github.com/RevenueCat/purchases-ios/pull/1841
We will update [PurchasesHybridCommon](https://github.com/RevenueCat/purchases-hybrid-common) and then we can re-enable SK2 here.

PR #448, opened 2022-08-23
Release/4.2.0

NachoSoto:
### Fixes
- Fixed crash when cancelling purchases (#403) via NachoSoto (@NachoSoto)

### Other Changes
* Updated `purchases-ios` to 4.10.2
* Re-enabled `SK2` as the default on iOS
* `PurchasesFlutterPlugin`: silenced unused warning (#444) via NachoSoto (@NachoSoto)
* `README`: removed `Amazon` note since it's now supported. (#442) via NachoSoto (@NachoSoto)
* Update fastlane plugin (#440) via Toni Rico (@tonidero)


taquitos (approved):
🚢 it

aboedo:
I think it'd be safer to only make SK2 the default if we also introduce the option to roll back to sk1 in the configure method. 

That way, if anything goes wrong or unexpected (like yesterday's report of differences with intro eligibility), customers have an easy way to roll back. 

Could we add the option to `configure` first?

NachoSoto:
That's a great point, I'll do that.

NachoSoto:
Done: #450

aboedo:
The added option will now make this a 4.2.0 release, right?

NachoSoto:
Good point 👍🏻 not sure ill have time to do this today so if somebody wants to wake over feel free!

PR #450, opened 2022-08-23
Added option to disable SK2

NachoSoto:
Also enabled by default.

### DIFF
--- a/lib/models/purchases_configuration.dart
+++ b/lib/models/purchases_configuration.dart
@@ -23,6 +23,10 @@ class PurchasesConfiguration {
   /// use standardUserDefaults.
   String? userDefaultsSuiteName;
 
+  /// iOS-only, will be ignored for Android.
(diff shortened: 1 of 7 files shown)


aboedo (approved):

lib/models/purchases_configuration.dart
```diff
@@ -23,6 +23,10 @@ class PurchasesConfiguration {
   /// use standardUserDefaults.
   String? userDefaultsSuiteName;
 
+  /// iOS-only, will be ignored for Android.
+  /// Set this to FALSE to disable StoreKit2.
```
tonidero, line 27: Maybe add `/// Default is true.` as the other comments? It's kind of implicit by this line, but I think it's good to be explicit.

tonidero (approved):

lib/models/purchases_configuration.dart
NachoSoto: Done.

PR #464, opened 2022-08-31
Remove wrong link from Changelog

vegaro:
I pointed to the wrong issue in the latest changelog


tonidero (approved):
Should we point to the correct PR? Though that's probably an iOS repo link.

vegaro:
Let's point to the React Native issue

tonidero (approved):

PR #461, opened 2022-08-30
add SDK reference docs link

aboedo:

PR #467, opened 2022-09-07
[AUTOMATIC] Release/4.2.2

RCGitBot:
**This is an automatic release.**

### Other Changes
* add SDK reference docs link (#461) via Andy Boedo (@aboedo)


vegaro:
Interesting, since both the plugin version and phc were the same `4.2.1`, both were updated to `4.2.2`. And there's no `4.2.2` for phc, so that's why the tests failed https://circleci.com/gh/RevenueCat/purchases-flutter/3609?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

I am not sure what we can do to make sure the `BumpVersionUpdateChangelogCreatePrAction` only updates the correct version and not any dependency. It only looks for a version to update, so if any dependency has the same version as the SDK being updated, it will also be updated

cc: @tonidero

tonidero:
Ohh yeah... I have 2 options come to mind:
- Parse the specific version within each file instead of a simple find-replace, but it would be difficult to make that generic... I guess we could try to pass a series of regexes besides the file path... But seems like that would complicate things...
- Make sure we don't have PHC and SDK version in the same file ever. Somewhat simpler but might require some extra wiring 

I would vote for the second option. Thoughts?

vegaro:
@tonidero Yes, first option is what I had in mind, but I can't think of a way of making it generic. How would you do second option?


btw, I am going to close this PR since there will be another release tomorrow

tonidero:
@vegaro The second options I mentioned would depend on each case as well. Basically the idea would be to move the version number to very small files that only contain that information. In this case we just found, we could create a new gradle file that only contains the SDK version and import it on that `android/build.gradle` file so we can separate it from the rest of the file and make it pretty much impossible to find a collision like this.

PR #468, opened 2022-09-08
Adds lane to update purchases-hybrid-common

vegaro:
Creates a `update_hybrid_common` lane


fastlane/Fastfile
```diff
@@ -128,6 +128,110 @@ lane :tag_current_branch do |options|
...
+    sh("pod install")
+  end
+  Dir.chdir(File.expand_path("revenuecat_examples/purchase_tester/macos", get_root_folder)) do
+    sh("pod update PurchasesHybridCommon --repo-update")
+    sh("pod install")
+  end
+end
+
+desc "Update iOS pod version during hybrid update"
+lane :update_ios_pod do |options|
```
vegaro, line 184: I copied this from React Native... I don't think this is actually doing anything right?

fastlane/Fastfile
```diff
@@ -128,6 +128,110 @@ lane :tag_current_branch do |options|
...
+    replace_text_in_files(
+      previous_text: previous_version_number,
+      new_text: new_version_number,
+      paths_of_files_to_update: files_to_update
+    )
+  end
+end
+
+desc "Build hybrid example"
+lane :build_hybrid_example do |options|
```
tonidero, line 167: I'm guessing all these will update the `Podfile.lock` files right? I wonder if we should just substitute the version as we do in the `podspec` files... That would not work locally though, since we still would need to run this command, so I guess I'm ok with keeping this one.

tonidero (commented):
I think we might want to refactor that code so we don't have that extra `update_ios_pod` lane... Other than that looks good!

fastlane/Fastfile
vegaro: I think I prefer running the `pod` command in case any transitive dependency gets updated right?

fastlane/Fastfile
```diff
@@ -128,6 +128,110 @@ lane :tag_current_branch do |options|
...
+  build_hybrid_example(options)
+end
+
+###############################################################################
+# Helper functions 🤜🤛                                                      #
+###############################################################################
+
+def parse_pod_version
+  Dir.chdir(get_root_folder) do
+    return sh("cat revenuecat_examples/MagicWeather/ios/Podfile.lock | grep \' RevenueCat (=\' | awk \'{print($4)}\' | sed \"s/)//g\"").strip
```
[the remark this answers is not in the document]
vegaro: actually, I don't think we need this function at all. I was checking its usage and it's useless haha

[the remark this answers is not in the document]
vegaro: actually the spec checksums get updated too when running `pod install` so we need to run that command

fastlane/Fastfile
tonidero: Ah yeah you're totally right 👍

tonidero (approved):
I guess what we lose by removing that previous lane are the logs, but I think this is fine, no need to be so verbose

vegaro:
The other lane is actually logging the same so we should be good

PR #473, opened 2022-09-26
Creates update-hybrid-common-versions and updated manual bump

vegaro:
Depends on https://github.com/RevenueCat/fastlane-plugin-revenuecat_internal/pull/34

The idea is to run this workflow from purchases-hybrid-common's CI, so everytime there's a new release we trigger an update in this repository.

See an example of how it's triggered in https://github.com/RevenueCat/purchases-hybrid-common/pull/238/files#diff-dff4b99d4e651ab9d7597876ec8dbe92aa934e42c0fb55f4aadd6c106fc96688R258


fastlane/Fastfile
```diff
@@ -128,6 +128,12 @@ lane :tag_current_branch do |options|
   push_git_tags(tag: version_number)
 end
 
+desc "Update hybrid common pod and gradle"
+lane :update_hybrid_common_versions_and_open_pr do |options|
+  options[:open_pr] = true
```
vegaro, line 133: I wanted to reuse the code for local and automatic updates. Local updates are most likely run while preparing a local release, so we don't really want to open a PR in those cases

.circleci/config.yml
```diff
@@ -242,6 +251,20 @@ jobs:
...
+      - image: cimg/ruby:3.1.2
+    steps:
+      - checkout
+      - revenuecat/install-gem-unix-dependencies:
+          cache-version: v1
+      - revenuecat/trust-github-key
+      - revenuecat/setup-git-credentials
+      - run:
+          name: Prepare next version
+          command: bundle exec fastlane update_hybrid_common_versions version:<< pipeline.parameters.hybrid-common-version >> open_pr:true
```
tonidero, line 266: Should we add the `automatic_release` parameter?

fastlane/Fastfile
```diff
@@ -139,12 +145,9 @@ lane :update_hybrid_common_versions do |options|
     UI.user_error!("Missing `version` argument")
   end
 
-  previous_version_number = get_phc_version
-
-  # Add to options so if we're chaining lanes together, the other lanes get it
-  options[:previous_hybrid_version] = previous_version_number
```
tonidero, line 145: Just to check, the `options[:previous_hybrid_version]` is not used?

fastlane/Fastfile
```diff
@@ -154,13 +157,16 @@ lane :update_hybrid_common_versions do |options|
...
-    )
+    next
   end
+
+  bump_phc_version_create_pr(
+    repo_name: "purchases-flutter",
+    files_to_update: files_to_update,
+    current_version: current_phc_version,
+    next_version: new_version_number,
+    open_pr: options[:open_pr]
```
tonidero, line 168: Should we add a default value for this?

tonidero (approved):
❤️ this

fastlane/Fastfile
```diff
@@ -154,13 +157,16 @@ lane :update_hybrid_common_versions do |options|
...
+  )
```
tonidero, line 169: Should we add the `automatic_release` parameter?

fastlane/Fastfile
vegaro: No. It also doesn't really work like this. If you add something to `options` in a lane, it's not accessible for the following lanes that are being executed. This basically didn't work anyway

fastlane/Fastfile
vegaro: Oh good idea

vegaro: I added here, but not when calling it from `update_hybrid_common_versions_and_open_pr` because that one may not always be used for automatic releases. I added it to the CircleCI workflow tho

PR #483, opened 2022-10-05
Fix dirty repository when upgrading phc version

vegaro:
Fixes https://app.circleci.com/pipelines/github/RevenueCat/purchases-flutter/1095/workflows/0b9958b7-328e-4c03-9286-0c7c99b05432/jobs/4075

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -100,10 +100,10 @@ commands:
     steps:
       - run:
           name: Download flutter SDK
-          command: git clone -b stable https://github.com/flutter/flutter.git ../flutter
+          command: git clone -b stable https://github.com/flutter/flutter.git flutter
       - run:
           name: Setup environment variables
-          command: echo 'export PATH="$PATH:`pwd`/../flutter/bin"'  >> $BASH_ENV
+          command: echo 'export PATH="$PATH:`pwd`/flutter/bin"'  >> $BASH_ENV
       - run:
           name: Flutter doctor
           command: flutter doctor
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -100,10 +100,10 @@ commands:
     steps:
       - run:
           name: Download flutter SDK
-          command: git clone -b stable https://github.com/flutter/flutter.git flutter
+          command: git clone -b stable https://github.com/flutter/flutter.git ../flutter
       - run:
           name: Setup environment variables
-          command: echo 'export PATH="$PATH:`pwd`/flutter/bin"'  >> $BASH_ENV
+          command: echo 'export PATH="$PATH:`pwd`/../flutter/bin"'  >> $BASH_ENV
       - run:
           name: Flutter doctor
           command: flutter doctor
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -100,10 +100,10 @@ commands:
     steps:
       - run:
           name: Download flutter SDK
-          command: git clone -b stable https://github.com/flutter/flutter.git flutter
+          command: git clone -b stable https://github.com/flutter/flutter.git ../flutter
```
vegaro, line 103: This was causing the repo to be dirty

Gemfile.lock
```diff
@@ -1,6 +1,6 @@
 GIT
   remote: https://github.com/RevenueCat/fastlane-plugin-revenuecat_internal
-  revision: 853cb3dd0a153dfdd747d661387602abef0f7623
+  revision: 7080bf23518eea5b53c88d9ccc5cde6ef7bfba07
```
vegaro, line 3: this will fix the reviewers

fastlane/Fastfile
```diff
@@ -189,6 +189,8 @@ desc "Update Hybrid Common, Build PurchaseTester, and updates download scripts.
 lane :update_hybrid_common do |options|
   update_hybrid_common_versions(options)
   build_hybrid_example(options)
+  commit_current_changes(commit_message: 'Update examples')
+  push_to_git_remote(set_upstream: true)
```
vegaro, line 193: This pushes the changes made by `build_hybrid_example`. We were not pushing those changes to the branch

.circleci/config.yml
tonidero: We could also add the flutter directory to gitignore. I kinda prefer that approach to keep everything in the same workplace.

fastlane/Fastfile
tonidero: Hmm this will push to the current branch, and if we call this from somewhere different than expected we might be coming to the wrong branch... Maybe we can check that we are not in main at least?

tonidero (commented):
A couple of small things but looking good

.circleci/config.yml
vegaro: ohhhh good point! haha

fastlane/Fastfile
vegaro: I see, we do make a change to a bump branch in `update_hybrid_common_versions` but it's true that if we change anything in that lane, this could have unexpected consequences. 

I think it makes sense to check if we are in `main` yes.

I wish there was an easy way to do this at the same time we commit the version updates in  `update_hybrid_common_versions`, but that happens in `bump_phc_version` and I think it would be a big refactor

fastlane/Fastfile
```diff
@@ -189,6 +189,13 @@ desc "Update Hybrid Common, Build PurchaseTester, and updates download scripts.
 lane :update_hybrid_common do |options|
   update_hybrid_common_versions(options)
   build_hybrid_example(options)
+  current_branch = sh("git branch --show-current").strip
+  unless current_branch == 'main'
+    commit_current_changes(commit_message: 'Update examples')
+    push_to_git_remote(set_upstream: true)
+  else
+    UI.message("Current branch is main. Skipping pushing to branch.")
+  end
```
tonidero, line 198: We might want to update the description of the lane to include that we are also committing and pushing

tonidero (approved):

PR #495, opened 2022-10-06
Update update-hybrid-common-versions parameters

vegaro:
- Changed `hybrid-common-version` to `version`
- Created `automatic` to make it clear which ones are triggered automatically

Depends on https://github.com/RevenueCat/purchases-hybrid-common/pull/248/files

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,7 +30,7 @@ orbs:
   android: circleci/android@0.2.1
   gcp-cli: circleci/gcp-cli@2.1.0
   swissknife: roopakv/swissknife@0.65.0
-  revenuecat: revenuecat/sdks-common-config@2.0.0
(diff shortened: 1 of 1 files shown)


tonidero (approved):

.circleci/config.yml
```diff
@@ -263,9 +266,12 @@ jobs:
       - revenuecat/setup-git-credentials
       - setup-flutter
       - run:
-          name: Prepare next version
-          command: bundle exec fastlane update_hybrid_common version:<< pipeline.parameters.hybrid-common-version >> open_pr:true automatic_release:true
-
+          name: Updating to next version of phc
+          command: |
+            bundle exec fastlane update_hybrid_common 
+            version:<< pipeline.parameters.version >> \
```
tonidero, line 272: Hmm I see you are adding the `\` from the second line but not on the first one. Does it need it there as well?

PR #522, opened 2022-11-05
Add setCleverTapId, setMixpanelDistinctID, setFirebaseAppInstanceId

sdurban:
Hi! In my team we wanted to send the FirebaseAppInstanceID to Revenuecat from the Flutter SDK, but there is no method to set it!

I know we can do it from the setAttributes method (`Purchases.setAttributes({r'$firebaseAppInstanceId': firebaseId});`) but it seems a bit ugly to do it this way having the others with their own setters.

So I started preparing a PR to add the FirebaseAppInstanceID but in the end I added other methods that I saw that are also missing in the Flutter SDK.

I just tested with iOS and Android with our app, and it seems to work properly. I didn't find any tests on setters, if there are any please let me know.

(Kinda bad english sorry ;). )

Changes:

* added `setCleverTapId` method
* added `setMixpanelDistinctID` method
* added `setFirebaseAppInstanceId` method


vegaro (approved):
Thanks so much for opening these. It looks good to me

PR #546, opened 2022-12-12
Prepare next version: 4.6.0-SNAPSHOT

RCGitBot:


ios/purchases_flutter.podspec
```diff
@@ -15,7 +15,7 @@ Pod::Spec.new do |s|
   s.source_files = 'Classes/**/*'
   s.public_header_files = 'Classes/**/*.h'
   s.dependency 'Flutter'
-  s.dependency 'PurchasesHybridCommon', '4.6.0-SNAPSHOT'
+  s.dependency 'PurchasesHybridCommon', '4.5.1'
```
tonidero, line 18: I think this has become a bit of a problem... One solution I can think of is to store the PHC version in a different file and get it here by reading that file. Something like: `phc_version = File.read(File.join(__dir__, '..', '.phc_version'))` and then set the version here to `phc_version`... A similar approach could work for the gradle file.

PR #569, opened 2023-01-05
Adds missing params for better changelogs when bumping

vegaro:
Adds params added in https://github.com/RevenueCat/fastlane-plugin-revenuecat_internal/pull/42


fastlane/Fastfile
```diff
@@ -48,7 +48,9 @@ lane :bump do |options|
     github_rate_limit: options[:github_rate_limit],
     editor: options[:editor],
     next_version: options[:next_version],
-    automatic_release: options[:automatic_release]
+    automatic_release: options[:automatic_release],
+    hybrid_common_version: phc_version,
+    versions_file_path: './VERSIONS.md'
```
tonidero, line 53: Hmm might make sense to extract to a constant since I see it's repeated.

tonidero (approved):

PR #576, opened 2023-01-13
Send customer info to first listener added

tonidero:
Fixes [CSDK-611](https://revenuecats.atlassian.net/browse/CSDK-611). 

This is the approach that stores the last customer info on the dart side and sends it to the first listener added. This approach, while not being the cleanest, wouldn't require API changes as I talked in the ticket. I opened this PR to discuss if we want to follow this approach

### DIFF
--- a/lib/purchases_flutter.dart
+++ b/lib/purchases_flutter.dart
@@ -43,13 +43,17 @@ class Purchases {
   static final Set<ReadyForPromotedProductPurchaseListener>
       _readyForPromotedProductPurchaseListeners = {};
 
+  static CustomerInfo? _lastReceivedCustomerInfo;
+
   static final _channel = const MethodChannel('purchases_flutter')
     ..setMethodCallHandler((call) async {
       switch (call.method) {
         case 'Purchases-CustomerInfoUpdated':
+          final args = Map<String, dynamic>.from(call.arguments);
+          final customerInfo = CustomerInfo.fromJson(args);
+          _lastReceivedCustomerInfo = customerInfo;
           for (final listener in _customerInfoUpdateListeners) {
-            final args = Map<String, dynamic>.from(call.arguments);
-            listener(CustomerInfo.fromJson(args));
+            listener(customerInfo);
           }
           break;
         case 'Purchases-ReadyForPromotedProductPurchase':
@@ -159,8 +163,15 @@ class Purchases {
   /// [customerInfoUpdateListener] CustomerInfo update listener.
   static void addCustomerInfoUpdateListener(
     CustomerInfoUpdateListener customerInfoUpdateListener,
-  ) =>
-      _customerInfoUpdateListeners.add(customerInfoUpdateListener);
+  ) {
+    _customerInfoUpdateListeners.add(customerInfoUpdateListener);
+    if (_customerInfoUpdateListeners.length == 1) {
+      final lastReceivedCustomerInfo = _lastReceivedCustomerInfo;
+      if (lastReceivedCustomerInfo != null) {
+        customerInfoUpdateListener(lastReceivedCustomerInfo);
+      }
+    }
+  }
 
   /// Removes a given CustomerInfoUpdateListener
   ///


lib/purchases_flutter.dart
```diff
@@ -43,13 +43,17 @@ class Purchases {
...
       _readyForPromotedProductPurchaseListeners = {};
 
+  static CustomerInfo? _lastReceivedCustomerInfo;
+
   static final _channel = const MethodChannel('purchases_flutter')
     ..setMethodCallHandler((call) async {
       switch (call.method) {
         case 'Purchases-CustomerInfoUpdated':
+          final args = Map<String, dynamic>.from(call.arguments);
+          final customerInfo = CustomerInfo.fromJson(args);
```
tonidero, line 53: I moved this outside the for loop. It didn't make sense to perform this operation on every iteration.

lib/purchases_flutter.dart
```diff
@@ -159,8 +163,15 @@ class Purchases {
...
-      _customerInfoUpdateListeners.add(customerInfoUpdateListener);
+  ) {
+    _customerInfoUpdateListeners.add(customerInfoUpdateListener);
+    if (_customerInfoUpdateListeners.length == 1) {
+      final lastReceivedCustomerInfo = _lastReceivedCustomerInfo;
+      if (lastReceivedCustomerInfo != null) {
+        customerInfoUpdateListener(lastReceivedCustomerInfo);
+      }
+    }
+  }
```
tonidero, line 174: I haven't added tests yet. I wanted to validate if we're ok with this approach first.

lib/purchases_flutter.dart
```diff
@@ -43,13 +43,17 @@ class Purchases {
   static final Set<ReadyForPromotedProductPurchaseListener>
       _readyForPromotedProductPurchaseListeners = {};
 
+  static CustomerInfo? _lastReceivedCustomerInfo;
```
tonidero, line 46: I don't like static variables like this, that's why I think this is not the preferred approach. But if we consider it a temporal solution and plan to solve it better in the future, I would be ok with it.

lib/purchases_flutter.dart
tonidero, line 174: Also note that this is actually a behavior change, so it should be prominently displayed in the changelog.

lib/purchases_flutter.dart
```diff
@@ -159,8 +163,15 @@ class Purchases {
   /// [customerInfoUpdateListener] CustomerInfo update listener.
   static void addCustomerInfoUpdateListener(
     CustomerInfoUpdateListener customerInfoUpdateListener,
-  ) =>
-      _customerInfoUpdateListeners.add(customerInfoUpdateListener);
+  ) {
+    _customerInfoUpdateListeners.add(customerInfoUpdateListener);
+    if (_customerInfoUpdateListeners.length == 1) {
```
NachoSoto, line 168: Why do this only if there's one listener?

NachoSoto (commented):
I think this makes sense.
How does this interact with the fact that iOS already does this? Would it send it twice?

tonidero:
What do you think of this approach @RevenueCat/coresdk? I'm ok with this being a temporal solution, but in the long term, i would vote to move towards a different system that allows to add multiple listeners in the native layer.

lib/purchases_flutter.dart
tonidero: I believe the preferred approach was to only send the customer info on the first added listener. But if we think we should send it to all listeners as soon as we add them, I can change it easily.

tonidero:
> How does this interact with the fact that iOS already does this? Would it send it twice?

If I'm not wrong, in iOS we can only have one delegate, and it would send the customer info as soon as the delegate is set. In flutter, we set a listener as soon as we `configure` the SDK and then, we support adding multiple listeners but since those are added after setting the iOS/Android delegates, we miss that initial customer info. This PR is a possible approach so we can propagate that initial customer info to the listeners as soon as they are added. More details in the Jira ticket. Do lmk if I'm missing something!

lib/purchases_flutter.dart
tonidero: Note that this will also send it when we remove all listeners and add one again.

NachoSoto: > But if we think we should send it to all listeners as soon as we add them, I can change it easily.

I think so. That's how it works in iOS with the delegate or with the `CustomerInfo` `AsyncStream`.

NachoSoto:
> If I'm not wrong, in iOS we can only have one delegate, and it would send the customer info as soon as the delegate is set. In flutter, we set an internal listener in the native layer as soon as we configure the SDK and then, 

Oh that makes sense then 👍🏻

lib/purchases_flutter.dart
tonidero: Well, in iOS we can only have one delegate. It would send the last customer info if they changed it, but we can't have multiple.

That said, I'm not sure of the use cases for people to add multiple listeners, so I'm not sure what the desired behavior would be... But well, it might make sense for consistency to send the customer info for all listeners as soon as they are added, so I will change it. CC @aboedo @vegaro

lib/purchases_flutter.dart
NachoSoto: Yup this makes sense 👍🏻

NachoSoto (approved):
Looks good for now, you're going to add tests in this PR or another one?

lib/purchases_flutter.dart
vegaro: yes, thankfully it said in the docs that this was the expected behavior, so I think everyone was assuming this was the behavior anyways

vegaro (approved):

tonidero:
> you're going to add tests in this PR or another one? 

Ah will add them in a followup PR since this has already been reviewed https://github.com/RevenueCat/purchases-flutter/pull/580

PR #360, opened 2022-04-30
[WIP] Update to phc 3.0.0 (Part 10): BeginRefundRequest

vegaro:
WIP
[CF-611]

[CF-611]: https://revenuecats.atlassian.net/browse/CF-611?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #577, opened 2023-01-16
Add beginRefundRequest APIs for iOS 15+

tonidero:
Adds [CSDK-94](https://revenuecats.atlassian.net/browse/CSDK-94)

This PR adds 3 new methods to the flutter API that will only be available in iOS 15+.
- `beginRefundRequestForActiveEntitlement`
- `beginRefundRequestForProduct`
- `beginRefundRequestForEntitlement`
In Android/iOS < 15, we will throw an `UnsupportedPlatformException`.

In this PR, I'm also updating the error codes to match the latest in PHC.

This PR is partially based on the work on https://github.com/RevenueCat/purchases-flutter/pull/360

[CSDK-94]: https://revenuecats.atlassian.net/browse/CSDK-94?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ios/Classes/PurchasesFlutterPlugin.m
+++ b/ios/Classes/PurchasesFlutterPlugin.m
@@ -168,6 +168,14 @@ NSString *PurchasesReadyForPromotedProductPurchaseEvent = @"Purchases-ReadyForPr
     } else if ([@"canMakePayments" isEqualToString:call.method]) {
           NSArray<NSNumber*> *features = arguments[@"features"];
           [self canMakePaymentsWithFeatures:features result:result];
+    } else if ([@"beginRefundRequestForActiveEntitlement" isEqualToString:call.method]) {
+        [self beginRefundRequestForActiveEntitlementWithResult:result];
+    } else if ([@"beginRefundRequestForProduct" isEqualToString:call.method]) {
+        NSString *productID = arguments[@"productIdentifier"];
+        [self beginRefundRequestForProduct:productID result:result];
+    } else if ([@"beginRefundRequestForEntitlement" isEqualToString:call.method]) {
+        NSString *entitlementID = arguments[@"entitlementIdentifier"];
+        [self beginRefundRequestForEntitlement:entitlementID result:result];
     } else if ([@"getPromotionalOffer" isEqualToString:call.method]) {
         NSString *productIdentifier = arguments[@"productIdentifier"];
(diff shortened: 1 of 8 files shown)


ios/Classes/PurchasesFlutterPlugin.m
```diff
@@ -526,11 +560,23 @@ - (void)rejectWithResult:(FlutterResult)result error:(RCErrorContainer *)errorCo
     };
 }
 
+- (void (^)(RCErrorContainer *))getBeginRefundResponseCompletionBlock:(FlutterResult)result {
+    return ^(RCErrorContainer * _Nullable error) {
+        if (error == nil) {
+            result(@0);
+        } else if ([error.info[@"userCancelled"] isEqual: @YES]) {
```
tonidero, line 567: So, I originally wanted to just expose the error in the flutter side, but after testing it out, accessing this property on the error in the flutter side was very verbose and difficult... Due to that, I decided to just mimick the API in iOS which gives an enum with 3 possible values. 

The part I don't like of this is that now we have to process the error from PHC to map to the enum, instead of just returning an integer from PHC and avoiding this extra processing in all hybrids... So I was thinking about changing PHC but that would technically be a breaking change there... So in the end I decided to map the existing response. Lmk what you think!

lib/errors.dart
```diff
@@ -80,6 +80,39 @@ enum PurchasesErrorCode {
...
+
+  /// Requests to RevenueCat are being blocked. See: https://rev.cat/dnsBlocking for more info.
+  apiEndpointBlocked,
+
+  /// The information associated with this PromotionalOffer is not valid.
+  /// See https://rev.cat/ios-subscription-offers for more info.
+  invalidPromotionalOfferError,
+
+  /// Error performing request because the internet connection appears to be offline.
+  offlineConnectionError
```
tonidero, line 115: This was not directly related to this, but one of the errors is `beginRefundRequestError` so I decided to add all the ones currently in PHC

ios/Classes/PurchasesFlutterPlugin.m
NachoSoto: This makes sense. PHC is generic but here you're making the best API for this particular platform.

NachoSoto (approved):

PR #582, opened 2023-01-19
Fix macOS compilation

vegaro:
This job when preparing the release was failing https://app.circleci.com/pipelines/github/RevenueCat/purchases-flutter/1439/workflows/9771e1cc-1bae-4725-8f6c-234f202fcf81/jobs/5916

It was running on mac so I noticed the compilation is broken in mac.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -288,7 +288,7 @@ workflows:
       not:
         equal: [ scheduled_pipeline, << pipeline.trigger_source >> ]
     jobs:
-      - ios-integration-test: *release-branches
+      - ios-integration-test
       - android-integration-test-build: *release-branches
       - run-firebase-tests:
           requires:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -288,7 +288,7 @@ workflows:
       not:
         equal: [ scheduled_pipeline, << pipeline.trigger_source >> ]
     jobs:
-      - ios-integration-test: *release-branches
+      - ios-integration-test
```
vegaro, line 291: this is to test if they are working now

NachoSoto (approved):

tonidero:
Oh interesting... I thought the `@available` should have avoided this, but I guess these methods are not part of the API in macos... Thanks for fixing it! Also makes me wonder if we should run integration tests more often...

tonidero:
Oh I understand now, I added [this](https://github.com/RevenueCat/purchases-hybrid-common/blob/main/ios/PurchasesHybridCommon/PurchasesHybridCommon/CommonFunctionality.swift#L124), so it's normal that it's not available in mac. Again, thanks for fixing it!

PR #584, opened 2023-01-19
Fix integration tests

vegaro:
- Added macos tests
- Fixed iOS tests


.circleci/config.yml
```diff
@@ -289,6 +339,7 @@ workflows:
         equal: [ scheduled_pipeline, << pipeline.trigger_source >> ]
     jobs:
       - ios-integration-test: *release-branches
+      - macos-integration-test: *release-branches
```
tonidero, line 342: We should make the hold job depend on this as well

tonidero (approved):
Looks great, thanks!

PR #585, opened 2023-01-20
Fix error code parsing from native layer

tonidero:
During #577, we updated the error codes in Flutter to match those in iOS. However, I didn't notice that we are missing error code 27 in iOS, which was causing the parsing to be incorrect in the flutter layer. This PR fixes that issue

### DIFF
--- a/lib/errors.dart
+++ b/lib/errors.dart
@@ -89,6 +89,9 @@ enum PurchasesErrorCode {
   /// This is a required property and likely an AppStore quirk that it is missing.
   productDiscountMissingIdentifierError,
 
+  /// Unknown error
+  unknownNonNativeError,
+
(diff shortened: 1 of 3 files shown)


lib/errors.dart
```diff
@@ -89,6 +89,9 @@ enum PurchasesErrorCode {
   /// This is a required property and likely an AppStore quirk that it is missing.
   productDiscountMissingIdentifierError,
 
+  /// Unknown error
+  unknownNonNativeError,
```
tonidero, line 93: I added a new error so,

I thought adding a new error was better than just processing the error code in PHC with the missing error, since if any dev was using the enum indexes for this (which they shouldn't) it would still match the error codes. This new error should never happen though. Let me know what you think!

NachoSoto (approved):
I thought I approved this already sorry!

PR #587, opened 2023-01-23
Change default usesStoreKit2IfAvailable to false

vegaro:
Changed usesStoreKit2IfAvailable default to match iOS


tonidero (approved):

tonidero:
Ah looks like we need to update the tests

PR #591, opened 2023-01-24
Added Purchases.setLogHandler

vegaro:
[CSDK-548](https://revenuecats.atlassian.net/browse/CSDK-548)

[CSDK-548]: https://revenuecats.atlassian.net/browse/CSDK-548?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
```diff
@@ -610,6 +614,16 @@ private void close(final Result result) {
         result.success(null);
     }
 
+    private void setLogHandler(final Result result) {
+        CommonKt.setLogHandler(logData -> {
+            if (channel != null && this.activity != null) {
+                this.activity.runOnUiThread(() -> channel.invokeMethod(LOG_HANDLER_EVENT, logData));
```
vegaro, line 620: Not doing this causes an exception complaining that invokeMethod needs to be called on the UiThread

test/purchases_flutter_test.dart
```diff
@@ -76,17 +83,26 @@ void main() {
     );
   });
 
-  test('addCustomerInfoUpdateListener does not call listener if it does not have one', () async {
+  test(
+      'addCustomerInfoUpdateListener does not call listener if it does not have one',
```
tonidero, line 87: Whoops, I'm guessing we're going over the line limit. Should this be caught by the linter?

test/purchases_flutter_test.dart
```diff
@@ -759,4 +800,22 @@ void main() {
       )
     ]);
   });
+
+  test('setupPurchases with amazon', () async {
```
tonidero, line 804: Hmm probably test name is from a copy/paste?

test/purchases_flutter_test.dart
```diff
@@ -41,6 +41,11 @@ void main() {
     'nonSubscriptionTransactions': []
   };
 
+  final mockLog = {
+    'logLevel': 'DEBUG',
+    'message': 'A log sent',
+  };
```
tonidero, line 47: Hmm since this is currently only used in one test, should we move it inside?

lib/purchases_flutter.dart
```diff
@@ -45,6 +50,8 @@ class Purchases {
 
   static CustomerInfo? _lastReceivedCustomerInfo;
 
+  static LogHandler? _logHandler;
```
tonidero, line 53: Hmm we currently don't log much/anything in the hybrids. If we do in the future, we should probably redirect those logs over here as well... Nothing to do here right now, just something to think about.

tonidero (commented):
Looking good!

android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
vegaro: I didnt' notice! but probably worth it to change threads there. I will add it

android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
```diff
@@ -610,6 +614,16 @@ private void close(final Result result) {
         result.success(null);
     }
 
+    private void setLogHandler(final Result result) {
+        CommonKt.setLogHandler(logData -> {
+            if (channel != null && this.activity != null) {
```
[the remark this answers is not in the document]
vegaro: I could just create a new Handler instead and not use the Activity, that might be safer

test/purchases_flutter_test.dart
vegaro: this is interesting, idk why but the linter didn't catch this. When I ran `pub publish --dry-run` it formatted these and also gave me some errors from `dart analyze`. I ran `dart analyze` though and got so many other issues that I think `pub publish` is using both a dart analyze and linter custom configuration

lib/purchases_flutter.dart
```diff
@@ -770,6 +780,25 @@ class Purchases {
...
 
+  /// Set a custom log handler for redirecting logs to your own logging system.
+  /// By default, this sends info, warning, and error messages.
+  /// If you wish to receive Debug level messages, see [setLogLevel].
+  ///
+  /// [logHandler] It will get called for each log event.
+  /// Implement this function to redirect the log to your own logging system
+  static Future<void> setLogHandler(LogHandler logHandler) async {
+    _logHandler = logHandler;
+    return await _channel.invokeMethod('setLogHandler');
```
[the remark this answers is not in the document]
vegaro: I could make it accept a nullable LogHandler?

lib/purchases_flutter.dart
```diff
@@ -770,6 +780,25 @@ class Purchases {
...
+  }
+
+  static void handleLogHandlerEvent(MethodCall call) {
+    final args = Map<String, dynamic>.from(call.arguments);
+    final logLevelName = args['logLevel'];
+    final logLevel = LogLevel.values.firstWhere((e) => e.name.toUpperCase() == logLevelName);
```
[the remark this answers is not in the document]
vegaro: I think I am going to pick `info`

lib/purchases_flutter.dart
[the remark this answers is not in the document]
tonidero: Well, we should just be careful setting a null LogHandler reenables default logs on the native SDKs, but yeah that could be a solution.

android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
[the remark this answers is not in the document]
tonidero: Yeah, I think that makes sense 👍

android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
```diff
@@ -342,7 +350,8 @@ private void setupPurchases(String apiKey, String appUserID, @Nullable Boolean o
     private void setUpdatedCustomerInfoListener() {
         Purchases.getSharedInstance().setUpdatedCustomerInfoListener(customerInfo -> {
             if (channel != null) {
-                channel.invokeMethod(CUSTOMER_INFO_UPDATED, CustomerInfoMapperKt.map(customerInfo));
+                Map<String, Object> customerInfoMap = CustomerInfoMapperKt.map(customerInfo);
+                runOnUiThread(() -> channel.invokeMethod(CUSTOMER_INFO_UPDATED, customerInfoMap));
```
tonidero, line 354: This is a behavior change, so we should mention it somewhere in the changelog maybe? (same for the ones in iOS)

tonidero (approved):
Looks good! We should only be careful about that behavior change but I think it should be ok.

android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
vegaro: sure. In any case, I was getting crashes when sending the other event not in the UI thread, so I think this would actually be a bug fix more than a behavior change

vegaro: I will extract this to a different PR so it's clear

PR #609, opened 2023-02-10
Add `syncObserverModeAmazonPurchase`

vegaro:
[https://revenuecats.atlassian.net/browse/CSDK-579]

### DIFF
--- a/android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
+++ b/android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
@@ -321,6 +321,15 @@ public class PurchasesFlutterPlugin implements FlutterPlugin, MethodCallHandler,
             case "setLogHandler":
                 setLogHandler(result);
                 break;
+            case "syncObserverModeAmazonPurchase":
+                String productID = call.argument("productID");
+                String receiptID = call.argument("receiptID");
+                String amazonUserID = call.argument("amazonUserID");
+                String isoCurrencyCode = call.argument("isoCurrencyCode");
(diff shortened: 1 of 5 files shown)


android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
```diff
@@ -449,6 +458,14 @@ private void setDebugLogsEnabled(boolean enabled, final Result result) {
         result.success(null);
     }
 
+    private void syncObserverModeAmazonPurchase(String productID, String receiptID,
+                                                String amazonUserID, String isoCurrencyCode,
+                                                Double price, final Result result) {
```
tonidero, line 463: Nitpick: I kinda prefer having every parameter in its own line.

api_tester/lib/api_tests/purchases_flutter_api_test.dart
```diff
@@ -415,4 +415,11 @@ class _PurchasesFlutterApiTest {
       entitlement
     );
   }
+
+  void _checkSyncObserverModeAmazonPurchase() async {
+    Future<void> future = Purchases.syncObserverModeAmazonPurchase(
+        "productID", "receiptID", "amazonUserID", "isoCurrencyCode", 3.4);
+    future = Purchases.syncObserverModeAmazonPurchase(
+        "productID", "receiptID", "amazonUserID", null, null);
```
tonidero, line 423: Nitpick: One possible way to avoid calling this same method twice is adding the parameters as parameters of the check function with the correct type (including nullability). I think it's also more visual.

tonidero (approved):
LGTM!

PR #610, opened 2023-02-10
fix(example): fix LogLevel debug enum

mhadaily:
based on this enum https://github.com/RevenueCat/purchases-flutter/blob/fd79ff71f277e87a5c760f4582f2803a05c74d05/lib/purchases_flutter.dart#L963

The example doesn't run as the enum is passed wrong. all tests will pass though. 

  - [x] A description of what and why you are contributing, even if it's trivial.

  - [ ] The issue number(s) or PR number(s) in the description if you are contributing in response to those.

  - [ ] If applicable, unit tests.


NachoSoto (approved):
Thanks!

PR #612, opened 2023-02-14
Deprecate revenueCatId in favor of transactionIdentifier

tonidero:
In this PR we deprecate `revenueCatId` in favor of `transactionIdentifier`. The PHC change was done a long time ago: https://github.com/RevenueCat/purchases-hybrid-common/pull/211

### DIFF
--- a/api_tester/lib/api_tests/models/store_transaction_api_test.dart
+++ b/api_tester/lib/api_tests/models/store_transaction_api_test.dart
@@ -12,11 +12,13 @@ class _StoreTransactionApiTest {
     Map<String, dynamic> json = transaction.toJson();
   }
 
-  void _checkConstructor(String transactionIdentifier, String revenueCatIdentifier, 
-      String productIdentifier, String purchaseDate) {
+  void _checkConstructor(
+      String transactionIdentifier,
+      String revenueCatIdentifier,
+      String productIdentifier,
+      String purchaseDate) {
     StoreTransaction transaction =
-        StoreTransaction(transactionIdentifier, revenueCatIdentifier, 
-            productIdentifier, purchaseDate);
+        StoreTransaction(revenueCatIdentifier, productIdentifier, purchaseDate);
   }
 
   void _checkProperties(StoreTransaction transaction) {
--- a/api_tester/lib/api_tests/models/store_transaction_api_test.dart
+++ b/api_tester/lib/api_tests/models/store_transaction_api_test.dart
@@ -11,10 +11,11 @@ class _StoreTransactionApiTest {
     Map<String, dynamic> json = transaction.toJson();
   }
 
-  void _checkConstructor(String revenueCatIdentifier, String productIdentifier,
-      String purchaseDate) {
+  void _checkConstructor(String transactionIdentifier, String revenueCatIdentifier, 
+      String productIdentifier, String purchaseDate) {
     StoreTransaction transaction =
-        StoreTransaction(revenueCatIdentifier, productIdentifier, purchaseDate);
+        StoreTransaction(transactionIdentifier, revenueCatIdentifier, 
+            productIdentifier, purchaseDate);
   }
 
   void _checkProperties(StoreTransaction transaction) {
--- a/api_tester/lib/api_tests/models/store_transaction_api_test.dart
+++ b/api_tester/lib/api_tests/models/store_transaction_api_test.dart
@@ -2,6 +2,7 @@ import 'package:purchases_flutter/object_wrappers.dart';
 
 // ignore_for_file: unused_element
 // ignore_for_file: unused_local_variable
+// ignore_for_file: deprecated_member_use
 class _StoreTransactionApiTest {
   void _checkFromJsonFactory(Map<String, dynamic> json) {
     StoreTransaction transaction = StoreTransaction.fromJson(json);
--- a/api_tester/lib/api_tests/models/store_transaction_api_test.dart
+++ b/api_tester/lib/api_tests/models/store_transaction_api_test.dart
@@ -18,6 +18,7 @@ class _StoreTransactionApiTest {
   }
 
   void _checkProperties(StoreTransaction transaction) {
+    String transactionIdentifier = transaction.transactionIdentifier;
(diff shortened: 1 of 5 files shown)


lib/models/store_transaction.dart
```diff
@@ -9,10 +9,14 @@ part 'store_transaction.g.dart';
...
     /// RevenueCat Id associated to the transaction.
+    @JsonKey(name: 'transactionIdentifier') String transactionIdentifier,
+
+    /// Deprecated: Use transactionIdentifier instead.
+    @Deprecated('Use transactionIdentifier instead.')
     @JsonKey(name: 'revenueCatId') String revenueCatIdentifier,
 
     /// Product Id associated with the transaction.
-    @JsonKey(name: 'productId') String productIdentifier,
+    @JsonKey(name: 'productIdentifier') String productIdentifier,
```
tonidero, line 19: `productId` was also deprecated, but since flutter was using `productIdentifier`, devs shouldn't see any changes. I only changed the mapping to use the new name.

api_tester/lib/api_tests/models/store_transaction_api_test.dart
```diff
@@ -11,13 +11,15 @@ class _StoreTransactionApiTest {
...
   }
 
-  void _checkConstructor(String revenueCatIdentifier, String productIdentifier,
-      String purchaseDate) {
+  void _checkConstructor(String transactionIdentifier, String revenueCatIdentifier, 
+      String productIdentifier, String purchaseDate) {
     StoreTransaction transaction =
-        StoreTransaction(revenueCatIdentifier, productIdentifier, purchaseDate);
+        StoreTransaction(transactionIdentifier, revenueCatIdentifier, 
+            productIdentifier, purchaseDate);
```
tonidero, line 18: This is actually a breaking change, but devs shouldn't generally be using this, other than maybe in tests... Any thoughts on what to do here? @RevenueCat/coresdk. IMO, I think it would be ok to ship this and mention it in the changelog, but I recognize a few people could potentially be affected...

NachoSoto: Who knows if somebody is using it, so I'd avoid breaking it. Is it possible to keep the older version for backwards compatibility?

aboedo: +1, if we break even a single developer's app, even in debug, accidentally, it would be a failure on our side

aboedo: reminder: [At RevenueCat, a customer can’t hold the phone the wrong way](https://www.revenuecat.com/blog/company/why-were-customer-obsessed-and-you-should-be-too/)
even if they _shouldn't_ be using it, if they _can_, we shouldn't break things

tonidero: Good point! I will see if we can use custom constructors to keep backwards compatibility for this. Thanks!

lib/models/store_transaction.dart
```diff
@@ -7,17 +7,41 @@ part 'store_transaction.g.dart';
 
 /// Represents a purchase transaction
 class StoreTransaction with _$StoreTransaction {
-  const factory StoreTransaction(
+  const factory StoreTransaction._create(
```
tonidero, line 10: Here we are adding a name `_create` to the "full" constructor, making it private, so it won't be accessible by devs. Then, we create a new unnamed factory pointing to this one with only the previous parameters. This way we won't break existing users of the previous constructor. 

We should note that new constructors won't be available for devs so:
- This system would break once we don't have just a basic rename of a property, as in, if we add a new property in the future, unless we want to make it nullable, the deprecated constructor won't know what value to use.
- People can still create instances through the `fromJson` constructor.
- We should consider making all constructors private in the next major.

Let me know what you think of this approach @RevenueCat/coresdk

NachoSoto: It makes sense to me, but I know virtually 0 about Dart so I'll leave others to review this.

vegaro: We got feedback in the past that having no constructors and only the `fromJson` constructor made testing hard since instantiating objects was cumbersome, so I am not sure this is the right approach to be honest, but I also don't have a better suggestion 🤔

tonidero: Hmm that is tricky since it would mean we can't ever modify these models easily. So either we break the constructor API or we add constructors for every new field with some default value, which might not be straightforward every time, or this solution, which could also break the fromJson constructor if we change the body and we don't handle missing parameters...

I'm thinking we could also make the constructor an experimental API to mark is as it could potentially change at anytime: https://api.flutter.dev/flutter/meta/experimental-constant.html . It's not great since it would still be breaking... But people would be more aware of it at least... Thoughts @vegaro ?

tonidero: Hmm so following the same process we did in Android (just add a new deprecated constructor) won't work in flutter. In dart you can only have one default/unnamed constructor. So I think if we want to have a public constructor with always up-to-date information, we can make the full constructor public and mark it as experimental as I suggested above. Lmk what you think @RevenueCat/coresdk

vegaro: yeah... I think that's probably the best to be honest

I was trying to think if maybe we could make instead of constructors, static functions that create objects. Like factories. So we could have a private constructor but have public factory methods (like the fromJson one) that accept the properties of the object as parameters. It would be easier to add new parameters to these factory methods.

tonidero: Well, that's kinda what factories do right? The thing is that to keep backwards compatibility, we need to keep the unnamed constructor be the one that existed until now (we can only have one). The latest changes are basically:
- Adding the unnamed factory constructor deprecated with the signature that existed until now
- Make the full constructor/factory name `create` with a comment indicating that it's subject to changes. So it would be used like `StoreTransaction.create(...)`.  

If we want to add more constructors in the future, we would need to add more factory methods with different names. Lmk what you think of this approach!

vegaro (approved):

PR #641, opened 2023-03-14
[BC5] Add subscription options support 

joshdholtz:
## Motivation

[CF-1271](https://revenuecats.atlassian.net/browse/CF-1271)
[CF-1269](https://revenuecats.atlassian.net/browse/CF-1269)
[CF-1270](https://revenuecats.atlassian.net/browse/CF-1270)

## Description

- Created a new `v5-MIGRATION.md` doc
- Adds `SubscriptionOption`, `PricingPhase`, `Period`, `Price`, `Unit`, and `RecurrenceMode` models
  - Added API tests
  - Added to the migration doc
- Adds `purchaseSubscriptionOption()` method
  - Added API tests
  - Added to the migration doc
- Adds `isPersonalizedPrice` to `purchaseProduct()`, `purchasePackage()`, and `purchaseSubscriptionOption()`

⚠️ There will be a follow up PR for the `// TODO: Rename`. `UpgradeInfo` is going to be changed to `ProductChange` and that is too big for this PR 🤷‍♂️ 

## Demo

It's a pretty ugly demo but... this iterates over all the packages in an offering and shows:
- A button to buy the package (which buys the default option)
- Buttons for each subscription option (if its a subscription)

This demo also shows purchasing a one-time purchase and prepaid.

https://user-images.githubusercontent.com/401294/225200995-2031dbba-d340-45fb-8d52-311171992834.mov

[CF-1271]: https://revenuecats.atlassian.net/browse/CF-1271?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CF-1269]: https://revenuecats.atlassian.net/browse/CF-1269?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CF-1270]: https://revenuecats.atlassian.net/browse/CF-1270?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
```diff
@@ -151,17 +152,27 @@ public void onMethodCall(@NonNull MethodCall call, @NonNull Result result) {
                 break;
             case "purchaseProduct":
                 String productIdentifier = call.argument("productIdentifier");
-                String oldSKU = call.argument("oldSKU");
-                Integer prorationMode = call.argument("prorationMode");
+                String oldSKU = call.argument("oldSKU"); // TODO: Rename
+                Integer prorationMode = call.argument("prorationMode"); // TODO: Rename
+                Boolean isPersonalizedPrice = call.argument("isPersonalizedPrice"); // TODO: Rename
```
joshdholtz, line 157: Going to handle these renames in a followup PR since they involve some kind of deprecations / doc changes

lib/purchases_flutter.dart
```diff
@@ -306,6 +306,33 @@ class Purchases {
     return customerInfo;
   }
 
+  /// Google Play only.
```
joshdholtz, line 309: Is this an okay way of doing this? Or should we also do a platform check in the code?

I mean... Getting a `SubscriptionOption` isn't even possible on iOS or Amazon so this can't really be called anyway 🤷‍♂️

lib/purchases_flutter.dart
```diff
@@ -794,7 +821,10 @@ class Purchases {
   static void handleLogHandlerEvent(MethodCall call) {
     final args = Map<String, dynamic>.from(call.arguments);
     final logLevelName = args['logLevel'];
-    final logLevel = LogLevel.values.firstWhere((e) => e.name.toUpperCase() == logLevelName, orElse: () => LogLevel.info);
+    final logLevel = LogLevel.values.firstWhere(
+      (e) => e.name.toUpperCase() == logLevelName,
+      orElse: () => LogLevel.info,
+    );
```
joshdholtz, line 827: Formatter did this for me 🙃

tonidero (commented):
Nothing really blocking right now. Left a few comments.

tonidero:
Also, looks like freezed failed.

android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
```diff
@@ -448,6 +456,29 @@ private void purchasePackage(final String packageIdentifier,
...
     }
 
+    private void purchaseSubscriptionOption(final String productIdentifier,
+                                            final String optionIdentifier,
+                                            final String googleOldProductId,
+                                            @Nullable final Integer googleProrationMode,
+                                            @Nullable final Boolean googleIsPersonalizedPrice,
+                                            final Result result) {
+        // TODO: Convert this somehow
+        GoogleProrationMode googleProrationModeEnum = null;
```
beylmk, line 466: you can check out the canMakePayments `BillingFeature` conversion, might be similar..

lib/models/price_wrapper.dart
```diff
@@ -0,0 +1,29 @@
...
+    /// Formatted price of the item, including its currency sign. For example $3.00
+    @JsonKey(name: 'formatted') String formatted,
+
+    /// Price in micro-units, where 1,000,000 micro-units equal one unit of the currency.
+    ///
+    /// For example, if price is "€7.99", price_amount_micros is 7,990,000. This value represents
+    /// the localized, rounded price for a particular currency.
+    @JsonKey(name: 'amountMicros') int amountMicros,
+
+    /// Returns ISO 4217 currency code for price and original price.
```
beylmk, line 20: isn't it 8601?

lib/models/pricing_phase_wrapper.dart
```diff
@@ -0,0 +1,30 @@
...
+
+@freezed
+
+/// Contains all the details associated with a PricingPhase
+class PricingPhase with _$PricingPhase {
+  const factory PricingPhase(
+    /// Billing period for which the PricingPhase applies
+    @JsonKey(name: 'billingPeriod') Period billingPeriod,
+
+    /// Recurrence mode of the PricingPhase
```
beylmk, line 17: examples might be good again... are we adding the RecurrenceMode enum too ?

lib/models/subscription_option_wrapper.dart
```diff
@@ -0,0 +1,58 @@
...
+
+import 'period_wrapper.dart';
+import 'pricing_phase_wrapper.dart';
+
+part 'subscription_option_wrapper.freezed.dart';
+part 'subscription_option_wrapper.g.dart';
+
+@freezed
+
+/// Contains all details associated with a SubscriptionOption
```
beylmk, line 11: can we call out at the class-level that this is google-only?

beylmk (approved):
mostly just comments on the docs! awesome work :)

android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
```diff
@@ -170,6 +170,14 @@ public void onMethodCall(@NonNull MethodCall call, @NonNull Result result) {
                 isPersonalizedPrice = call.argument("isPersonalizedPrice"); // TODO: Rename
                 purchasePackage(packageIdentifier, offeringIdentifier, oldSKU, prorationMode, isPersonalizedPrice, result);
                 break;
+            case "purchaseSubscriptionOption":
+                productIdentifier = call.argument("productIdentifier");
+                String optionIdentifier = call.argument("optionIdentifier");
+                oldSKU = call.argument("oldSKU"); // TODO: Rename
+                prorationMode = call.argument("prorationMode"); // TODO: Rename
+                isPersonalizedPrice = call.argument("isPersonalizedPrice"); // TODO: Rename
```
vegaro, line 178: This is never sent from the Dart side

lib/models/period_wrapper.dart
```diff
@@ -0,0 +1,19 @@
...
+@freezed
+
+/// Contains all the details associated with a Period
+class Period with _$Period {
+  const factory Period(
+    /// The number of period units
+    @JsonKey(name: 'periodUnit') String unit,
+
+    /// The increment of time that a subscription period is specified in
+    @JsonKey(name: 'periodNumberOfUnits') int value,
```
vegaro, line 15: We also have a field `iso8601` in purchases-android. Should we add it here now that we are adding this file

lib/models/price_wrapper.dart
vegaro: That's for time. 4217 is for currency codes

lib/purchases_flutter.dart
vegaro: I would do whatever we do in `beginRefundRequestForActiveEntitlement`, which is throw an `UnsupportedPlatformException`. That's the most recent case where we've had to deal with one-platform functions and I think it's the most predictable and descriptive way.

lib/purchases_flutter.dart
```diff
@@ -306,6 +306,33 @@ class Purchases {
...
+  /// [PurchasesErrorCode.purchaseCancelledError] to check if the user cancelled
+  /// the purchase.
+  ///
+  /// [packageToPurchase] The Package you wish to purchase
+  ///
+  /// [upgradeInfo] Android only. Optional UpgradeInfo you wish to upgrade from
+  /// containing the oldSKU and the optional prorationMode.
+  static Future<CustomerInfo> purchaseSubscriptionOption(
+    SubscriptionOption subscriptionOption, {
+    UpgradeInfo? upgradeInfo,
```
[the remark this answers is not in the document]
vegaro: UpgradeInfo is optional here so I think this function is enough no?

android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
joshdholtz: Not yet 😇 There is a separate ticket for this one but I should probably just do it in this PR

revenuecat_examples/purchase_tester/lib/src/app.dart
```diff
@@ -194,7 +206,46 @@ class _PurchaseButton extends StatelessWidget {
...
+            } else if (errorCode == PurchasesErrorCode.paymentPendingError) {
+              print('Payment is pending');
+            }
+          }
+          return const InitialScreen();
+        },
+        child:
+            Text('Buy Option: - (${option.id}\n${option.pricingPhases.map((e) {
+          return '${e.price.formatted} for ${e.billingPeriod.value} ${e.billingPeriod.unit}';
+        }).join(' -> ')})'),
```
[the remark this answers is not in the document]
joshdholtz: I caaannn but its so ugly 😛 Don't judge my terrible Flutter UI skills but going to make a JIRA tasks to improve this (if it doesn't exist already)

lib/purchases_flutter.dart
joshdholtz: I think I'm confused on how and where this should be handle (especially since this is specifically for Google Android and not Amazon).

1) Is it best to do the check at the start of this method before invoking anything? If so. what is the best way to check if Amazon?
2) Or do we do this deeper into the stack? Like on an iOS and Android level? Do I need to add this method to iOS?

@tonidero @vegaro Not _as_ familiar here so any suggestion would be appreciated 😇

joshdholtz:
> Also, looks like freezed failed.

@tonidero Fixed! Apparently I had to update my local flutter? 🤷‍♂️

api_tester/lib/api_tests/purchases_flutter_api_test.dart
```diff
@@ -87,6 +87,15 @@ class _PurchasesFlutterApiTest {
         await Purchases.purchasePackage(package, upgradeInfo: upgradeInfo);
   }
 
+  void _checkPurchaseSubscriptionOption(
+      SubscriptionOption subscriptionOption) async {
+    UpgradeInfo? upgradeInfo;
+    PurchaseType purchaseType = PurchaseType.subs;
```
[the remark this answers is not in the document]
joshdholtz: Ooops, that was left over from a copy/paste 🤦‍♂️

lib/purchases_flutter.dart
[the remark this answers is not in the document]
tonidero: Hmm I think it's currently a named parameter but not optional right? We would have to pass a default value if so. Also, if we add a default value, we should update the API tests to make sure we can call this without that parameter.

lib/purchases_flutter.dart
tonidero: I think the earlier the better. If we can do it in the flutter level, and that way avoid having to add the iOS methods, that makes the most sense.

android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
```diff
@@ -170,6 +170,14 @@ public void onMethodCall(@NonNull MethodCall call, @NonNull Result result) {
                 isPersonalizedPrice = call.argument("isPersonalizedPrice"); // TODO: Rename
                 purchasePackage(packageIdentifier, offeringIdentifier, oldSKU, prorationMode, isPersonalizedPrice, result);
                 break;
+            case "purchaseSubscriptionOption":
+                productIdentifier = call.argument("productIdentifier");
```
tonidero, line 174: Hmm this drove me crazy... It seems we are sharing the variable declaration between cases. That's not great. Still no need to change it in this PR (it was like this before)

lib/purchases_flutter.dart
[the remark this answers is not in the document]
tonidero: Nvm you are right, if it's nullable like this, I *think* it would default to null. We just need to add those cases to the api tests @joshdholtz

lib/purchases_flutter.dart
```diff
@@ -271,13 +271,15 @@ class Purchases {
     String productIdentifier, {
     UpgradeInfo? upgradeInfo,
     PurchaseType type = PurchaseType.subs,
+    bool? isPersonalizedPrice,
```
tonidero, line 274: I think we need to add this parameter to the docs of these purchase methods. Also mention that this is Android only.

lib/purchases_flutter.dart
```diff
@@ -295,13 +297,48 @@ class Purchases {
...
+
+  /// Google Play only.
+  ///
+  /// Makes a purchase. Returns a [CustomerInfo] object. Throws a
+  /// [PlatformException] if the purchase is unsuccessful.
+  /// Check if [PurchasesErrorHelper.getErrorCode] is
+  /// [PurchasesErrorCode.purchaseCancelledError] to check if the user cancelled
+  /// the purchase.
+  ///
+  /// [packageToPurchase] The Package you wish to purchase
```
tonidero, line 321: We need to update these docs

pubspec.lock
```diff
@@ -5,18 +5,18 @@ packages:
     dependency: transitive
     description:
       name: _fe_analyzer_shared
-      sha256: "0c80aeab9bc807ab10022cd3b2f4cf2ecdf231949dc1ddd9442406a003f19201"
+      sha256: "3444216bfd127af50bbe4862d8843ed44db946dd933554f0d7285e89f10e28ac"
       url: "https://pub.dev"
     source: hosted
-    version: "52.0.0"
+    version: "50.0.0"
```
tonidero, line 11: Hmm interesting that most of these were downgraded... Do you need to update something?

api_tester/lib/api_tests/purchases_flutter_api_test.dart
```diff
@@ -76,15 +76,38 @@ class _PurchasesFlutterApiTest {
...
+    CustomerInfo customerInfo = await Purchases.purchaseSubscriptionOption(
+        subscriptionOption,
+        upgradeInfo: upgradeInfo);
+    customerInfo = await Purchases.purchaseSubscriptionOption(
+        subscriptionOption,
+        upgradeInfo: upgradeInfo,
+        isPersonalizedPrice: true);
+    customerInfo = await Purchases.purchaseSubscriptionOption(
+        subscriptionOption,
+        isPersonalizedPrice: true);
```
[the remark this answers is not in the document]
joshdholtz: Ooops, did I not?! I definitely will 😇 I tried to get all the cases

tonidero (commented):
I'm mostly concerned about the downgrade in versions... Other than that looks good

lib/purchases_flutter.dart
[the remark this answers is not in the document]
vegaro: named parameters are optional by default, I think in this case it is going to be null if not passed

android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
vegaro: are these TODO: Rename still applicable?

lib/models/price_wrapper.dart
beylmk: thank you, TIL. was trying to see how 8601 could be typoed to 4217 and it didn't seem to make sense haha

lib/models/period_wrapper.dart
```diff
@@ -0,0 +1,32 @@
...
+
+part 'period_wrapper.freezed.dart';
+part 'period_wrapper.g.dart';
+
+@freezed
+
+/// Contains all the details associated with a Period
+class Period with _$Period {
+  const factory Period(
+    /// The number of period units: DAY, WEEK, MONTH, YEAR, UNKNOWN
```
vegaro, line 11: nitpick but the values are actually lowercased in Dart

lib/models/subscription_option_wrapper.dart
```diff
@@ -0,0 +1,59 @@
...
+    /// True if this SubscriptionOption represents a Google subscription base plan (rather than an offer).
+    /// Not applicable for Amazon subscriptions.
+    @JsonKey(name: 'isBasePlan') bool isBasePlan,
+
+    /// The subscription period of fullPricePhase (after free and intro trials).
+    @JsonKey(name: 'billingPeriod', nullable: true) Period? billingPeriod,
+
+    /// The full price PricingPhase of the subscription.
+    /// Looks for the last price phase of the SubscriptionOption.
+    @JsonKey(name: 'fullPricePhase', nullable: true)
```
beylmk, line 43: just out of curiosity -- why have this instead of price like in native?

lib/models/pricing_phase_wrapper.dart
```diff
@@ -0,0 +1,42 @@
...
+part 'pricing_phase_wrapper.freezed.dart';
+part 'pricing_phase_wrapper.g.dart';
+
+@freezed
+
+/// Contains all the details associated with a PricingPhase
+class PricingPhase with _$PricingPhase {
+  const factory PricingPhase(
+    /// Billing period for which the PricingPhase applies
+    @JsonKey(name: 'billingPeriod') Period billingPeriod,
```
vegaro, line 15: Can this be null? I see phc is doing         `"billingPeriod" to billingPeriod?.mapPeriod()`

lib/models/subscription_option_wrapper.dart
```diff
@@ -0,0 +1,59 @@
...
+
+    /// The free trial PricingPhase of the subscription.
+    /// Looks for the first pricing phase of the SubscriptionOption where amountMicros is 0.
+    /// There can be a freeTrialPhase and an introductoryPhase in the same SubscriptionOption.
+    @JsonKey(name: 'freePhase', nullable: true) PricingPhase? freePhase,
+
+    /// The intro trial PricingPhase of the subscription.
+    /// Looks for the first pricing phase of the SubscriptionOption where amountMicros is greater than 0.
+    /// There can be a freeTrialPhase and an introductoryPhase in the same SubscriptionOption.
+    @JsonKey(name: 'introPhase', nullable: true) PricingPhase? introPhase,
```
beylmk, line 54: these are cool, shoudl we try to get these in native?

also - i guess we are assuming google won't ever allow more than one freeTrial/intro phase?

beylmk (approved):
🐦

lib/models/period_wrapper.dart
joshdholtz: Ahhh, I had this comment before I made it an enum 🤦‍♂️  Will fix!

lib/models/subscription_option_wrapper.dart
joshdholtz: Native has these on `SubscriptionOption` already!

pubspec.lock
joshdholtz: @tonidero  I had an old version of flutter when I started this so I think it downgraded stuff because of that... I just reset this file from `main` and now its just minor upgrades to some things 😇

lib/models/pricing_phase_wrapper.dart
joshdholtz: Ahhhh, it probably can be! Will fix 💪

PR #653, opened 2023-04-04
Release/5.0.0-alpha.1

joshdholtz:
The first alpha of **RevenueCat Purchases Flutter v5** is here!! 😻

This latest release updates the Android SDK dependency from v5 to [v6](https://github.com/RevenueCat/purchases-android/releases/tag/6.0.0) to use BillingClient 5. This version of BillingClient brings an entire new subscription model which has resulted in large changes across the entire SDK.

### Migration Guides
- See [Android Native - 5.x to 6.x Migration](https://www.revenuecat.com/docs/android-native-5x-to-6x-migration) for a
  more thorough explanation of the new Google subscription model announced with BillingClient 5 and how to take
  advantage of it in V6. This guide includes tips on product setup with the new model.

#### Purchasing
In v4, a Google Play Android `Package` or `StoreProduct` represented a single purchaseable entity, and free trials or intro
offers would automatically be applied to the purchase if the user was eligible.

Now, in Flutter v5, an Google Play Android `Package` or `StoreProduct` represents a duration of a subscription and contains all the ways to
purchase that duration -- any offers and its base plan. Each of these purchase options are `SubscriptionOption`s.
When passing a `Package` to `purchasePackage()` or `StoreProduct` to `purchaseStoreProduct()`, the SDK will use the following logic to choose which
`SubscriptionOption` to purchase:
- Filters out offers with "rc-ignore-offer" tag
- Uses `SubscriptionOption` with the longest free trial or cheapest first phase
    - Only offers the user is eligible will be applied
- Falls back to base plan

For more control, purchase subscription options with the new `purchaseSubscriptionOption()` method.

#### Models

`StoreProduct` now has a few new properties use for Google Play Android:
- `defaultOption`
  - A subscription option that will automatically be applie when purchasing a `Package` or `StoreProduct`
- `subscriptionOptions`
  - A list of subscription options (could be null)

### Observer Mode

Observer mode is still supported in v5. Other than updating the SDK version, there are no changes required.


test/purchases_flutter_test.dart
joshdholtz, line 1: `flutter analyze lib` didn't like this 🤷‍♂️

revenuecat_examples/purchase_tester/android/app/build.gradle
```diff
@@ -34,7 +34,7 @@ android {
     defaultConfig {
         applicationId "com.revenuecat.purchases_sample"
         minSdkVersion 16
-        targetSdkVersion 31
+        targetSdkVersion 30
```
joshdholtz, line 37: Target SDK version of 31 was causing this issue...

```
android:exported needs to be explicitly specified for element <activity#androidx.test.core.app.InstrumentationActivityInvoker$BootstrapActivity>. Apps targeting Android 12 and higher are required to specify an explicit value for `android:exported` when the corresponding component has an intent filter defined. See https://developer.android.com/guide/topics/manifest/activity-element#exported for details.
/Users/joshholtz/Developer/purchases-flutter/revenuecat_examples/purchase_tester/build/app/intermediates/tmp/manifest/androidTest/debug/tempFile1ProcessTestManifest3235334484840798035.xml:34:9-40:20
```

This target version didn't seem _super_ critical since this is the example app but still very weird and probably needs to be address in a better way for then the target version does need to be 31 again? 🤔

tonidero: Hmm weird. I just tried this branch and setting this version back to 31 and was able to build the android app without issues... Were you running through Android studio or something else?

joshdholtz: I was trying through command line with `./gradlew` but this was also failing on CI for me 😬

joshdholtz: I still can't really figure out what is causing this and another fix 😮‍💨 

@tonidero @vegaro Is this okay to ship in alpha and beta?

tonidero: I was taking a look and was able to repro when building integration tests, not the app itself. And I was taking a look at the cause. Looks like we increased the gradle and gradle plugin versions since the last release. My bet would be that broke this. Considering this is only in the purchase tester app. I think it would be ok to do this for alpha/beta, but we should fix it before final release to confirm there are no issues. My guess it should be updating some dependencies

tonidero (approved):

PR #655, opened 2023-04-06
[BC5] Rename `UpgradeInfo` to `ProductChangeInfo` and hide unsupported `ProrationMode`s

joshdholtz:
## Motivation

Replace `UpgradeInfo` with a new `ProductChangeInfo` 
- Android terminology is "changing products"
- The `UpgradeInfo` class used the term "sku" when we need to use "product identifier"

## Description

- Deprecated `UpgradeInfo` and `ProrationMode` for new `GoogleProductChangeInfo` and `GoogleProrationMode`
- Deprecated `UpgradeInfo` and added `GoogleProductChangeInfo` in `purchaseProduct()` and `purchasePackage()`
- Added `GoogleProductChangeInfo` to `purchaseSubscriptionOption()`

🔨 Also enabled `deprecated_member_use_from_same_package` in the analyze options so that `flutter analyze lib` and `flutter pub publish --dry-run` pass with the `@Deprecated UpgradeInfo? upgradeInfo` parameter

💪  Also updated tests that test the method call and arguments that are being sent to that native layer 

### Logic behind `GoogleProductChangeInfo` and  `GoogleProrationMode`

This naming with the "Google" prefix is to make it a little more clear to developers that these fields are for Google only without need to read the documentation 😇 

`GoogleProrationMode` also only allows `immediateWithTimeProration` and `immediateWithoutProration` and will add the others when Android V6 supports them 💪

### DIFF
--- a/android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
+++ b/android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
@@ -156,27 +156,27 @@ public class PurchasesFlutterPlugin implements FlutterPlugin, MethodCallHandler,
                 break;
             case "purchaseProduct":
                 String productIdentifier = call.argument("productIdentifier");
-                String oldProductIdentifer = call.argument("oldProductIdentifer");
-                Integer prorationMode = call.argument("prorationMode");
-                Boolean isPersonalizedPrice = call.argument("isPersonalizedPrice");
+                String googleOldProductIdentifer = call.argument("googleOldProductIdentifer");
+                Integer googleProrationMode = call.argument("googleProrationMode");
+                Boolean googleIsPersonalizedPrice = call.argument("googleIsPersonalizedPrice");
                 type = call.argument("type");
-                purchaseProduct(productIdentifier, type, oldProductIdentifer, prorationMode, isPersonalizedPrice, result);
+                purchaseProduct(productIdentifier, type, googleOldProductIdentifer, googleProrationMode, googleIsPersonalizedPrice, result);
                 break;
             case "purchasePackage":
                 String packageIdentifier = call.argument("packageIdentifier");
                 String offeringIdentifier = call.argument("offeringIdentifier");
-                oldProductIdentifer = call.argument("oldProductIdentifer");
-                prorationMode = call.argument("prorationMode");
-                isPersonalizedPrice = call.argument("isPersonalizedPrice");
-                purchasePackage(packageIdentifier, offeringIdentifier, oldProductIdentifer, prorationMode, isPersonalizedPrice, result);
+                googleOldProductIdentifer = call.argument("googleOldProductIdentifer");
+                googleProrationMode = call.argument("googleProrationMode");
+                googleIsPersonalizedPrice = call.argument("googleIsPersonalizedPrice");
+                purchasePackage(packageIdentifier, offeringIdentifier, googleOldProductIdentifer, googleProrationMode, googleIsPersonalizedPrice, result);
                 break;
             case "purchaseSubscriptionOption":
                 productIdentifier = call.argument("productIdentifier");
                 String optionIdentifier = call.argument("optionIdentifier");
-                oldProductIdentifer = call.argument("oldProductIdentifer");
-                prorationMode = call.argument("prorationMode");
-                isPersonalizedPrice = call.argument("isPersonalizedPrice");
-                purchaseSubscriptionOption(productIdentifier, optionIdentifier, oldProductIdentifer, prorationMode, isPersonalizedPrice, result);
+                googleOldProductIdentifer = call.argument("googleOldProductIdentifer");
+                googleProrationMode = call.argument("googleProrationMode");
+                googleIsPersonalizedPrice = call.argument("googleIsPersonalizedPrice");
+                purchaseSubscriptionOption(productIdentifier, optionIdentifier, googleOldProductIdentifer, googleProrationMode, googleIsPersonalizedPrice, result);
                 break;
             case "getAppUserID":
                 getAppUserID(result);
--- a/android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
+++ b/android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
@@ -156,7 +156,7 @@ public class PurchasesFlutterPlugin implements FlutterPlugin, MethodCallHandler,
                 break;
             case "purchaseProduct":
                 String productIdentifier = call.argument("productIdentifier");
-                String googleOldProductIdentifer = call.argument("googleOldProductIdentifer");
+                String googleOldProductIdentifer = call.argument("googleOldProductIdentifier");
                 Integer googleProrationMode = call.argument("googleProrationMode");
                 Boolean googleIsPersonalizedPrice = call.argument("googleIsPersonalizedPrice");
                 type = call.argument("type");
@@ -165,7 +165,7 @@ public class PurchasesFlutterPlugin implements FlutterPlugin, MethodCallHandler,
             case "purchasePackage":
                 String packageIdentifier = call.argument("packageIdentifier");
                 String offeringIdentifier = call.argument("offeringIdentifier");
-                googleOldProductIdentifer = call.argument("googleOldProductIdentifer");
+                googleOldProductIdentifer = call.argument("googleOldProductIdentifier");
                 googleProrationMode = call.argument("googleProrationMode");
                 googleIsPersonalizedPrice = call.argument("googleIsPersonalizedPrice");
                 purchasePackage(packageIdentifier, offeringIdentifier, googleOldProductIdentifer, googleProrationMode, googleIsPersonalizedPrice, result);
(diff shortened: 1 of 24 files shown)


android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
```diff
@@ -156,27 +156,27 @@ public void onMethodCall(@NonNull MethodCall call, @NonNull Result result) {
                 break;
             case "purchaseProduct":
                 String productIdentifier = call.argument("productIdentifier");
-                String oldSKU = call.argument("oldSKU"); // TODO: Rename
-                Integer prorationMode = call.argument("prorationMode"); // TODO: Rename
-                Boolean isPersonalizedPrice = call.argument("isPersonalizedPrice"); // TODO: Rename
+                String oldProductIdentifer = call.argument("oldProductIdentifer");
+                Integer prorationMode = call.argument("prorationMode");
```
beylmk, line 160: i think we called this googleProrationMode in native?

beylmk, line 160: i think we called this googleProrationMode in native?

lib/purchases_flutter.dart
```diff
@@ -946,8 +961,20 @@ class Purchases {
...
+class ProductChangeInfo {
+  /// The oldProductIdentifer to change from.
+  String oldProductIdentifer;
+
+  /// The [ProrationMode] to use when changing from the given oldProductIdentifer.
+  ProrationMode? prorationMode;
+
+  /// Constructs an ProductChangeInfo
+  ProductChangeInfo(this.oldProductIdentifer, {this.prorationMode});
+}
```
vegaro, line 973: I know UpgradeInfo is in this file, but now that we are making this from scratch I think it's a good opportunity to create a `product_change_info.dart` to hold this

lib/purchases_flutter.dart
```diff
@@ -272,6 +272,10 @@ class Purchases {
   /// [upgradeInfo] Android only. Optional UpgradeInfo you wish to upgrade from
   /// containing the oldSKU and the optional prorationMode.
   ///
+  /// [productChangeInfo] Android only. Optional ProductChangeInfo you wish to
```
beylmk, line 275: Android and Google Play Only**

lib/purchases_flutter.dart
```diff
@@ -956,35 +983,40 @@ class UpgradeInfo {
...
   /// be charged at the same time.
-  deferred,
+  // deferred,
 
+  // LATER: Reenable when supported by Android V6
   /// Replacement takes effect immediately, and the user is charged full price
   /// of new plan and is given a full billing cycle of subscription,
   /// plus remaining prorated time from the old plan.
-  immediateAndChargeFullPrice
+  // immediateAndChargeFullPrice
```
vegaro, line 1019: What's going to happen with these commented out ones?

lib/purchases_flutter.dart
```diff
@@ -284,14 +288,17 @@ class Purchases {
   /// for more info.
   static Future<CustomerInfo> purchaseProduct(
     String productIdentifier, {
-    UpgradeInfo? upgradeInfo,
+    @Deprecated('Use ProductChangeInfo') UpgradeInfo? upgradeInfo,
```
beylmk, line 291: ohh interesting, you can deprecate a single parameter on a method?

lib/purchases_flutter.dart
```diff
@@ -342,8 +356,9 @@ class Purchases {
   ///
   /// [packageToPurchase] The Package you wish to purchase
   ///
-  /// [upgradeInfo] Android only. Optional UpgradeInfo you wish to upgrade from
-  /// containing the oldSKU and the optional prorationMode.
+  /// [productChangeInfo] Android only. Optional ProductChangeInfo you wish to
+  /// change from containing the oldProductIdentifer and the
+  /// optional prorationMode.
```
beylmk, line 361: this will default to IMMEDIATE_WITHOUT_PRORATION at the native level, we should call that out.

lib/purchases_flutter.dart
joshdholtz: These will get commented back in when we support these modes in V6 (which should be soon ™) 😊

lib/purchases_flutter.dart
joshdholtz: Aye aye, can do that 💪

api_tester/lib/api_tests/purchases_flutter_api_test.dart
```diff
@@ -72,42 +72,49 @@ class _PurchasesFlutterApiTest {
   void _checkPurchaseProduct() async {
     String productIdentifier = "fakeProductId";
     UpgradeInfo? upgradeInfo;
+    GoogleProductChangeInfo? googleProductChangeInfo;
     PurchaseType purchaseType = PurchaseType.subs;
     CustomerInfo customerInfo = await Purchases.purchaseProduct(
         productIdentifier,
         upgradeInfo: upgradeInfo,
-        isPersonalizedPrice: true);
+        googleIsPersonalizedPrice: true);
```
beylmk, line 80: hmm i wonder if we can remove the google prefix now that these are within the GoogleProductChangeInfo class? to make it a little less wordy

joshdholtz: @beylmk `googleIsPersonalizedPrice` isn't in the `GoogleProductChangeInfo` but... should it be? 🤔 It can be used without product change though, right?

api_tester/lib/api_tests/purchases_flutter_api_test.dart
```diff
@@ -72,42 +72,49 @@ class _PurchasesFlutterApiTest {
...
+    customerInfo = await Purchases.purchasePackage(package,
+        upgradeInfo: upgradeInfo, googleIsPersonalizedPrice: true);
+    customerInfo = await Purchases.purchasePackage(package,
+        googleIsPersonalizedPrice: true);
   }
 
-  void _checkPurchaseSubscriptionOption(
-      SubscriptionOption subscriptionOption, UpgradeInfo? upgradeInfo) async {
+  void _checkPurchaseSubscriptionOption(SubscriptionOption subscriptionOption,
```
beylmk, line 106: this rename sooooort of makes me wonder if we should reanme to `GoogleSubscriptionOption` too 🙈

joshdholtz: I really want to 😬 Probably do that during the great module migration whenever that happens 😇

lib/purchases_flutter.dart
```diff
@@ -269,32 +273,39 @@ class Purchases {
...
+        googleProductChangeInfo?.prorationMode ?? upgradeInfo?.prorationMode;
     final customerInfo = await _invokeReturningCustomerInfo('purchaseProduct', {
       'productIdentifier': productIdentifier,
-      'oldSKU': upgradeInfo?.oldSKU,
-      'prorationMode': prorationMode?.index,
       'type': describeEnum(type),
-      'isPersonalizedPrice': isPersonalizedPrice,
+      'googleOldProductIdentifier':
+          googleProductChangeInfo?.oldProductIdentifier ?? upgradeInfo?.oldSKU,
+      'googleProrationMode': prorationMode?.index,
```
beylmk, line 307: should we still pass through the deprecated one somewhere?

test/purchases_flutter_test.dart
```diff
@@ -639,11 +639,10 @@ void main() {
         null,
         null,
       );
-      final purchasePackageResult =
-          await Purchases.purchaseSubscriptionOption(mockSubscriptionOption);
+      await Purchases.purchaseSubscriptionOption(mockSubscriptionOption);
 
       fail('an exception should have been thrown');
```
beylmk, line 644: should we be updating some unit tests to use the new ProductChangeInfo?

joshdholtz: Good call 😅 Will add!

lib/purchases_flutter.dart
joshdholtz: jweoifjapwioefjaiowpef... this is based on index which is going to be bad once we add new ones to this 🤦‍♂️ 

I need to fix this and also add unit tests around this

joshdholtz:
@beylmk @vegaro Did a pretty big refactor after your reviews! 

- Deprecated both `UpgradeInfo` and `ProrationMode`
- Made new `GoogleProductChangeInfo` and `GoogleProrationMode`

I updated the PR description with this and with the reasons why

lib/upgrade_info.dart
```diff
@@ -0,0 +1,43 @@
+/// This class holds the information used when upgrading from another sku.
+/// To be used with purchaseProduct and purchasePackage.
+@Deprecated('Use GoogleProductChangeInfo')
+class UpgradeInfo {
```
vegaro, line 4: This file should be added to `object_wrappers.dart`. Otherwise it's not exported

lib/google_product_change.dart
```diff
@@ -0,0 +1,35 @@
+class GoogleProductChangeInfo {
```
vegaro, line 1: This file should also be added to `object_wrappers.dart`

vegaro (changes requested):

lib/google_product_change.dart
joshdholtz: I exported these in `purchases_flutter.dart` but I'll move it to `object_wrappers.dart` 😁 

```dart
export 'google_product_change.dart';
export 'object_wrappers.dart';
export 'upgrade_info.dart';
```

vegaro: oh I see, yeah thanks for moving those over there

vegaro (approved):

PR #656, opened 2023-04-07
[BC5] Apply `presentedOfferingID`, `offerPaymentMode`,add new `purchaseStoreProduct()`, deprecated `purchaseProduct()`

joshdholtz:
⚠️ Merges into #655 

## Motivation

- Support `presentedOfferingID` on `StoreProduct` and `SubscriptionOption` and on purchasing
- Support `OfferPaymentMode?`on `PricingPhase`

## Description

- Support `presentedOfferingID` on `StoreProduct` and `SubscriptionOption` and  on purchasing with:
  - `purchaseProduct()`
  - `purchaseSubscriptionOption()`
  - `purchaseDiscountedProduct()`
- Added `offerPaymentMode: OfferPaymentMode?` onto `PricingPhase`
- Deprecated `PurchaseType` and replaced with `ProductType`
  - Not necessary but `ProductType` is the naming we use in other places
- New `purchaseStoreProduct(StoreProduct)` method
  - Make it even easier and less error prone for developers to make purchases with by just using the `StoreProdouct`
  - The `StoreProduct` already contains the identifier, product type, and presented offer identifier so developers should have to split that up

```dart
final customerInfo = await Purchases.purchaseStoreProduct(storeProduct);
```


lib/models/store_product_wrapper.dart
```diff
@@ -38,6 +38,9 @@ class StoreProduct with _$StoreProduct {
     @JsonKey(name: 'discounts', nullable: true)
         List<StoreProductDiscount>? discounts,
 
+    /// Product type. Null for iOS.
+    @JsonKey(name: 'productCategory', nullable: true) PurchaseType? productType,
```
beylmk, line 42: is now a good time to unify the naming of these? or do you think it should stay..

lib/models/store_product_wrapper.dart
```diff
@@ -46,6 +49,10 @@ class StoreProduct with _$StoreProduct {
     @JsonKey(name: 'subscriptionOptions', nullable: true)
         List<SubscriptionOption>? subscriptionOptions,
 
+    // Offering identifier the store product was presented from
```
beylmk, line 52: i added a line "Null if not using offerings or if fetched directly from store via getProducts" in native

lib/purchase_type.dart
```diff
@@ -0,0 +1,12 @@
+import 'package:json_annotation/json_annotation.dart';
+
+/// Supported SKU types.
+enum PurchaseType {
```
beylmk, line 4: i'm forgetting why this wouldn't just be `ProductType`? it's not really a type of purchase

beylmk (approved):
👍 just a couple questions/comments, nothing blocking!

lib/models/store_product_wrapper.dart
joshdholtz: I'm so torn! I think we could do the "productCategory" name change at a different time (because we do lock into specific PHC versions) so that could be a future change.

I don't _love_ `PurchaseType` for this and think it probably should be `ProductType` and... now is probably a good time to deprecate it and make a `ProductType` 🤷‍♂️  I'll make that change!

lib/purchase_type.dart
joshdholtz: This is what already excited but let's deprecate this and add a new one! 💪

beylmk: thank you!!

lib/purchases_flutter.dart
```diff
@@ -240,16 +240,30 @@ class Purchases {
...
   /// [productIdentifiers] Array of product identifiers
   ///
+  /// [productType] If the products are Android INAPPs, this needs to be
+  /// ProductType.INAPP otherwise the products won't be found.
+  /// ProductType.SUBS by default. This parameter only has effect in Android.
+  ///
   /// [type] If the products are Android INAPPs, this needs to be
   /// PurchaseType.INAPP otherwise the products won't be found.
-  /// PurchaseType.Subs by default. This parameter only has effect in Android.
+  /// PurchaseType.SUBS by default. This parameter only has effect in Android.
```
vegaro, line 249: It should be `subs` lowercased, right?

lib/purchases_flutter.dart
```diff
@@ -289,23 +309,74 @@ class Purchases {
...
     return customerInfo;
   }
 
+  /// Makes a purchase. Returns a [CustomerInfo] object. Throws a
+  /// [PlatformException] if the purchase is unsuccessful.
+  /// Check if [PurchasesErrorHelper.getErrorCode] is
+  /// [PurchasesErrorCode.purchaseCancelledError] to check if the user cancelled
+  /// the purchase.
+  ///
+  /// [storeProduct] The [StoreProduct]you want to purchase.
```
vegaro, line 347:
```suggestion
  /// [storeProduct] The [StoreProduct] you want to purchase.
```

lib/purchases_flutter.dart
```diff
@@ -260,6 +274,9 @@ class Purchases {
         .toList();
   }
 
+  /// It is now recommended to use [Purchases.purchaseStoreProduct]
+  /// to make a purchase with a [StoreProduct] if you can.
```
vegaro, line 278: Is there a reason why we haven't just deprecated this function? It requires a `type` which is a source of confusion for a lot of devs that want to offer consumables. A lot of them forget to pass it and the purchase doesn't work 😢

vegaro (approved):

lib/purchases_flutter.dart
joshdholtz: @vegaro No reason that I know of 😬  But I would be happy to deprecate this function, TBH 😇  It just looks super error prone to me

PR #666, opened 2023-04-14
[BC5] Update to PHC beta.6

joshdholtz:
## Motivation

Update to Purchases Hybrid Common Beta 6 which undid breaking change `productType`

## Description

- Rename `ProductType` to `ProductCategory` (unbreaking change)
- Undid deprecated parameters and removed new types in `purchaseProduct()`
  - The whole method is deprecated so need to add new parameter types to it since developer should use `purchaseStoreProduct()` anyway
  - Removed some tests that were no longer with this undo
- Removed duplicate `Unit` and replaced with already existing `PeriodUnit`
  -  Also moved this into its own file

### DIFF
--- a/lib/models/store_product_wrapper.dart
+++ b/lib/models/store_product_wrapper.dart
@@ -38,7 +38,7 @@ class StoreProduct with _$StoreProduct {
     @JsonKey(name: 'discounts', nullable: true)
         List<StoreProductDiscount>? discounts,
 
-    /// Product type. Null for iOS.
+    /// Product category.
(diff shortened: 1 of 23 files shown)


lib/models/store_product_wrapper.dart
```diff
@@ -39,7 +39,8 @@ class StoreProduct with _$StoreProduct {
         List<StoreProductDiscount>? discounts,
 
     /// Product type. Null for iOS.
```
vegaro, line 41: This is not true anymore

vegaro (approved):
After fixing api tests

PR #679, opened 2023-04-27
Fix MacOS integration tests on latest Flutter

vegaro:
According to https://github.com/flutter/flutter/issues/118469?notification_referrer_id=NT_kwDOAAoj4LE1MzI0NjE5MjQ2OjY2NDU0NA&notifications_query=repo%3Aflutter%2Fflutter#issuecomment-1525702770 this has been fixed in latest GitHub Actions macos-13 version https://github.com/actions/runner-images/blob/main/images/macos/macos-13-Readme.md which uses MacOS Ventura 13.3.1. Looks like latest version of CircleCI's Intel image is still on 13.2 (xcode 14.2.0) according to https://circleci.com/docs/using-macos/#supported-xcode-versions-intel and I tried to use that version and it's still not working.

I updated to use an M1 machine and 14.2.0 and it looks like it's working fine though


.circleci/config.yml
```diff
@@ -250,14 +250,13 @@ jobs:
 
   macos-integration-test:
     description: "Run integration tests for Flutter"
-    resource_class: macos.x86.medium.gen2
+    resource_class: macos.m1.large.gen1
```
tonidero, line 253: Hmm I believe these machines are much more expensive though... What are we using in the `purchases-ios` repo?

.circleci/config.yml
```diff
@@ -355,7 +354,7 @@ workflows:
         equal: [ scheduled_pipeline, << pipeline.trigger_source >> ]
     jobs:
       - ios-integration-test: *release-branches
-      - macos-integration-test: *release-branches
+      - macos-integration-test
```
tonidero, line 357: I guess this should still run on release branches only?

.circleci/config.yml
```diff
@@ -250,14 +250,13 @@ jobs:
...
-    resource_class: macos.x86.medium.gen2
+    resource_class: macos.m1.large.gen1
     macos:
-      xcode: 13.4.1
+      xcode: 14.2.0
     shell: /bin/bash --login -o pipefail
     steps:
       - checkout
-      - setup-flutter:
-          version: '3.3.10'
```
tonidero, line 260: Hmm if I'm not wrong, by removing this, we will be pointing to the latest `main` in flutter? But seems to be what we do for the other jobs, so this might be ok.

tonidero (commented):
left some questions

.circleci/config.yml
vegaro: oh wow yes, I just checked and they are 400 credits vs 50... Let's wait until they release a more recent version of Ventura with their x86 machine then. I think that will fix it like it did in GitHub Actions.

Thanks for pointing this out!

vegaro:
Tests last 2m 27s (MacOS Medium M1 367.5 credits) vs 3m 9s (MacOS Medium Intel Gen2 236.25 credits) with the intel machine

vegaro:
CircleCI replied 

> After speaking with our engineering team, there is no GPU access with the Intel Machines and you will need to utilize the M1 medium or large resource class to have access.

vegaro:
Looks like there's now a cocoapods issue https://github.com/flutter/flutter/issues/131527 so this can't be merged yet

PR #683, opened 2023-05-09
Support Gradle 8+

davidmartos96:
Hello, currently the plugin doesn't build when Gradle 8+ is used. That is because the property `namespace` is mandatory.

* I updated the Android example project.
* I included the namespace property with backwards compatibility

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -10,7 +10,7 @@ buildscript {
     }
 
     dependencies {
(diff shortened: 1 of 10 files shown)


aboedo:
Hey @davidmartos96 👋 Thanks so much for bringing this to our attention and opening up a PR! 
We'll review in a bit and ship it!

PR #684, opened 2023-05-10
moved method and param names in constants.dart

Arunshaik2001:
# what does this pr do?
moved method names and parameters names to constants dart file. it will be easy to reuse same method and param names and make less human errors while typing string names.


vegaro:
Thanks for opening this @Arunshaik2001! And sorry we haven't replied earlier.

While the argument for never having any magic strings can certainly be made, there are a few instances where in practice we've found them to be the lesser evil, given the inconvenience from the perspective of the reader (maintainer) to have indirection introduced whenever we want to understand and change what a method does.

We know this goes against the "no magic strings" policy and has downsides too, but in our experience the convenience and readability tradeoff is worth it, specially if the magic strings are only being used in one place.

So we think that since we are not reusing these constants we might keep these magic strings as they are right now.

PR #686, opened 2023-05-11
Migrated java PurchasesFlutterPlugin class to kotlin class.

Arunshaik2001:
# What does this pr do?
1. Migrated Java plugin file to Kotlin file to use kotlin features.
2. Moved method names and argument names in constant file


android/src/main/kotlin/com/revenuecat/purchases_flutter/Constants.kt
```diff
@@ -0,0 +1,112 @@
+package com.revenuecat.purchases_flutter
+
+internal object Constants {
+
+    object MethodNameConstants{
```
tonidero, line 5: We believe extracting all these constants doesn't provide much value and can even decrease legibility. The exception to this would be when they are repeated. Could you move the constants back to the plugin class? It would be fine to leave any that are repeated.

android/src/main/kotlin/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.kt
```diff
@@ -0,0 +1,652 @@
...
+
+    private fun onAttachedToEngine(messenger: BinaryMessenger, applicationContext: Context) {
+        channel = MethodChannel(messenger, "purchases_flutter")
+        this.applicationContext = applicationContext
+        channel!!.setMethodCallHandler(this)
+    }
+
+    override fun onDetachedFromEngine(binding: FlutterPluginBinding) {
+        if (channel != null) {
+            channel!!.setMethodCallHandler(null)
```
tonidero, line 52: Could this be `channel?.setMethodCallHandler(null)`

android/src/main/kotlin/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.kt
```diff
@@ -0,0 +1,652 @@
...
+    override fun onDetachedFromActivity() {
+        activity = null
+    }
+
+    override fun onDetachedFromActivityForConfigChanges() {
+        onDetachedFromActivity()
+    }
+
+    private fun getActivity(): Activity? {
+        return if (registrar != null) registrar!!.activity() else activity
```
tonidero, line 75: This could be `registrar?.activity() ?: activity` I think?

android/src/main/kotlin/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.kt
```diff
@@ -0,0 +1,652 @@
...
+        useAmazon: Boolean?, result: MethodChannel.Result
+    ) {
+        if (applicationContext != null) {
+            val platformInfo = PlatformInfo(PLATFORM_NAME, PLUGIN_VERSION)
+            var store = Store.PLAY_STORE
+            if (useAmazon != null && useAmazon) {
+                store = Store.AMAZON
+            }
+            configure(
+                applicationContext!!, apiKey!!, appUserID, observerMode,
```
tonidero, line 278: Instead of crashing here, it would be better to try to handle this gracefully by calling `result.error` (I understand it would have crashes with the Java version as well, but we can improve that now)

android/src/main/kotlin/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.kt
```diff
@@ -0,0 +1,652 @@
...
+    private fun getOfferings(result: MethodChannel.Result) {
+        getOfferings(getOnResult(result))
+    }
+
+    private fun getProductInfo(
+        productIDs: ArrayList<String>,
+        type: String?,
+        result: MethodChannel.Result
+    ) {
+        getProductInfo(productIDs, type!!, object : OnResultList {
```
tonidero, line 339: Same here and other places we use `!!`. Ideally we handle it gracefully

android/src/main/kotlin/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.kt
```diff
@@ -0,0 +1,652 @@
...
+    }
+
+    private fun reject(errorContainer: ErrorContainer?, result: MethodChannel.Result) {
+        result.error(errorContainer!!.code.toString(), errorContainer.message, errorContainer.info)
+    }
+
+    private fun invokeChannelMethodOnUiThread(method: String, argumentsMap: Any) {
+        runOnUiThread {
+            if (channel != null) {
+                channel!!.invokeMethod(method, argumentsMap)
```
tonidero, line 622: This could be `channel?.invokeMethod(method, argumentsMap)`

android/src/main/kotlin/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.kt
```diff
@@ -0,0 +1,652 @@
...
+            }
+        }
+    }
+
+    companion object {
+        private const val CUSTOMER_INFO_UPDATED = "Purchases-CustomerInfoUpdated"
+        protected const val LOG_HANDLER_EVENT = "Purchases-LogHandlerEvent"
+        private const val PLATFORM_NAME = "flutter"
+        private const val PLUGIN_VERSION = "4.13.0-SNAPSHOT"
```
tonidero, line 631: We will need to update to use the latest version of the plugin.

tonidero (commented):
Hi @Arunshaik2001, really sorry for taking so long to answer. 

We definitely would like to move this code to Kotlin! So this is a great contribution, thanks! I left a few comments here, let me know if you have any questions or anything I can help with!

PR #699, opened 2023-05-24
Fix issues when running Flutter 3.10.2

vegaro:
- Fixes CI images cherry picking [8b37b175c4ba476f69aab74ac4b2dceebbd93720](https://github.com/RevenueCat/purchases-flutter/commit/8b37b175c4ba476f69aab74ac4b2dceebbd93720)
- Also fixing the examples since the won't build with latest Flutter (https://linear.app/revenuecat/issue/SDK-3153/[flutter]-make-flutter-examples-null-safe)
- Fixes new warnings https://github.com/RevenueCat/purchases-flutter/pull/700
- Removes pubspec.lock


MarkVillacampa (approved):

vegaro:
Lint job is blocked by #700

NachoSoto (approved):
Looks good!

pubspec.lock
```diff
@@ -587,5 +579,5 @@ packages:
     source: hosted
     version: "3.1.1"
 sdks:
-  dart: ">=2.19.0 <3.0.0"
+  dart: ">=3.0.0-0 <4.0.0"
```
tonidero, line 582: Hmm should this be considered a breaking change?

revenuecat_examples/MagicWeather/lib/src/views/user.dart
```diff
@@ -36,7 +36,7 @@ class _UserScreenState extends State<UserScreen> {
 
     try {
       if (task == "login") {
-        await Purchases.logIn(newAppUserID);
+        await Purchases.logIn(newAppUserID!);
```
tonidero, line 39: Hmm wonder if we should separate the `login` operation to a different method to avoid the optional and avoid this...

revenuecat_examples/MagicWeather/lib/src/views/weather.dart
```diff
@@ -70,7 +72,7 @@ class _WeatherScreenState extends State<WeatherScreen> {
             return StatefulBuilder(
                 builder: (BuildContext context, StateSetter setModalState) {
               return Paywall(
-                offering: offerings.current,
+                offering: offerings!.current!,
```
tonidero, line 75: Hmm is Dart not smart enough to understand that this can't be null in the `else`? :(

tonidero (commented):
Just have that question about the lower bound of dart being bumped... If that's required, then this would be breaking. However, we haven't changed the `pubspec.yaml` so it might still be compatible? Not sure

pubspec.lock
MarkVillacampa: This is weird. Just checked on pub.dev that none of the updated transitive dependencies require dart 3.0, plus we have `sdk: ">=2.12.0 <3.0.0"` in `pubspec.yaml`. 🤔

Latest flutter stable does require 3.0, I wonder if by running `flutter pub` commands it updates the lockfile?

vegaro: I think this is due to building locally using the latest flutter. Does the latest flutter require dart 3? Also this is just a lock file so no breaking change here

revenuecat_examples/MagicWeather/lib/src/views/weather.dart
vegaro: I don't think so... Only Kotlin haha

revenuecat_examples/MagicWeather/lib/src/views/user.dart
vegaro: I split _manageUser into three functions so it's safer

pubspec.lock
MarkVillacampa: Turns out Dart recommends not committing lock files for libraries, so I removed it (merged into this branch)

https://github.com/RevenueCat/purchases-flutter/pull/701

MarkVillacampa (approved):

PR #701, opened 2023-05-26
Remove pubspec.lock

MarkVillacampa:
Guidelines recommend *not* to commit the lock file for library packages in order to always test the package against the latest compatible versions of its dependencies.

https://dart.dev/guides/libraries/private-files#pubspeclock


vegaro (approved):
let's also remove its references in the circleci config file

PR #703, opened 2023-05-30
Fix integration tests

MarkVillacampa:
We only used to run integration tests in the `deploy` workflow. I've added the ios/macos integration test runs plus *building* the Android integration tests to the PR workflow. This should increment CI times but will catch many more issues. 

We can potentially enable running firebase integration tests if needed, or adding a hold so they can be run conditionally.

Apart from that, this PR updates Kotlin and Android Gradle Plugin versions, as well as test dependencies.

One thing to note is I didn't upgrade to Kotlin 1.8 because I stumbled upon this duplicate class issue. Will revisit it with more time.

https://kotlinlang.org/docs/whatsnew18.html#usage-of-the-latest-kotlin-stdlib-version-in-transitive-dependencies

```
flutter build apk --release

Note: /Users/mark/src/revenuecat/purchases-flutter/revenuecat_examples/purchase_tester/android/app/src/main/java/com/revenuecat/purchases_sample/EmbeddingV1Activity.java uses or overrides a deprecated API.
Note: Recompile with -Xlint:deprecation for details.

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':app:checkReleaseDuplicateClasses'.
> A failure occurred while executing com.android.build.gradle.internal.tasks.CheckDuplicatesRunnable
   > Duplicate class kotlin.collections.jdk8.CollectionsJDK8Kt found in modules jetified-kotlin-stdlib-1.8.21 (org.jetbrains.kotlin:kotlin-stdlib:1.8.21) and jetified-kotlin-stdlib-jdk8-1.5.30 (org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.5.30)
     Duplicate class kotlin.internal.jdk7.JDK7PlatformImplementations found in modules jetified-kotlin-stdlib-1.8.21 (org.jetbrains.kotlin:kotlin-stdlib:1.8.21) and jetified-kotlin-stdlib-jdk7-1.6.20 (org.jetbrains.kotlin:kotlin-stdlib-jdk7:1.6.20)
     Duplicate class kotlin.internal.jdk7.JDK7PlatformImplementations$ReflectSdkVersion found in modules jetified-kotlin-stdlib-1.8.21 (org.jetbrains.kotlin:kotlin-stdlib:1.8.21) and jetified-kotlin-stdlib-jdk7-1.6.20 (org.jetbrains.kotlin:kotlin-stdlib-jdk7:1.6.20)
[...]

PR #706, opened 2023-06-01
Fix analyzer issues

MarkVillacampa:
After merging #703 the dry-run job detected issues:

https://app.circleci.com/pipelines/github/RevenueCat/purchases-flutter/1961/workflows/f9d8556e-775a-4860-a717-6eb9a2bb533e/jobs/8901

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -19,7 +19,6 @@ linter:
     avoid_void_async: true
     cascade_invocations: true
     directives_ordering: true
-    invalid_annotation_target: false
     omit_local_variable_types: true
     prefer_const_constructors: true
     prefer_const_constructors_in_immutables: true
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -7,8 +7,8 @@ issue_tracker: https://github.com/RevenueCat/purchases-flutter/issues
 documentation: https://docs.revenuecat.com/docs
 
 environment:
-  sdk: ">=2.12.0 <4.0.0"
-  flutter: "^1.10.0"
+  sdk: ">=2.15.0 <4.0.0"
+  flutter: ">=1.10.0"
 
 dependencies:
   flutter:
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -8,6 +8,7 @@ documentation: https://docs.revenuecat.com/docs
 
 environment:
   sdk: ">=2.12.0 <4.0.0"
+  flutter: "^1.10.0"
 
 dependencies:
   flutter:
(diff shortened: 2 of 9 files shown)


MarkVillacampa:
A few more issues popped up, fixing those asap 😄

analysis_options.yaml
```diff
@@ -19,7 +19,6 @@ linter:
     avoid_void_async: true
     cascade_invocations: true
     directives_ordering: true
-    invalid_annotation_target: false
```
MarkVillacampa, line 22: Got this error in VSCode, nothing in the console, so I'm not 100% if it's accurate.

`'invalid_annotation_target' is not a recognized lint ruledart(undefined_lint_warning)`

MarkVillacampa:
The two remaining issues are:

```
* /root/project/CHANGELOG.md doesn't mention current version (4.13.0-SNAPSHOT).
  Consider updating it with notes on this version prior to publication.
```

I think we don't add the current version changelog until we create a release so this is a non-issue.

```
Package validation found the following hint:
* The latest published version is 5.0.0-beta.2.
  Your version 4.13.0-SNAPSHOT is earlier than that.
```

Does this mean releasing a fix for an earlier version if we have published a beta for a newer version would fail?

The analysis `flutter pub publish --dry-run` does is great and runs fast, I wonder if we can make these two particular checks run only on deploy?

pubspec.yaml
```diff
@@ -7,7 +7,8 @@ issue_tracker: https://github.com/RevenueCat/purchases-flutter/issues
 documentation: https://docs.revenuecat.com/docs
 
 environment:
-  sdk: ">=2.12.0 <4.0.0"
+  sdk: ">=2.15.0 <4.0.0"
```
MarkVillacampa, line 10: `warning • This API is available since SDK 2.15.0, but constraints '>=2.12.0 <4.0.0' don't guarantee it • test/purchases_flutter_test.dart:810:32 • sdk_version_since`

pubspec.yaml
```diff
@@ -7,7 +7,8 @@ issue_tracker: https://github.com/RevenueCat/purchases-flutter/issues
...
+  flutter: ">=1.10.0"
```
MarkVillacampa, line 11:
```
* pubspec.yaml allows Flutter SDK version 1.9.x, which does not support the flutter.plugin.platforms key.
  Please consider increasing the Flutter SDK requirement to ^1.10.0 (environment.sdk.flutter)

  See https://flutter.dev/docs/development/packages-and-plugins/developing-packages#plugin
```

vegaro (approved):

PR #707, opened 2023-06-01
allow running Firebase integration tests in every PR if needed

MarkVillacampa:


.circleci/config.yml
```diff
@@ -381,11 +381,9 @@ workflows:
           type: approval
```
vegaro, line 381: we could maybe rename from `hold` to `hold-firebase-tests`, so it's more clear what the hold does

MarkVillacampa:
closing this for now as we dont want to have a old on every PR instead of a green checkmark ✅

PR #716, opened 2023-06-07
Adding metadata to offering

joshdholtz:
## Motivation

Add `metadata` to `Offering`

## Description

- Bumped Android, iOS, and macOS to hybrid common `5.0.0-rc.1`
- Added `metadata: Map<String, Object>` to `Offering`
- Updated API tester
- Added `getMetadataString(String key, String defaultValue)` to `Offering`
  - This is the same that native Android has

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -3,7 +3,7 @@ version '5.0.0-beta.2'
 
 buildscript {
     ext.kotlin_version = '1.7.21'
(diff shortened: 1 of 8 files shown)


tonidero (approved):
I don't want to block the release, but I think we can do better in metadata so users don't have to do the conversion themselves

NachoSoto (approved):
Awesome

PR #721, opened 2023-06-09
Add `isPrepaid` to `SubscriptionOffering`

joshdholtz:
## Description

- Adds `isPrepaid: bool` to `SubscriptionOffering`

### Demo

Screenshot 2023-06-09 at 1 36 34 PM

### DIFF
--- a/lib/models/subscription_option_wrapper.dart
+++ b/lib/models/subscription_option_wrapper.dart
@@ -28,20 +28,21 @@ class SubscriptionOption with _$SubscriptionOption {
     /// Pricing phases defining a user's payment plan for the product over time.
     @JsonKey(name: 'pricingPhases') List<PricingPhase> pricingPhases,
 
-    /// Tags defined on the base plan or offer. Empty for Amazon.
+    /// Tags defined on the base plan or offer.
     @JsonKey(name: 'tags') List<String> tags,
 
     /// True if this SubscriptionOption represents a Google subscription base plan (rather than an offer).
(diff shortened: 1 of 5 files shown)


lib/models/subscription_option_wrapper.dart
```diff
@@ -28,20 +28,21 @@ class SubscriptionOption with _$SubscriptionOption {
...
 
     /// True if this SubscriptionOption represents a Google subscription base plan (rather than an offer).
-    /// Not applicable for Amazon subscriptions.
     @JsonKey(name: 'isBasePlan') bool isBasePlan,
 
     /// The subscription period of fullPricePhase (after free and intro trials).
     @JsonKey(name: 'billingPeriod') Period? billingPeriod,
 
+    /// True if the subscription is pre-paid.
+    @JsonKey(name: 'isPrepaid') bool isPrepaid,
```
tonidero, line 41: I wonder if this property should live in the `StoreProduct` model... As in, in our abstraction, StoreProduct corresponds to a base plan, which is what will be linked to be `prepaid` or not... Thoughts?

tonidero (approved):
Left one comment, lmk what you think, other than that, looks good!

lib/models/subscription_option_wrapper.dart
joshdholtz: @tonidero This is probably a good point! I'll start the PR changing up from native Android to start that 😛

PR #722, opened 2023-06-12
Fix purchase tester reconfiguring the SDK

tonidero:
Purchase tester currently reconfigures the RC SDK on every InitialScreen instantiation. Due to the way we do navigation, we instantiate that screen after a purchase and possibly in some other situations. This PR does the following changes:
- Makes it so we don't recreate the `InitialScreen` without need in purchase tester.
- Moves the SDK initialization to the `main.dart` file so we ensure it's executed only once (This is done both in purchase tester and MagicWeather)

### DIFF
--- a/revenuecat_examples/purchase_tester/lib/store_config.dart
+++ b/revenuecat_examples/purchase_tester/lib/store_config.dart
@@ -1,11 +1,11 @@
-enum Store { appleStore, googlePlay, amazonAppstore }
+enum TesterStore { appleStore, googlePlay, amazonAppstore }
 
 class StoreConfig {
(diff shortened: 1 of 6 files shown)


revenuecat_examples/purchase_tester/lib/store_config.dart
```diff
@@ -1,4 +1,4 @@
-enum Store { appleStore, googlePlay, amazonAppstore }
```
tonidero, line 1: I removed this enum since I think we can use the library `Store` instead.

NachoSoto (approved):
Nice catch

PR #726, opened 2023-06-13
Fix flutter static analysis in pub.dev

tonidero:
Looks like we had a bunch of warnings in pub.dev due to freezed annotations. This PR hopes to fix those. Most of the warnings were in the form: 
```
warning • The annotation 'JsonKey' can only be used on fields or getters • lib/models/offering_wrapper.dart:45:6 • invalid_annotation_target
```
This is an issue recognized by the `freezed` library maintainer and their recommendation is to just disable the lint rule, which we already do locally... https://github.com/rrousselGit/freezed/issues/488. However, it seems that in pub.dev, they don't recognize our ignore rules and just run the entire set of rules.

In this PR, I removed the unnecessary `@JsonKey` annotations in our models, and left a few that were causing some changes in behavior and added a explicit ignore comment in those. 

Additionally, in this PR, I updated flutter_lints


analysis_options.yaml
```diff
@@ -1,14 +1,12 @@
...
 analyzer:
-  errors:
-    invalid_annotation_target: ignore
-    deprecated_member_use_from_same_package: ignore
   exclude:
     - lib/models/*.freezed.dart
     - lib/models/*.g.dart
     - lib/models/**/*.freezed.dart
     - lib/models/**/*.g.dart
+    - api_tester/**
```
tonidero, line 9: The api_tester already runs the linter separately as part of the api tests. It was conflicting with the new version of flutter_lints if I didn't do this.

lib/models/customer_info_wrapper.dart
```diff
@@ -12,44 +12,39 @@ part 'customer_info_wrapper.g.dart';
 class CustomerInfo with _$CustomerInfo {
   const factory CustomerInfo(
     /// Entitlements attached to this customer info
-    @JsonKey(name: 'entitlements') EntitlementInfos entitlements,
+    EntitlementInfos entitlements,
```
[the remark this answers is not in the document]
NachoSoto: Is it possible to do this?
```suggestion
    @JsonKey EntitlementInfos entitlements,
```
That was it's explicit that these are json keys at least.

tonidero:
I'm not sure if the places where I left the offending annotations will still be brought up by pub.dev. But this should reduce the number of errors substantially at least. Lmk what you think

analysis_options.yaml
```diff
@@ -1,9 +1,8 @@
-include: package:flutter_lints/flutter.yaml
+include:
+  - package:flutter_lints/flutter.yaml
+  - package:lints/recommended.yaml
 
 analyzer:
-  errors:
-    invalid_annotation_target: ignore
-    deprecated_member_use_from_same_package: ignore
```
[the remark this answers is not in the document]
vegaro: do you mind adding a comment here so we don't add more rules in the future?

lib/models/entitlement_info_wrapper.dart
```diff
@@ -43,39 +44,33 @@ enum OwnershipType {
...
-    @JsonKey(name: 'isSandbox')
-        bool isSandbox, {
+    bool isSandbox, {
 
     /// Use this property to determine whether a purchase was made by the current
     /// user or shared to them by a family member. This can be useful for
     /// onboarding users who have had an entitlement shared with them, but might
     /// not be entirely aware of the benefits they now have.
     @Default(OwnershipType.unknown)
+    // ignore: invalid_annotation_target
```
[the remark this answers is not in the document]
vegaro: Let's hope pub.dev respects it

lib/models/entitlement_info_wrapper.g.dart
```diff
@@ -49,7 +49,7 @@ Map<String, dynamic> _$$_EntitlementInfoToJson(_$_EntitlementInfo instance) =>
 const _$OwnershipTypeEnumMap = {
   OwnershipType.purchased: 'PURCHASED',
   OwnershipType.familyShared: 'FAMILY_SHARED',
-  OwnershipType.unknown: 'unknown',
+  OwnershipType.unknown: 'UNKNOWN',
```
vegaro, line 52: This has changed. Is it ok? Looks like the new value is more accurate anyways

lib/models/introductory_price.dart
```diff
@@ -11,27 +11,28 @@ part 'introductory_price.g.dart';
...
 
     /// Billing period of the introductory price, specified in
     /// ISO 8601 format.
-    @JsonKey(name: 'period') String period,
+    String period,
 
     /// Number of subscription billing periods for which the
     /// user will be given the introductory price, such as 3.
-    @JsonKey(name: 'cycles') int cycles,
+     int cycles,
```
vegaro, line 26:
```suggestion
    int cycles,
```

lib/models/introductory_price.freezed.dart
```diff
@@ -61,18 +57,13 @@ abstract class $IntroductoryPriceCopyWith<$Res> {
...
-          String priceString,
-      @JsonKey(name: 'period')
-          String period,
-      @JsonKey(name: 'cycles')
-          int cycles,
+      {double price,
+      String priceString,
+      String period,
+      int cycles,
       @JsonKey(name: 'periodUnit', unknownEnumValue: PeriodUnit.unknown)
```
vegaro, line 64: I think this will probably break the score too. But I doubt we can control adding the ignore here since this is autogenerated.

vegaro (approved):

lib/models/introductory_price.freezed.dart
tonidero: Yeah... Maybe the rules in pub.dev ignore freezed file? But not sure... I tried to remove these but it didn't work well. Wanted to get the "easier" issues fixed and we can see how that affects the pub.dev score

lib/models/entitlement_info_wrapper.g.dart
tonidero: Yeah I actually had to change this here: https://github.com/RevenueCat/purchases-flutter/pull/726/files#r1228355050

This is the actual value that gets sent from PHC so as you said, it's more correct. And I don't think this will cause any breaking changes at all (unless someone was deliberately trying to create an entitlement info with a `unknown` ownership type.

PR #698, opened 2023-05-24
Support Gradle 8+

vegaro:
Merging #683 into main

PR #749, opened 2023-06-28
Adds package back to AndroidManifest.xml

### LINKED ISSUE #748: Package name missing in AndroidManifest.xml

# ‼️ Required data ‼️ #

## Environment ##
Screenshot 2023-06-28 at 11 28 09 AM

     **v5.2.1**

     **Redmi 9A, Android 10 (API level 29)**
 
     **The issue prevents the app from running.**
- [ ] [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue

## Description: ##
I updated the package from version 4.12.0 to v 5.2.1, and faced the below issue when trying to run the app:
Screenshot 2023-06-28 at 11 25 44 AM

## Possible fix: ##

I tried adding the package name from v4.12.0 `package="com.revenuecat.purchases_flutter"` to the local cached package 5.2.1, and it seems to run the app. 
The package name was removed in commit 7b32814
I haven't been able to test the entire functionality after this, but I'm thinking it should work.

```xml
<manifest xmlns:android="http://schemas.android.com/apk/res/android"
  package="com.revenuecat.purchases_flutter">
</manifest>

```

Please let me know if I can provide any additional info, or create a PR.

vegaro:
Fixes #748 

We removed this property from the AndroidManifest.xml in #698 because we migrated to use the namespace gradle property. We added a conditional in the build.gradle to support AGP < 4.2 by checking if the namespace property exists, but if that happens, the package property is missing from the AndroidManifest.xml, which prevents compilation

I tried building this using AGP 8.0.2 and it works fine

### DIFF
--- a/android/src/main/AndroidManifest.xml
+++ b/android/src/main/AndroidManifest.xml
@@ -1,2 +1,3 @@
-<manifest xmlns:android="http://schemas.android.com/apk/res/android">
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+    package="com.revenuecat.purchases_flutter">
 </manifest>


MarkVillacampa (approved):

android/src/main/AndroidManifest.xml
```diff
@@ -1,2 +1,3 @@
-<manifest xmlns:android="http://schemas.android.com/apk/res/android">
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+    package="com.revenuecat.purchases_flutter">
```
tonidero, line 2: So this will be duplicated both here and in the namespace property right in the gradle file? I think it's ok though.

tonidero (approved):

android/src/main/AndroidManifest.xml
vegaro: Yes, I was worried AGP 8 would warn but it doesn't, so I think it's ok yes. I also noticed the official plugins have it duplicated too

https://github.com/flutter/packages/blob/main/packages/camera/camera_android/android/src/main/AndroidManifest.xml

https://github.com/flutter/packages/blob/main/packages/camera/camera_android/android/build.gradle#L31

PR #753, opened 2023-06-29
`Trusted Entitlements`: add support for setting `EntitlementVerificationMode` and getting verification result in `EntitlementInfos` and `EntitlementInfo`

NachoSoto:
See http://rev.cat/trusted-entitlements for more info on this feature.

Depends on https://github.com/RevenueCat/purchases-hybrid-common/pull/451


lib/models/entitlement_verification_mode.dart
```diff
@@ -0,0 +1,29 @@
...
+  ///
+  /// This can be useful if you want to handle verification failures to display an error/warning to the user
+  /// or to track this situation but still grant access.
+  informational,
+
+  // Disabled temporarily until ENFORCED is supported.
+  // enforced
+}
+
+extension EntitlementVerificationModeExtension on EntitlementVerificationMode {
```
tonidero, line 20: Note that this is something new. Basically, we're trying to pass an enum back to the native layer, which is not something we did before (afaik). Instead of using freezed to perform this mapping, I'm hardcoding it here, since it seemed easier.

vegaro (approved):

PR #772, opened 2023-07-25
[CI] Optionally run firebase tests only on non-main, non-release branches

MarkVillacampa:
That way we avoid keeping orange all commits in main.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,6 +19,13 @@ aliases:
         ignore: /.*/
       branches:
         only: main
+  except-main-and-release: &except-main-and-release
+    filters:
+      tags:
+        ignore: /.*/
+      branches:
+        ignore: /^release\/.*/
+        ignore: main
 
 orbs:
   android: circleci/android@0.2.1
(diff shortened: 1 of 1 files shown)


NachoSoto (approved):
Thanks!

.circleci/config.yml
```diff
@@ -381,11 +372,11 @@ workflows:
           type: approval
           requires:
             - android-integration-test-build
-          <<: *release-branches-and-main
+          <<: *release-branches
       - run-firebase-tests:
           requires:
             - hold
-          <<: *release-branches-and-main
+          <<: *release-branches
```
tonidero, line 379: Hmm I think we should probably just remove this. Note that we are already running the integration tests on release branches as part of the `deploy` workflow. IMO, I think this should be running in the `main` branch only. Thoughts?

.circleci/config.yml
```diff
@@ -381,11 +379,11 @@ workflows:
           type: approval
           requires:
             - android-integration-test-build
-          <<: *release-branches-and-main
+          <<: *except-main-and-release
       - run-firebase-tests:
           requires:
             - hold
-          <<: *release-branches-and-main
+          <<: *except-main-and-release
```
MarkVillacampa, line 386: With this change, firebase tests will run:
- Never in main branch
- Always on release branches
- At least once per PR (if we want to have the green check before merging)

tonidero (approved):
Nice!

PR #827, opened 2023-10-03
Remove hold/test job

vegaro:
I noticed the `run-firebase-tests` job is a requirement to merge a branch into `main`, but this test is behind a hold job. I think we should remove the hold job and just always run that `run-firebase-tests` on deploy, same way we do in purchases-android

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,14 +19,6 @@ aliases:
         ignore: /.*/
       branches:
         only: main
-  except-main-and-release: &except-main-and-release
-    filters:
-      tags:
-        ignore: /.*/
-      branches:
-        ignore:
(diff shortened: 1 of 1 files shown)


MarkVillacampa (approved):

.circleci/config.yml
```diff
@@ -376,15 +376,6 @@ workflows:
...
       - android-integration-test-build
-      - hold:
-          type: approval
-          requires:
-            - android-integration-test-build
-          <<: *except-main-and-release
-      - run-firebase-tests:
-          requires:
-            - hold
-          <<: *except-main-and-release
```
tonidero, line 387: I think this alias can be removed now?

tonidero (approved):

PR #830, opened 2023-10-04
Add support for showing in-app messages

MarkVillacampa:


revenuecat_examples/purchase_tester/lib/src/app.dart
```diff
@@ -55,6 +55,8 @@ class _MyAppState extends State<InitialScreen> {
       }
     });
 
+    Purchases.showStoreMessages(types: {InAppMessageType.billingIssue});
```
tonidero, line 58: Also need to remove this. Maybe we can add a button somewhere to call this if needed.

ios/Classes/PurchasesFlutterPlugin.m
```diff
@@ -201,6 +203,9 @@ - (void)handleMethodCall:(FlutterMethodCall *)call
         NSNumber *callbackID = arguments[@"callbackID"];
         [self startPromotedProductPurchase:callbackID
                                     result:result];
+    } else if ([@"showInAppMessages" isEqualToString:call.method]) {
+        NSSet<NSNumber*>* types = arguments[@"types"];
```
tonidero, line 207: Oh interesting, so this converts directly to an `NSSet`? I thought it would be an NSArray and we can convert it internally to an NSSet

ios/Classes/PurchasesFlutterPlugin.m
```diff
@@ -48,12 +48,14 @@ - (void)handleMethodCall:(FlutterMethodCall *)call
         NSString *appUserID = arguments[@"appUserId"];
         BOOL observerMode = [arguments[@"observerMode"] boolValue];
         BOOL usesStoreKit2IfAvailable = [arguments[@"usesStoreKit2IfAvailable"] boolValue];
+        BOOL shouldShowInAppMessagesAutomatically = [arguments[@"shouldShowInAppMessagesAutomatically"] boolValue];
```
tonidero, line 51: If this is null/parameter not passed, will it default to YES?

lib/purchases_flutter.dart
```diff
@@ -985,6 +986,21 @@ class Purchases {
     );
   }
 
+  /// Displays the specified store in-app message types to the user if there are any available to be shown.
+  /// - Important: This should only be used if you disabled these messages from showing automatically
+  /// during SDK configuration setting ``shouldShowInAppMessagesAutomatically`` to ``false``.
+  ///
+  /// @param [types] The types of messages to show.
+  static Future<void> showInAppMessages({
```
tonidero, line 994: Should we add this to the API tests for flutter? Both with and without the parameter.

lib/purchases_flutter.dart
```diff
@@ -985,6 +986,21 @@ class Purchases {
...
+    Set<InAppMessageType>? types,
+  }) =>
+      _channel.invokeMethod(
+        'showInAppMessages',
+        {
+          'types': types?.map((e) => e.index).toList(),
```
tonidero, line 1000: Here we're converting the set to a list, which makes sense, but we should make sure we parse the list in objc.

revenuecat_examples/purchase_tester/lib/main.dart
```diff
@@ -36,6 +36,7 @@ Future<void> _configureSDK() async {
   } else {
     configuration = PurchasesConfiguration(StoreConfig.instance.apiKey);
   }
+  configuration.shouldShowInAppMessagesAutomatically = false;
```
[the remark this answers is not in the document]
tonidero: Bump

ios/Classes/PurchasesFlutterPlugin.m
MarkVillacampa: Actually, I was overly optimistic about the casting. It doesn't do proper automatic conversion :D good catch

revenuecat_examples/purchase_tester/lib/main.dart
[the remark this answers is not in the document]
MarkVillacampa: I think in order for the`showInAppMessages` button to work properly I need to configure the SDK in this way?

lib/purchases_flutter.dart
MarkVillacampa: Already added some to `test/purchases_flutter_test.dart` (last file in the list). Anything extra I should add?

revenuecat_examples/purchase_tester/lib/main.dart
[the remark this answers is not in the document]
tonidero: Hmm that's true, but I think it could be inconvenient to disable the default behavior in our test app. I would leave the default, and maybe a comment on the method indicating that it will only work if the code is changed to disable messages from showing automatically?

lib/purchases_flutter.dart
tonidero: I think we also need to add these to the API tests. For example: https://github.com/RevenueCat/purchases-flutter/blob/main/api_tester/lib/api_tests/purchases_flutter_api_test.dart#L7

api_tester/lib/api_tests/purchases_flutter_api_test.dart
```diff
@@ -494,4 +501,9 @@ class _PurchasesFlutterApiTest {
     Future<void> future = Purchases.syncObserverModeAmazonPurchase(
         productID, receiptID, amazonUserID, isoCurrencyCode, price);
   }
+
+  void _showInAppMessages() async {
+    Future<void> future = Purchases.showInAppMessages(types: {InAppMessageType.billingIssue,
+      InAppMessageType.priceIncreaseConsent, InAppMessageType.generic});
```
tonidero, line 507: We should add one more API test for this enum, to make sure we validate all possible values of the enum

PR #860, opened 2023-11-13
`Purchase Tester`: remove unused `Activity`

NachoSoto:


revenuecat_examples/purchase_tester/android/app/src/main/java/com/revenuecat/purchases_sample/EmbeddingV1Activity.java
```diff
@@ -1,15 +0,0 @@
-package com.revenuecat.purchases_sample;
-
-import android.os.Bundle;
-import com.revenuecat.purchases_flutter.PurchasesFlutterPlugin;
-import io.flutter.app.FlutterActivity;
```
tonidero, line 5: I've been reading a little bit about this: https://docs.flutter.dev/release/breaking-changes/plugin-api-migration. Seems that we had this to ensure compatibility with V1 of the plugin APIs. Looks like there are 2 different imports for `FlutterActivity`, and this uses the old one.

Having said that, I saw that we are not actually testing this activity, as it mentions on that guide. So I think it should be ok to remove it? cc @vegaro 

Also, we currently support flutter [>=1.10.0](https://github.com/RevenueCat/purchases-flutter/blob/main/pubspec.yaml#L11), whereas the new APIs were added in 1.12.0. We might want to increase that min version if/when we do a major version in the future.

tonidero (approved):
I think since we're not actually testing the activity, this is really unused and should be safe to remove.

revenuecat_examples/purchase_tester/android/app/src/main/java/com/revenuecat/purchases_sample/EmbeddingV1Activity.java
NachoSoto: Thanks for looking into this 👍🏻

vegaro: Yes, it should be safe to remove. I think I left it because it's getting compiled, but it was adding the plugin using a string, which won't fail on compile time

PR #903, opened 2023-12-21
Update to use PHC PaywallHelpers instead of custom activity

tonidero:
After https://github.com/RevenueCat/purchases-hybrid-common/pull/602, we can now use the `PaywallHelpers.presentPaywallFromFragment` method to present the paywall and receive a result back using an internal view model.

This makes it so the `PurchasesFlutterActivity` isn't needed anymore and we can make use of that functionality, shared with other hybrids. 

Note that this does mean developers need to make sure their android activity inherits from `FlutterFragmentActivity` (or other `FragmentActivity`) and not `FlutterActivity`, as we need to present the fragment.

### TODO
- [x] Hide action bar: https://github.com/RevenueCat/purchases-android/pull/1532
- [x] Update PHC with those 2 changes.


android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
```diff
@@ -130,11 +134,11 @@ public Activity getActivity() {
         return registrar != null ? registrar.activity() : activity;
     }
 
-    private @Nullable PurchasesFlutterActivity getActivityFragment() {
+    private @Nullable FlutterFragmentActivity getActivityFragment() {
```
NachoSoto, line 137: Let's update the docs and put this in the release notes for the next beta :)

android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
```diff
@@ -130,11 +134,11 @@ public Activity getActivity() {
...
         final Activity activity = getActivity();
 
-        if (activity instanceof PurchasesFlutterActivity) {
-            return (PurchasesFlutterActivity) activity;
+        if (activity instanceof FlutterFragmentActivity) {
+            return (FlutterFragmentActivity) activity;
         } else {
             Log.e(TAG, "Paywalls require your activity to subclass FlutterFragmentActivity");
```
NachoSoto, line 143: We need to update this error now.

vegaro (approved):
LGTM

android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
tonidero: I believe our only docs for this beta currently live in the release notes right? Will update those when I make the release

Issue #902, opened 2023-12-20
Fatal Crash when using Flutter 6.3.0-beta.2 Paywall


bagintz:
using 6.3.0-beta.2 I am able to get the paywall to show up correctly but when I click the X to close on Android I get a fatal error that closes the app with the following error:

`D/RevenueCatUI(12225): Paywalls: Close paywall initiated
D/[Purchases] - DEBUG(12225): Tracking paywall event: PaywallEvent(creationData=CreationData(id=f46492e4-dc43-442e-9b9e-e53deb48f1a7, date=Wed Dec 20 14:47:54 EST 2023), data=Data(offeringIdentifier=default, paywallRevision=1, sessionIdentifier=62343b21-9f11-4269-b0b7-94047ad6e45c, displayMode=full_screen, localeIdentifier=en_US, darkMode=true), type=CLOSE)
D/CompatibilityChangeReporter(12225): Compat change id reported: 78294732; UID 10507; state: ENABLED
D/VRI[MainActivity](12225): applyTransactionOnDraw applyImmediately
D/AndroidRuntime(12225): Shutting down VM
E/AndroidRuntime(12225): FATAL EXCEPTION: main
E/AndroidRuntime(12225): Process: [MYAPP], PID: 12225
E/AndroidRuntime(12225): java.lang.RuntimeException: Failure delivering result ResultInfo{who=null, request=2112266436, result=0, data=null} to activity {[MYAPP]/....MainActivity}: java.lang.IllegalStateException: Reply already submitted
E/AndroidRuntime(12225): 	at android.app.ActivityThread.deliverResults(ActivityThread.java:5715)
E/AndroidRuntime(12225): 	at android.app.ActivityThread.handleSendResult(ActivityThread.java:5754)
E/AndroidRuntime(12225): 	at android.app.servertransaction.ActivityResultItem.execute(ActivityResultItem.java:69)
E/AndroidRuntime(12225): 	at android.app.servertransaction.ActivityTransactionItem.execute(ActivityTransactionItem.java:60)
E/AndroidRuntime(12225): 	at android.app.servertransaction.TransactionExecutor.executeNonLifecycleItem(TransactionExecutor.java:195)
E/AndroidRuntime(12225): 	at android.app.servertransaction.TransactionExecutor.executeCallbacks(TransactionExecutor.java:157)
E/AndroidRuntime(12225): 	at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:90)
[...]

PR #904, opened 2023-12-22
Fix `presentPaywallIfNeeded`

tonidero:
We forgot a `break` when calling this method, which caused it to fall to the `default` case, resulting in the result being called twice, causing a crash. This deals with #902


android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
```diff
@@ -371,6 +371,7 @@ public void onMethodCall(@NonNull MethodCall call, @NonNull Result result) {
             case "presentPaywallIfNeeded":
                 final String requiredEntitlementIdentifier = call.argument("requiredEntitlementIdentifier");
                 presentPaywall(result, requiredEntitlementIdentifier);
+                break;
```
tonidero, line 374: We probably would want to also add some tests for these but didn't want to delay fixing this... Definitely we should improve our coverage in the hybrids.

vegaro: oh wow, yes. Nice find

vegaro (approved):

tonidero:
The freezed tests are unrelated to this PR

Issue #909, opened 2023-12-23
productPlanIdentifier parameter missing from EntitlementInfo 


splangi:
- [x] I have updated Purchases SDK to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-flutter/blob/main/CONTRIBUTING.md)
- [x] I have searched the [Community](https://community.revenuecat.com)
- [x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [x] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-flutter/issues)

# ‼️ Required data ‼️ #
**Do not remove any of the steps from the template below. If a step is not applicable to your issue, please leave that step empty.**

There are a lot of things that can contribute to things not working. Having a very basic understanding of your environment will help us understand your issue faster!

## Environment ##
- [ ] Output of `flutter doctor`
- [ ] Version of `purchases-flutter`
- [ ] Testing device version e.g.: iOS 15.5, Android API 30, etc.
- [ ] How often the issue occurs- every one of your customers is impacted? Only in dev?
- [ ] [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue
- [ ] Steps to reproduce, with a description of expected vs. actual behavior
**Other information** (e.g. stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc.)

## Describe the bug ##
productPlanIdentifier is missing from EntitlementInfo making it impossible to identify a base plan for the subscription which gave this entitlement. 

## Additional context ##
This is available on the Android SDK and is being encoded when sent to flutter side, but it is not being decoded. Only adding the parameter to EntitlementInfo and re-running the build_runner codegen will fix the issue.

RCGitBot:
👀 We've just linked this [issue to our internal tracker](https://revenuecat.zendesk.com/agent/tickets/35027) and notified the team. Thank you for reporting, we're checking this out!

rglanz-rc:
Hi @splangi,

I'm going to mark this issue as closed as I see you are working with Cesar in a PR. Let me know if you have any more questions on this.

PR #910, opened 2023-12-25
Adds productPlanIdentifier parameter to EntitlementInfo

splangi:

  - [x] Adds missing productPlanIdentifier parameter to EntitlementInfo class.

  - [x] #909 

  - [ ] No unit tests added/modified


vegaro (changes requested):
Thanks for opening this PR! We missed that property indeed.

This will be a breaking change so `productPlanIdentifier` should be an optional property to not be breaking. 

Also, can you add it to the `_checkConstructor` function in `entitlement_info_wrapper_api_test.dart` file.

Thank you!

vegaro:
merging into our own branch to make those updates

Issue #886, opened 2023-11-28
presentPaywallIfNeeded always return Null


imSaharukh:
```
  if (Platform.isIOS) {
    //get all Entitlement
    var purchaserInfo = await Purchases.getCustomerInfo();
    var entitlements = purchaserInfo.entitlements.all;

    log("entitlements $entitlements");

    var result = await Purchases.presentPaywallIfNeeded(
        entitlements.values.first.identifier);

    log("result $result");

    return;
  }
```

i'm always getting flutter: type 'Null' is not a subtype of type 'FutureOr<bool>' error

tonidero:
Hi @imSaharukh, thanks for reporting this. Indeed this is a problem in the current beta of paywalls. We're working on a fix and will update this issue once it's shipped.

In the meantime, we suggest handling this error as a non-fatal exception. Not ideal, but we hope to have it fixed soon.

BorisKest:
Same here. BTW are you going to make it possible to use a paywall from different offerings rather than the current one?

NachoSoto:
Thanks for reporting! A few questions:
- Is this happening on Android, iOS, or both?
- Are you seeing anything in the logs?

Looking at the code, if this is Android only, I believe what might be happening is that your `MainActivity` probably still inherits `FlutterActivity`, and you need to change it to inherit from `PurchasesFlutterActivity` to support paywalls.

I just updated the implementation so an error is forwarded instead: 704bd671816f2793835306b59d9ab2120fb2f4da

tonidero:
Hi everyone! We're aware of this issue. #916 should fix it temporarily by removing the return value from `presentPaywall`/`presentPaywallIfNeeded`. We plan to add a more future-proof API later so we can receive more data of what happened with the paywall.

tonidero:
> Same here. BTW are you going to make it possible to use a paywall from different offerings rather than the current one?

Hi @BorisKest, sorry for missing this! Yeah, that's part of the roadmap but might not be in V1 of paywalls for flutter.

PR #916, opened 2024-01-02
Remove result from presentPaywall methods

tonidero:
Currently the `presentPaywall` and `presentPaywallIfNeeded` methods return a boolean indicating whether a purchase happened in the paywall. This API is confusing and not extensible and it wasn't properly working in iOS, see: #886. This first PR removes the return value from the method. In a future PR we will add proper result handling for the paywalls. Until then, the clients won't know what happened when the paywall was presented and will need to rely on the `Purchases.getCustomerInfo()` method to respond appropriately.

This is a breaking change from the previous beta.

### DIFF
--- a/android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
+++ b/android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
@@ -751,9 +751,9 @@ public class PurchasesFlutterPlugin implements FlutterPlugin, MethodCallHandler,
         if (fragment != null) {
             presentPaywallFromFragment(
                     fragment,
-                    requiredEntitlementIdentifier,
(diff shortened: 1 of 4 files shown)


android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
```diff
@@ -751,9 +751,9 @@ private void presentPaywall(final Result result, final @Nullable String required
         if (fragment != null) {
             presentPaywallFromFragment(
                     fragment,
-                    requiredEntitlementIdentifier,
-                    paywallResult -> result.success(paywallResult instanceof PaywallResult.Purchased)
+                    requiredEntitlementIdentifier
             );
+            result.success(null);
```
tonidero, line 756: In iOS, we were already just returning null, so no changes needed there.

vegaro (approved):

PR #919, opened 2024-01-03
Create `purchases_flutter_ui` package

tonidero:
Creates `purchases_flutter_ui` flutter package in a folder with the functionality to `presentPaywall`/`presentPaywallIfNeeded` and removes it from the main sdk `purchases_flutter`.

This allows to keep the min sdk version of the main sdk lower, while still requiring a higher minSdk version for the new library.

The way to import the library now is to add in the pubspec.yaml:
```
dependencies:
   purchases_flutter: X.Y.Z
   purchases_flutter_ui: X.Y.Z
```


revenuecat_examples/purchase_tester/lib/src/app.dart
```diff
@@ -3,6 +3,7 @@ import 'dart:async';
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
 import 'package:purchases_flutter/purchases_flutter.dart';
+import 'package:purchases_flutter_ui/purchases_flutter_ui.dart';
```
tonidero, line 6: This is the new import, but the class is called `RevenueCatUI`. I thought it was ok, but lmk if you would prefer to rename the package as well.

purchases_flutter_ui/lib/purchases_flutter_ui.dart
```diff
@@ -0,0 +1,20 @@
+
+import 'package:flutter/services.dart';
+
+class RevenueCatUI {
```
tonidero, line 4: I modified the generated code from the wizard to resemble more closely what we do in the existing plugin. The generated code changed things so everything wasn't all static methods and better supported different communication systems with the platform (other than methodchannel). But it felt more complex so I simplified it like this.

purchases_flutter_ui/android/src/main/kotlin/com/revenuecat/purchases_flutter_ui/PurchasesFlutterUiPlugin.kt
```diff
@@ -0,0 +1,92 @@
...
+import io.flutter.plugin.common.MethodChannel.Result
+
+class PurchasesFlutterUiPlugin: FlutterPlugin, MethodCallHandler, ActivityAware {
+    private val TAG = "PurchasesFlutter"
+
+    private var activity: Activity? = null
+
+    private lateinit var channel : MethodChannel
+
+    override fun onAttachedToEngine(flutterPluginBinding: FlutterPlugin.FlutterPluginBinding) {
```
tonidero, line 25: In the existing plugin we also have the `registerWith` static method. That seems to be used for very old flutter versions, but we are not supporting those so I think it's ok to just have this.

purchases_flutter_ui/example/README.md
```diff
@@ -0,0 +1,6 @@
+# Examples
+Our examples are located at [revenuecat_examples](https://github.com/RevenueCat/purchases-flutter/tree/main/revenuecat_examples).
+
+### [Purchase Tester](https://github.com/RevenueCat/purchases-flutter/tree/main/revenuecat_examples/purchase_tester)
```
tonidero, line 4: I didn't add MagicWeather here since it's not using paywalls at the moment.

tonidero:
Note that I'm not reverting the minSdkVersion changes in the `paywalls` branch on the `api-tester` and `purchase-tester` apps, since those are still using the new library which does require the higher minSdk version.

.pubignore
```diff
@@ -0,0 +1,26 @@
+revenuecat_examples
+scripts
+migrations
+fastlane
+api_tester
+codegen.sh
+Dangerfile
+Gemfile
+Gemfile.lock
+pubspec.lock
```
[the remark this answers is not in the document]
vegaro: amazing

fastlane/Fastfile
```diff
@@ -141,6 +164,9 @@ lane :update_hybrid_common do |options|
       'ios/purchases_flutter.podspec' => ['s.dependency \'PurchasesHybridCommon\', \'{x}\''],
       'macos/purchases_flutter.podspec' => ['s.dependency \'PurchasesHybridCommon\', \'{x}\''],
       'android/build.gradle' => ['ext.common_version = \'{x}\''],
+      'purchases_flutter_ui/ios/purchases_flutter.podspec' => ['s.dependency \'PurchasesHybridCommon\', \'{x}\''],
```
vegaro, line 167: should `purchases_flutter.podspec` be `purchases_flutter_ui.podspec`?

purchases_flutter_ui/README.md
```diff
@@ -0,0 +1,15 @@
...
+## Getting Started
+
+This project is a starting point for a Flutter
+[plug-in package](https://flutter.dev/developing-packages/),
+a specialized package that includes platform-specific implementation code for
+Android and/or iOS.
+
+For help getting started with Flutter development, view the
+[online documentation](https://flutter.dev/docs), which offers tutorials,
+samples, guidance on mobile development, and a full API reference.
```
vegaro, line 14: Maybe update the text here

purchases_flutter_ui/analysis_options.yaml
```diff
@@ -0,0 +1,4 @@
+include: package:flutter_lints/flutter.yaml
```
vegaro, line 1: we have some things specified in this file in the root

purchases_flutter_ui/android/src/main/kotlin/com/revenuecat/purchases_flutter_ui/PurchasesFlutterUiPlugin.kt
```diff
@@ -0,0 +1,92 @@
...
+import io.flutter.embedding.engine.plugins.FlutterPlugin
+import io.flutter.embedding.engine.plugins.activity.ActivityAware
+import io.flutter.embedding.engine.plugins.activity.ActivityPluginBinding
+import io.flutter.plugin.common.MethodCall
+import io.flutter.plugin.common.MethodChannel
+import io.flutter.plugin.common.MethodChannel.MethodCallHandler
+import io.flutter.plugin.common.MethodChannel.Result
+
+class PurchasesFlutterUiPlugin: FlutterPlugin, MethodCallHandler, ActivityAware {
+    private val TAG = "PurchasesFlutter"
```
vegaro, line 19: should this be `PurchasesFlutterUi`?

purchases_flutter_ui/android/src/main/kotlin/com/revenuecat/purchases_flutter_ui/PurchasesFlutterUiPlugin.kt
```diff
@@ -0,0 +1,92 @@
...
+    override fun onReattachedToActivityForConfigChanges(binding: ActivityPluginBinding) {
+        onAttachedToActivity(binding)
+    }
+
+    override fun onDetachedFromActivity() {
+        activity = null
+    }
+
+    private fun presentPaywall(result: Result, requiredEntitlementIdentifier: String?) {
+        val activity = getActivityFragment()
```
vegaro, line 64: Is this an `activity` or a `fragment`? Because the function is `presentPaywallFromFragment(activity,`

vegaro (approved):

purchases_flutter_ui/analysis_options.yaml
tonidero: Good point! I think I will just create a symlink for this, since the ignored files in the root shouldn't matter for this one.

vegaro (approved):

PR #926, opened 2024-01-05
Add result to presentPaywall methods

tonidero:
Adds a `PaywallResult` enum and makes it the return value of the `presentPaywall` and `presentPaywallIfNeeded` methods

### DIFF
--- a/purchases_ui_flutter/lib/paywall_result.dart
+++ b/purchases_ui_flutter/lib/paywall_result.dart
@@ -1,0 +1,7 @@
+enum PaywallResult {
+  notPresented,
+  cancelled,
+  error,
+  purchased,
+  restored
+}
--- a/purchases_ui_flutter/lib/paywall_result.dart
+++ b/purchases_ui_flutter/lib/paywall_result.dart
@@ -1,7 +1,18 @@
+/// Possible values for the result of a paywall presentation.
 enum PaywallResult {
+  /// Only returned when using [RevenueCatUI.presentPaywallIfNeeded]. Returned
+  /// if the paywall was not presented.
   notPresented,
+  /// Returned when the paywall was presented and the user cancelled without
+  /// executing an action.
   cancelled,
+  /// Returned when the paywall was presented and an error occurred performing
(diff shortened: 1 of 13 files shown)


revenuecat_examples/purchase_tester/lib/src/app.dart
```diff
@@ -147,7 +148,8 @@ class _UpsellScreenState extends State<UpsellScreen> {
 
         buttonThings.add(ElevatedButton(
           onPressed: () async {
-            await RevenueCatUI.presentPaywall();
+            final paywallResult = await RevenueCatUI.presentPaywall();
+            log('Paywall result: $paywallResult');
```
tonidero, line 152: Right now I'm only logging this. I did run into some issues getting this log to show for iOS, so I used the `dart:developer` log method for that which fixed the issue for me.

tonidero:
Still need to update PHC with the fixes in https://github.com/RevenueCat/purchases-hybrid-common/pull/639 but tested locally and seems to be working fine now!

purchases_ui_flutter/lib/paywall_result.dart
```diff
@@ -0,0 +1,18 @@
+/// Possible values for the result of a paywall presentation.
+enum PaywallResult {
+  /// Only returned when using [RevenueCatUI.presentPaywallIfNeeded]. Returned
+  /// if the paywall was not presented.
+  notPresented,
+  /// Returned when the paywall was presented and the user cancelled without
+  /// executing an action.
+  cancelled,
```
NachoSoto, line 8: What if you tap on purchase, cancel, and then purchase again? Would you only get this?

tonidero: You would only get the last result before the paywall closed. In this case, the purchase.

NachoSoto: Ooooh when it's closed, got it 👍🏻

NachoSoto (approved):

PR #950, opened 2024-01-25
Refactor and cleanup of purchase_tester

vegaro:
Extracted screens into their own file and improved UpsellScreen by adding `Card` widgets. The previous version wasn't even scrollable.

<table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>
  <tr>
    <td>
      Before
    </td>
    <td>
      After
    </td>
  </tr>
</table>


revenuecat_examples/purchase_tester/lib/src/cats.dart
```diff
@@ -0,0 +1,52 @@
+import 'package:flutter/material.dart';
+import 'package:purchases_flutter/purchases_flutter.dart';
+
+import 'app.dart';
```
tonidero, line 4: Hmm do we have a circular reference here? Not sure if it's not complaining though...

tonidero (approved):
Nice, this was much overdue!

revenuecat_examples/purchase_tester/lib/src/cats.dart
vegaro: I was also surprised but it's apparently allowed in dart, I think this is only for the initial screen. Do you think it's worth it to extract it to remove the cicle?

tonidero: Hmm yeah, seems to be supported in dart... It might be a good idea, but I'm ok leaving it for now if it would take too much since it's just a tester app.

PR #954, opened 2024-01-26
PaywallView support

vegaro:
Added PaywallView for iOS and Android. Note this requires bumping the compile version in android to 34 if you're using purchases_ui_flutter.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -10,7 +10,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:8.2.0'
+        classpath 'com.android.tools.build:gradle:8.0.2'
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
     }
 }
@@ -26,7 +26,7 @@ apply plugin: 'com.android.library'
 apply plugin: 'kotlin-android'
 
 android {
-    compileSdk 34
+    compileSdk 31
 
     // Conditional for compatibility with AGP <4.2.
     if (project.android.hasProperty("namespace")) {
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -10,7 +10,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:8.0.2'
+        classpath 'com.android.tools.build:gradle:8.2.0'
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
     }
 }
@@ -26,7 +26,7 @@ apply plugin: 'com.android.library'
 apply plugin: 'kotlin-android'
 
 android {
-    compileSdkVersion 31
+    compileSdk 34
 
     // Conditional for compatibility with AGP <4.2.
     if (project.android.hasProperty("namespace")) {
--- a/purchases_ui_flutter/android/src/main/kotlin/com/revenuecat/purchases_ui_flutter/views/PaywallView.kt
+++ b/purchases_ui_flutter/android/src/main/kotlin/com/revenuecat/purchases_ui_flutter/views/PaywallView.kt
@@ -1,6 +1,7 @@
 package com.revenuecat.purchases_ui_flutter.views
 
 import android.content.Context
+import android.util.AttributeSet
 import android.view.View
 import com.revenuecat.purchases.ui.revenuecatui.ExperimentalPreviewRevenueCatUIPurchasesAPI
 import io.flutter.plugin.platform.PlatformView
@@ -8,8 +9,9 @@ import com.revenuecat.purchases.ui.revenuecatui.views.PaywallView as NativePaywa
 
 @OptIn(ExperimentalPreviewRevenueCatUIPurchasesAPI::class)
 internal class PaywallView(
-    context: Context, id: Int,
-    creationParams: Map<String?, Any?>?
+    context: Context,
+    id: Int,
+    creationParams: Map<String?, Any?>
 ) : PlatformView {
 
     private val nativePaywallView: NativePaywallView
@@ -21,6 +23,9 @@ internal class PaywallView(
     override fun dispose() {}
 
     init {
-        nativePaywallView = NativePaywallView(context)
+        val offeringIdentifier = creationParams["offeringIdentifier"] as String?
+        nativePaywallView = NativePaywallView(context,)
+        // TODO add to constructor
+        nativePaywallView.setOfferingId(offeringIdentifier)
     }
 }
--- a/purchases_ui_flutter/android/src/main/kotlin/com/revenuecat/purchases_ui_flutter/views/PaywallView.kt
+++ b/purchases_ui_flutter/android/src/main/kotlin/com/revenuecat/purchases_ui_flutter/views/PaywallView.kt
@@ -4,8 +4,8 @@ import android.content.Context
 import android.util.AttributeSet
 import android.view.View
 import com.revenuecat.purchases.ui.revenuecatui.ExperimentalPreviewRevenueCatUIPurchasesAPI
-import io.flutter.plugin.platform.PlatformView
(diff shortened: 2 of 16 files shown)


purchases_ui_flutter/android/src/main/kotlin/com/revenuecat/purchases_ui_flutter/views/PaywallView.kt
```diff
@@ -0,0 +1,31 @@
...
+    override fun getView(): View {
+        return nativePaywallView
+    }
+
+    override fun dispose() {}
+
+    init {
+        val offeringIdentifier = creationParams["offeringIdentifier"] as String?
+        nativePaywallView = NativePaywallView(context,)
+        // TODO add to constructor
```
vegaro, line 28: this requires a phc release so doing it this way for now for the sake of shipping

purchases_ui_flutter/android/src/main/kotlin/com/revenuecat/purchases_ui_flutter/views/PaywallView.kt
```diff
@@ -0,0 +1,31 @@
+package com.revenuecat.purchases_ui_flutter.views
+
+import android.content.Context
+import android.util.AttributeSet
+import android.view.View
+import com.revenuecat.purchases.ui.revenuecatui.ExperimentalPreviewRevenueCatUIPurchasesAPI
+import io.flutter.plugin.platform.PlatformView
+import com.revenuecat.purchases.ui.revenuecatui.views.PaywallView as NativePaywallView
```
tonidero, line 8: Hmm I wonder if we should rename this class to `RNPaywallView` instead... but I'm ok with this for now.

tonidero (approved):
Looks good! I would hold until we have the iOS part ready, or merge into a different base branch though, so it's not broken in iOS.

android/build.gradle
```diff
@@ -26,7 +26,7 @@ apply plugin: 'com.android.library'
 apply plugin: 'kotlin-android'
 
 android {
-    compileSdkVersion 31
+    compileSdk 34
```
tonidero, line 29: I'm wondering if this was needed or if it was only on the `purchases_ui_flutter` library... However, I think it shouldn't hurt.

android/build.gradle
```diff
@@ -10,7 +10,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:8.0.2'
+        classpath 'com.android.tools.build:gradle:8.2.0'
```
tonidero, line 13: While this shouldn't cause any issues, was this needed? So we can extract to a different PR (In purchases Android we currently have 8.1.1)

purchases_ui_flutter/android/src/main/kotlin/com/revenuecat/purchases_ui_flutter/views/PaywallViewFactory.kt
```diff
@@ -0,0 +1,14 @@
+package com.revenuecat.purchases_ui_flutter.views
+
+import android.content.Context
+import io.flutter.plugin.common.StandardMessageCodec
+import io.flutter.plugin.platform.PlatformView
+import io.flutter.plugin.platform.PlatformViewFactory
+
+class PaywallViewFactory : PlatformViewFactory(StandardMessageCodec.INSTANCE) {
```
tonidero, line 8: We could make this internal I think

purchases_ui_flutter/lib/views/paywall_view.dart
```diff
@@ -0,0 +1,61 @@
...
+    final creationParams = <String, dynamic>{
+      'offeringIdentifier': offering?.identifier,
+    };
+
+    return Platform.isAndroid
+        ? buildAndroidPlatformViewLink(creationParams)
+        : buildUiKitView(creationParams);
+  }
+
+  UiKitView buildUiKitView(Map<String, dynamic> creationParams) => UiKitView(
```
tonidero, line 28: These should be private (I made this change in #993)

purchases_ui_flutter/android/src/main/kotlin/com/revenuecat/purchases_ui_flutter/views/PaywallView.kt
```diff
@@ -0,0 +1,31 @@
+package com.revenuecat.purchases_ui_flutter.views
+
+import android.content.Context
+import android.util.AttributeSet
+import android.view.View
+import com.revenuecat.purchases.ui.revenuecatui.ExperimentalPreviewRevenueCatUIPurchasesAPI
```
tonidero, line 6: Total nitpick, but these imports need sorting :P

purchases_ui_flutter/ios/Classes/PurchasesUiPaywallView.swift
```diff
@@ -0,0 +1,61 @@
...
+    }
+}
+
+class PurchasesUiPaywallView: NSObject, FlutterPlatformView {
+    private var _view: UIView
+
+    init(
+        frame: CGRect,
+        viewIdentifier viewId: Int64,
+        arguments args: Any?,
```
tonidero, line 38: We are not getting and setting the offering id in iOS

tonidero (commented):
Left some comments... Ideally we would try to avoid upgrading as many things as possible in this PR, specially, I think we shouldn't touch the main library and the gradle plugin if possible. We also need to fix using a specific offering in iOS. Other than that look good

android/build.gradle
tonidero: I reverted this in my followup commits

android/build.gradle
tonidero: Same, I reverted this in my followup commits

purchases_ui_flutter/android/src/main/kotlin/com/revenuecat/purchases_ui_flutter/views/PaywallViewFactory.kt
tonidero: Done in my followup commits

purchases_ui_flutter/lib/views/paywall_view.dart
tonidero: Done in my followup commits

purchases_ui_flutter/ios/Classes/PurchasesUiPaywallView.swift
tonidero: Done in https://github.com/RevenueCat/purchases-flutter/pull/954/commits/7c0c6efce95382e5ba326c592fb569c4aa0d024a

revenuecat_examples/purchase_tester/lib/src/paywall.dart
```diff
@@ -0,0 +1,31 @@
...
+
+  @override
+  void initState() {
+    super.initState();
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return Scaffold(
+      body: SafeArea( // Wrap your body content with SafeArea
```
vegaro, line 24: let's remove that comment `// Wrap your body content with SafeArea`

revenuecat_examples/purchase_tester/lib/src/upsell.dart
```diff
@@ -103,6 +104,18 @@ class _UpsellScreenState extends State<UpsellScreen> {
...
+                  ),
+                  ElevatedButton(
+                    onPressed: () async {
+                      Navigator.push(
+                        context,
+                        MaterialPageRoute(
+                            builder: (context) => PaywallScreen(
+                                  offering: offering,
+                                )),
+                      );
```
vegaro, line 116: I think this needs some formatting

vegaro (commented):
I can't approve it because it's my PR but looks good

purchases_ui_flutter/ios/Classes/PurchasesUiPaywallView.swift
tonidero: Yup

PR #955, opened 2024-01-26
PaywallView iOS

vegaro:
Added PaywallView support for iOS

I have to say first time the screen loads there are some hiccups, I still don't know where they are coming from, possibly from loading the image in the template, will need to investigate further, but it works

### DIFF
--- a/purchases_ui_flutter/ios/Classes/PurchasesUiPaywallView.swift
+++ b/purchases_ui_flutter/ios/Classes/PurchasesUiPaywallView.swift
@@ -1,0 +1,62 @@
+import Flutter
+import UIKit
+import PurchasesHybridCommonUI
+import RevenueCatUI
+
+class PurchasesUiPaywallViewFactory: NSObject, FlutterPlatformViewFactory {
+    private var messenger: FlutterBinaryMessenger
+
+    init(messenger: FlutterBinaryMessenger) {
+        self.messenger = messenger
+        super.init()
+    }
+
+    func create(
+        withFrame frame: CGRect,
+        viewIdentifier viewId: Int64,
+        arguments args: Any?
+    ) -> FlutterPlatformView {
+        return PurchasesUiPaywallView(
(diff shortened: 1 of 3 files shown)


purchases_ui_flutter/ios/Classes/PurchasesUiPaywallView.swift
```diff
@@ -0,0 +1,62 @@
...
+    ) {
+        if #available(iOS 15.0, *) {
+            let paywallProxy = PaywallProxy()
+            guard let paywallView = paywallProxy.createPaywallView().view else {
+                print("Error: error getting PaywallView.")
+                _view = UIView()
+                super.init()
+                return
+            }
+            paywallView.clipsToBounds = true
```
vegaro, line 49: Had to add this otherwise the image in the header of the template would render beyond the screen and would be seen when returning to the previous screen

NachoSoto: We shouldn't do this. Do you have a screenshot of what it looks like without it?

purchases_ui_flutter/ios/Classes/PurchasesUiPaywallView.swift
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    /// Implementing this method is only necessary when the `arguments` in `createWithFrame` is not `nil`.
+    public func createArgsCodec() -> FlutterMessageCodec & NSObjectProtocol {
+          return FlutterStandardMessageCodec.sharedInstance()
+    }
+}
+
+class PurchasesUiPaywallView: NSObject, FlutterPlatformView {
+    private var _view: UIView
```
NachoSoto, line 33: This can be immutable. Also underscores aren't the Swift convention:
```suggestion
    private let view: UIView
```

NachoSoto (commented):
Exciting!

PR #965, opened 2024-02-05
Bump PHC 6.3.1

tonidero:
Bumps PHC to 6.3.1


NachoSoto (approved):

tonidero:
Tests are fixed on `main` but since we're doing a hotfix, not going to fix them, in order to keep it the same way as the previous release.

PR #985, opened 2024-02-20
Fix ios integration tests

tonidero:
ios integration tests were failing getting stuck in macos x86 machines. This updates to using M1 machines and updates to newer Xcode version, which required some small changes in the integration tests.

### DIFF
--- a/revenuecat_examples/purchase_tester/integration_test/app_test.dart
+++ b/revenuecat_examples/purchase_tester/integration_test/app_test.dart
@@ -1,5 +1,4 @@
 // Imports the Flutter Driver API.
-import 'package:flutter/foundation.dart';
 import 'package:flutter/services.dart';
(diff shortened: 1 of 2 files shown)


revenuecat_examples/purchase_tester/integration_test/app_test.dart
```diff
@@ -8,18 +7,18 @@ import 'package:purchases_flutter/purchases_flutter.dart';
 void main() {
   IntegrationTestWidgetsFlutterBinding.ensureInitialized();
 
-  var uuid = UniqueKey().toString();
+  var userId = "integration-test-${DateTime.now().millisecondsSinceEpoch}";
```
tonidero, line 10: This was needed after upgrading to xcode 15.2. With the old system, the encoding of some characters didn't match the one returned from the SDK...

vegaro (approved):

PR #996, opened 2024-03-01
`PaywallFooterView` support

tonidero:
This adds initial support for the `PaywallFooterView`. Includes iOS support from #999 

https://github.com/RevenueCat/purchases-flutter/assets/808417/427341bc-d93e-4f87-bf9c-2207d2cff2dd

image

- [x] Holding until iOS is ready #999

### DIFF
--- a/purchases_ui_flutter/android/src/main/kotlin/com/revenuecat/purchases_ui_flutter/views/PaywallFooterView.kt
+++ b/purchases_ui_flutter/android/src/main/kotlin/com/revenuecat/purchases_ui_flutter/views/PaywallFooterView.kt
@@ -35,30 +35,6 @@ internal class PaywallFooterView(
         methodChannel = MethodChannel(messenger, "purchases_ui_flutter/PaywallFooterView/${id}")
         val offeringIdentifier = creationParams["offeringIdentifier"] as String?
         nativePaywallFooterView = object : NativePaywallFooterView(context) {
-//
-//            var shouldRemeasure = true
-//
-//            override fun requestLayout() {
-//                super.requestLayout()
-//                if (shouldRemeasure) {
-////                    post(measureAndLayout)
-//                }
-//            }
-//
-//            private val measureAndLayout = Runnable {
-//                Log.e("TEST", "MEASURE AND LAYOUT")
-//                measure(
-//                    MeasureSpec.makeMeasureSpec(width, MeasureSpec.EXACTLY),
-//                    MeasureSpec.makeMeasureSpec(height, MeasureSpec.EXACTLY)
-//                )
-//                layout(left, top, right, bottom)
-//            }
-//
-//            override fun onDetachedFromWindow() {
-//                super.onDetachedFromWindow()
-//                shouldRemeasure = false
-//            }
-
             public override fun onMeasure(widthMeasureSpec: Int, heightMeasureSpec: Int) {
                 super.onMeasure(widthMeasureSpec, heightMeasureSpec)
                 var maxWidth = 0
--- a/purchases_ui_flutter/android/src/main/kotlin/com/revenuecat/purchases_ui_flutter/views/PaywallFooterView.kt
+++ b/purchases_ui_flutter/android/src/main/kotlin/com/revenuecat/purchases_ui_flutter/views/PaywallFooterView.kt
@@ -1,0 +1,90 @@
+package com.revenuecat.purchases_ui_flutter.views
+
+import android.content.Context
+import android.util.AttributeSet
+import android.util.Log
+import android.view.Gravity
+import android.view.View
+import android.widget.FrameLayout
+import androidx.core.view.children
+import com.revenuecat.purchases.ui.revenuecatui.ExperimentalPreviewRevenueCatUIPurchasesAPI
+import com.revenuecat.purchases.ui.revenuecatui.views.PaywallFooterView as NativePaywallFooterView
+import io.flutter.plugin.common.BinaryMessenger
+import io.flutter.plugin.common.MethodChannel
+import io.flutter.plugin.platform.PlatformView
+
+@OptIn(ExperimentalPreviewRevenueCatUIPurchasesAPI::class)
+internal class PaywallFooterView(
+    context: Context,
+    id: Int,
+    messenger: BinaryMessenger,
+    creationParams: Map<String?, Any?>
+) : PlatformView {
+
+    private val methodChannel: MethodChannel
+
+    private val nativePaywallFooterView: NativePaywallFooterView
+
+    override fun getView(): View {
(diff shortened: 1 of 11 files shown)


tonidero:
There is a small delay when initially displaying the paywall view in android. Seems to be related to `PlatformView` performance issues:
 https://github.com/RevenueCat/purchases-flutter/assets/808417/f23dae50-c032-4812-8d51-29611b526897
I've tried to improve it without much luck... We could add an animation on top to make the transition cleaner but not sure if it's worth it, considering the paywall may be set anywhere within the app...

purchases_ui_flutter/android/src/main/kotlin/com/revenuecat/purchases_ui_flutter/views/PaywallFooterView.kt
```diff
@@ -0,0 +1,66 @@
...
+                setMeasuredDimension(finalWidth, finalHeight)
+                updateHeight(finalHeight.toDouble())
+            }
+        }
+        nativePaywallFooterView.layoutParams = FrameLayout.LayoutParams(
+            FrameLayout.LayoutParams.MATCH_PARENT,
+            FrameLayout.LayoutParams.MATCH_PARENT,
+            Gravity.BOTTOM
+        )
+        // TODO add to constructor
```
vegaro, line 59: I don't remember, do we need to do this?

purchases_ui_flutter/lib/views/paywall_footer_view.dart
```diff
@@ -0,0 +1,80 @@
...
+
+/// View that displays the paywall in footer mode.
+/// Not supported in macOS currently.
+///
+/// [offering] (Optional) The offering object to be displayed in the paywall.
+/// Obtained from [Purchases.getOfferings].
+///
+/// [content] The content to be displayed above the paywall. Make sure you
+/// apply a padding to the bottom of your content to avoid overlap. This padding
+/// can be obtained from [PaywallFooterView.roundedCornerRadius].
```
vegaro, line 14: Is there a way we can add the padding ourselves? In React Native we do add it

purchases_ui_flutter/lib/views/paywall_footer_view.dart
```diff
@@ -0,0 +1,80 @@
...
+class PaywallFooterView extends StatefulWidget {
+  /// The top corner radius of the footer view.
+  static const roundedCornerRadius = 20.0;
+
+  final Offering? offering;
+  final Widget content;
```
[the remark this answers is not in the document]
vegaro: Yeah I can see how someone might not want the scrollview... let's see how that is accepted in RN

purchases_ui_flutter/lib/views/paywall_footer_view.dart
```diff
@@ -0,0 +1,80 @@
...
+  @override
+  State<StatefulWidget> createState() => _PaywallFooterViewState();
+}
+
+class _PaywallFooterViewState extends State<PaywallFooterView> {
+
+  // Need to set it to a value > 0 so it's drawn. Setting it to a value that
+  // approximately reflects what the footer view height will be, so redrawing
+  // is not so noticeable. Need to improve this.
+  var _height = 264.0;
```
vegaro, line 37: Doesn't this depend a lot on the number of packages for example?

revenuecat_examples/purchase_tester/lib/src/paywall_footer_screen.dart
```diff
@@ -0,0 +1,44 @@
...
+  Widget build(BuildContext context) {
+    return Scaffold(
+      body: SafeArea( // Wrap your body content with SafeArea
+        child: Center(
+          child: PaywallFooterView(
+            offering: widget.offering,
+            content: Container(
+              color: Colors.blue.withAlpha(80),
+              child: SingleChildScrollView(
+                padding: const EdgeInsets.only(bottom: PaywallFooterView.roundedCornerRadius),
```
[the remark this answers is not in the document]
vegaro: we are adding the padding ourselves in Android and iOS right? I think we should match that behavior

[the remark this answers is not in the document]
tonidero: Well, in iOS, the system for the footer is different (it's not a container). In android it is and what we do is passing the padding as part of the function that accepts the children composables... I guess we could make the child here a function instead, that accepts a padding and returns a widget... I mostly wanted to do it this way to keep it in sync with the way it works in the native flutter components... But it might be cleaner yeah...

purchases_ui_flutter/lib/views/paywall_footer_view.dart
tonidero: We could add it, but then it would be much more difficult to extend the top view below the borders... Mentioned the alternative in this comment: https://github.com/RevenueCat/purchases-flutter/pull/996#discussion_r1512663338

purchases_ui_flutter/lib/views/paywall_footer_view.dart
tonidero: Yes, this is just an approximation, so it doesn't start with a non-visible footer and then it suddenly appears. This wouldn't be a problem without the delay I mentioned before when drawing using `PlatformView` in Android.

purchases_ui_flutter/android/src/main/kotlin/com/revenuecat/purchases_ui_flutter/views/PaywallFooterView.kt
tonidero: I think not (the native constructor accepts an `offering`, not an `offeringId`). Will remove the comment from here and the full screen versions.

revenuecat_examples/purchase_tester/lib/src/paywall_footer_screen.dart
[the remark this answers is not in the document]
tonidero: I made this change in https://github.com/RevenueCat/purchases-flutter/pull/996/commits/0624e6dbb251f66c1b9b54e2e3a6cc0f289cc214. Lmk what you think @vegaro

revenuecat_examples/purchase_tester/lib/src/paywall_footer_screen.dart
```diff
@@ -25,10 +25,10 @@ class _PaywallFooterScreenState extends State<PaywallFooterScreen> {
         child: Center(
           child: PaywallFooterView(
             offering: widget.offering,
-            content: Container(
+            contentCreator: (bottomPadding) => Container(
```
vegaro, line 28: ohh I like this

vegaro (approved):

PR #995, opened 2024-02-29
Add `PaywallView` listeners

vegaro:
Added listeners that can be used like:

```
return Scaffold(
      body: Center(
        child: PaywallView(
          onPurchaseStarted: (Package rcPackage) {
            print('Purchase started for package: ${rcPackage.identifier}');
          },
          onPurchaseCompleted:
              (CustomerInfo customerInfo, StoreTransaction storeTransaction) {
            print('Purchase completed for customerInfo:\n $customerInfo\n '
                'and storeTransaction:\n $storeTransaction');
          },
          onPurchaseError: (PurchasesError error) {
            print('Purchase error: $error');
          },
          onRestoreCompleted: (CustomerInfo customerInfo) {
            print('Restore completed for customerInfo:\n $customerInfo');
          },
          onRestoreError: (PurchasesError error) {
            print('Restore error: $error');
          },
        ),
      ),
    );
 ```

PR #1007, opened 2024-03-07
Add dependency_overrides to purchases_flutter_ui

vegaro:
I wasn't sure what was going on in #995 but there was some issues importing a newly added model. The problem is that we were depending on version ^6.0.0 of purchases_flutter. 

I made some changes to the pubspec.yaml, so now we use the local plugin. 

According to [the docs](https://dart.dev/tools/pub/dependencies), "As a result, if you publish a package to pub.dev, keep in mind that your package's dependency overrides are ignored by all users of your package.". So we should be fine to publish it like this.

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -25,7 +25,7 @@ files_with_version_number = {
     './ios/Classes/PurchasesFlutterPlugin.m' => ['return @"{x}"'],
     './android/build.gradle' => ['version \'{x}\''],
     './android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java' => ['PLUGIN_VERSION = "{x}"'],
-    './purchases_ui_flutter/pubspec.yaml' => ['version: {x}', 'purchases_flutter: ^{x}'],
+    './purchases_ui_flutter/pubspec.yaml' => ['version: {x}', 'purchases_flutter: {x}'],
     './purchases_ui_flutter/ios/purchases_ui_flutter.podspec' => ['s.version          = \'{x}\''],
     './purchases_ui_flutter/macos/purchases_ui_flutter.podspec' => ['s.version          = \'{x}\''],
     './purchases_ui_flutter/android/build.gradle' => ['version \'{x}\''],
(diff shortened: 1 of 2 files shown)


purchases_ui_flutter/pubspec.yaml
```diff
@@ -11,14 +11,15 @@ environment:
...
-  # We can't depend on exactly the same version
-  # because it won't be available when bumping the version
-  # until it's released.
-  purchases_flutter: ^6.0.0
+  purchases_flutter: ^6.22.0
   flutter:
     sdk: flutter
   plugin_platform_interface: ^2.0.2
 
+dependency_overrides:
```
tonidero, line 19: Maybe add a comment explaining this shouldn't affect published package?

tonidero (approved):

fastlane/Fastfile
```diff
@@ -25,7 +25,7 @@ files_with_version_number = {
     './ios/Classes/PurchasesFlutterPlugin.m' => ['return @"{x}"'],
     './android/build.gradle' => ['version \'{x}\''],
     './android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java' => ['PLUGIN_VERSION = "{x}"'],
-    './purchases_ui_flutter/pubspec.yaml' => ['version: {x}'],
+    './purchases_ui_flutter/pubspec.yaml' => ['version: {x}', 'purchases_flutter: ^{x}'],
```
tonidero, line 28: I guess if we are doing this, we could specify the specific package (which could probably be a good thing IMO...

tonidero (approved):

fastlane/Fastfile
vegaro: do you mean the specific package version? I got a warning when doing that:
```
* Your dependency on "purchases_flutter" should allow more than one version. For example:
  
  dependencies:
    purchases_flutter: ^6.22.0
  
  Constraints that are too tight will make it difficult for people to use your package
  along with other packages that also depend on "purchases_flutter".
```

should I add the specific version and ignore that?

tonidero: Oh sorry! Yeah I meant that... Basically so people don't accidentally use a mismatched version of purchases_flutter and purchases_ui_flutter

vegaro: perfect I did that. Makes sense. We will get a warning when publishing but it's fine

PR #1012, opened 2024-03-08
Add `PaywallFooterView` listeners

tonidero:
This PR adds initial listeners to the `PaywallFooterView`. Follow up from #995

### DIFF
--- a/purchases_ui_flutter/android/src/main/kotlin/com/revenuecat/purchases_ui_flutter/views/PaywallFooterView.kt
+++ b/purchases_ui_flutter/android/src/main/kotlin/com/revenuecat/purchases_ui_flutter/views/PaywallFooterView.kt
@@ -7,6 +7,7 @@ import android.view.Gravity
 import android.view.View
 import android.widget.FrameLayout
 import androidx.core.view.children
+import com.revenuecat.purchases.hybridcommon.ui.PaywallListenerWrapper
 import com.revenuecat.purchases.ui.revenuecatui.ExperimentalPreviewRevenueCatUIPurchasesAPI
 import com.revenuecat.purchases.ui.revenuecatui.views.PaywallFooterView as NativePaywallFooterView
 import io.flutter.plugin.common.BinaryMessenger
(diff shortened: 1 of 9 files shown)


purchases_ui_flutter/ios/Classes/PurchasesUiPaywallFooterView.swift
```diff
@@ -29,51 +34,76 @@ class PurchasesUiPaywallFooterViewFactory: NSObject, FlutterPlatformViewFactory
...
+        _paywallProxy?.delegate = self;
     }
 
     func view() -> UIView {
         return _view
     }
+}
 
-    @available(iOS 15.0, macOS 12.0, tvOS 15.0, *)
-    func paywallViewController(_ controller: PaywallViewController, didChangeSizeTo size: CGSize) {
```
tonidero, line 74: I hadn't actually setup the delegate in #999 so this was never called. After actually setting the delegate, this was returning a different value than the size as given by the view initially, which was making it draw incorrectly. I need to revisit whether this is needed and if so, why is it returning the wrong size...

purchases_ui_flutter/android/src/main/kotlin/com/revenuecat/purchases_ui_flutter/views/PaywallFooterView.kt
```diff
@@ -32,7 +33,7 @@ internal class PaywallFooterView(
     override fun dispose() {}
 
     init {
-        methodChannel = MethodChannel(messenger, "purchases_ui_flutter/PaywallFooterView/${id}")
+        methodChannel = MethodChannel(messenger, "com.revenuecat.purchasesui/PaywallFooterView/${id}")
```
tonidero, line 36: Changed the channel name to be consistent with the full screen version

purchases_ui_flutter/lib/views/paywall_view.dart
```diff
@@ -22,6 +24,16 @@ import 'package:purchases_flutter/models/store_transaction.dart';
...
+///
+/// [onPurchaseError] (Optional) Callback that gets called when a purchase
+/// fails.
+///
+/// [onRestoreCompleted] (Optional) Callback that gets called when a restore
+/// is completed. Note that this may get called even if no entitlements have
+/// been granted in case no relevant purchases were found.
+///
+/// [onRestoreError] (Optional) Callback that gets called when a restore
+/// fails.
```
vegaro, line 36: thanks for adding these!

purchases_ui_flutter/lib/views/paywall_view_method_handler.dart
```diff
@@ -0,0 +1,80 @@
+
+import 'package:flutter/services.dart';
+import 'package:purchases_flutter/models/customer_info_wrapper.dart';
+import 'package:purchases_flutter/models/package_wrapper.dart';
+import 'package:purchases_flutter/models/purchases_error.dart';
+import 'package:purchases_flutter/models/store_transaction.dart';
+
+class PaywallViewMethodHandler {
```
vegaro, line 8: so clean!

vegaro (approved):
Thanks for all that cleanup!

PR #1014, opened 2024-03-12
 Adds`Purchases.getCurrentOfferingForPlacement(placementID)` and `Purchases.syncAttributesAndOfferingsIfNeeded()`

joshdholtz:
## Motivation

Bump to PHC `10.1.0` and add support for new Targeting methods

## Description

⚠️ This removes `offeringIdentifier`/`presentedOfferingIdentifier` from the constructor for `Package`, `StoreProduct`, and `SubscriptionOption` which is _technically_ a breaking change

However, the property for `offeringIdentifier`/`presentedOfferingIdentifier` still exists (as an extension) which is how most developer should be using it. 

### Get Current Offering for Placement

Get an offering specific to the place of your paywall (eg: onboarding, settings, feature gate, etc). This is determined by Targeting.

```dart
final offering = await Purchases.getCurrentOfferingForPlacement("your-placement-identifier")
if (offering != null) {
    // Navigate to paywall
} else {
    // Do nothing or continue to another screen
}
```

### Sync Attributes and Offerings

A blocking call to sync attributes and fetch new offerings to to be used with Targeting that uses Custom Attributes.

```dart
// Using this result is optional - offerings will be cached for next Purchases.getOfferings() call
final offerings = await Purchases.syncAttributesAndOfferingsIfNeeded()
```

### DIFF
--- a/android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
+++ b/android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
@@ -457,12 +457,12 @@ public class PurchasesFlutterPlugin implements FlutterPlugin, MethodCallHandler,
     }
 
     private void purchaseProduct(final String productIdentifier,
-            final String type,
-            final String googleOldProductId,
-            @Nullable final Integer googleProrationMode,
-            @Nullable final Boolean googleIsPersonalizedPrice,
-            @Nullable final Map<String, Object> presentedOfferingContext,
-            final Result result) {
+                                 final String type,
+                                 final String googleOldProductId,
+                                 @Nullable final Integer googleProrationMode,
+                                 @Nullable final Boolean googleIsPersonalizedPrice,
+                                 @Nullable final Map<String, Object> presentedOfferingContext,
+                                 final Result result) {
         CommonKt.purchaseProduct(
                 getActivity(),
                 productIdentifier,
--- a/android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
+++ b/android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
@@ -17,6 +17,7 @@ import com.revenuecat.purchases.Store;
 import com.revenuecat.purchases.common.PlatformInfo;
 import com.revenuecat.purchases.hybridcommon.CommonKt;
 import com.revenuecat.purchases.hybridcommon.ErrorContainer;
+import com.revenuecat.purchases.hybridcommon.OnNullableResult;
 import com.revenuecat.purchases.hybridcommon.OnResult;
 import com.revenuecat.purchases.hybridcommon.OnResultAny;
 import com.revenuecat.purchases.hybridcommon.OnResultList;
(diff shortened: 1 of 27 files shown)


api_tester/lib/api_tests/models/package_wrapper_api_test.dart
```diff
@@ -26,16 +27,21 @@ class _PackageApiTest {
...
-      StoreProduct storeProduct, String offeringIdentifier) {
-    Package package =
-        Package(identifier, packageType, storeProduct, offeringIdentifier);
+  void _checkConstructor(
+      String identifier,
+      PackageType packageType,
+      StoreProduct storeProduct,
+      PresentedOfferingContext presentedOfferingContext) {
+    Package package = Package(
+        identifier, packageType, storeProduct, presentedOfferingContext);
```
joshdholtz, line 36: This replaces `String offeringIdentifier` with `PresentedOfferingContext presentedOfferingContext` in the constructor.  Technically a breaking change but I developer should not be using this one.

api_tester/lib/api_tests/models/store_product_wrapper_api_test.dart
```diff
@@ -22,7 +23,7 @@ class _StoreProductApiTest {
       List<StoreProductDiscount>? discounts,
       ProductCategory? productCategory,
       String? subscriptionPeriod,
-      String? presentedOfferingIdentifier) {
+      PresentedOfferingContext? presentedOfferingContext) {
```
joshdholtz, line 26: Same with this being a breaking change but shouldn't be used by developers

api_tester/lib/api_tests/models/subscription_option_wrapper_api_test.dart
```diff
@@ -23,7 +24,7 @@ class _SubscriptionOptionApiTest {
       PricingPhase? fullPricePhase,
       PricingPhase? freePhase,
       PricingPhase? introPhase,
-      String? presentedOfferingIdentifier) {
+      PresentedOfferingContext? presentedOfferingContext) {
```
joshdholtz, line 27: Same with this being a breaking change but shouldn't be used by developers

lib/models/package_wrapper.dart
```diff
@@ -61,17 +61,22 @@ class Package with _$Package {
...
 
   factory Package.fromJson(Map<String, dynamic> json) =>
       _$PackageFromJson(json);
 }
+
+extension ExtendedPackage on Package {
+  /// Offering this package belongs to.
+  @Deprecated('use presentedOfferingContext')
+  String get offeringIdentifier => presentedOfferingContext.offeringIdentifier;
+}
```
joshdholtz, line 82: Adding a getting property in an extension to prevent a breaking change. `fromJson` breaks if this getting property is directly on the class.

revenuecat_examples/purchase_tester/lib/src/upsell.dart
joshdholtz, line 1: This is kind of gross but I have plans to fully rework this purchase tester soon ™️

android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
```diff
@@ -434,12 +457,12 @@ public void onError(ErrorContainer errorContainer) {
     }
 
     private void purchaseProduct(final String productIdentifier,
-                                 final String type,
-                                 final String googleOldProductId,
-                                 @Nullable final Integer googleProrationMode,
-                                 @Nullable final Boolean googleIsPersonalizedPrice,
-                                 @Nullable final String presentedOfferingIdentifier,
-                                 final Result result) {
+            final String type,
```
tonidero, line 460: Hmm nitpick, but I think the indentation in these was correct following common java practices?

api_tester/lib/api_tests/models/package_wrapper_api_test.dart
tonidero: I was thinking we could potentially create a new constructor for the presentedOfferingContext to avoid the breaking change, but dart is annoying and only lets you have a single unnamed constructor, which should be the old one, making things complex... So I'm ok with this I think.

tonidero (approved):
Looks great!

android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
joshdholtz: Oh! Not sure how this happen 🤔 Undoing!

PR #1018, opened 2024-03-13
Fix footer sizing issues in iOS

tonidero:
The iOS footer was not sized appropriately. There were some issues:
- Delegate was not set correctly so we were not receiving the callback with the size from the native side.
- We get the size in pixel-independent units, contrary to android. But the initial time we were getting the size was in pixel units, which was confusing.

This fixes the issue so it has the correct height in all devices.

### DIFF
--- a/purchases_ui_flutter/lib/views/paywall_footer_view.dart
+++ b/purchases_ui_flutter/lib/views/paywall_footer_view.dart
@@ -49,11 +49,7 @@ class _PaywallFooterViewState extends State<PaywallFooterView> {
         top: 0,
         left: 0,
         right: 0,
-        // iOS is passing the size without including the top margin with the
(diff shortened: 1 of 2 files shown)


purchases_ui_flutter/lib/views/paywall_footer_view.dart
```diff
@@ -73,7 +69,10 @@ class _PaywallFooterViewState extends State<PaywallFooterView> {
   );
 
   void _updateHeight(double newHeight) {
-    final pixelRatio = MediaQuery.of(context).devicePixelRatio;
+    // In android we get pixels but in iOS we get pixel independent units.
```
vegaro, line 72: oh wow

vegaro (approved):

Issue #1038, opened 2024-04-10
macOS purchases ui flutter registrant error and paywall opening not supported


Albertbol:
- [X] I have updated Purchases SDK to the latest version
- [ X] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-flutter/blob/main/CONTRIBUTING.md)
- [ X] I have searched the [Community](https://community.revenuecat.com)
- [ X] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [ X] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-flutter/issues)

# ‼️ Required data ‼️ #
**Do not remove any of the steps from the template below. If a step is not applicable to your issue, please leave that step empty.**

There are a lot of things that can contribute to things not working. Having a very basic understanding of your environment will help us understand your issue faster!

## Environment ##
- [ ] Output of `flutter doctor`
- [ ] Version of `purchases-flutter`
- [ ] Testing device version e.g.: iOS 15.5, Android API 30, etc.
- [ ] How often the issue occurs- every one of your customers is impacted? Only in dev?
- [ ] [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue
- [ ] Steps to reproduce, with a description of expected vs. actual behavior
**Other information** (e.g. stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc.)

## Describe the bug ##
Firstly, opening paywall on macOS not working from purchases flutter ui, it sends error ios platform is required, why would you add macOS badge to package in that case?
Secondly, PaywallFooterView is throwing error:
```
"A UIKitView widget is trying to create a PlatformView with an unregistered type: < com.revenuecat.purchasesui/PaywallFooterView >"
"If you are the author of the PlatformView, make sure `registerViewFactory` is invoked.
See: https://docs.flutter.dev/development/platform-integration/platform-views#on-the-platform-side-1 for more details.
If you are not the author of the PlatformView, make sure to call `GeneratedPluginRegistrant.register`." 
```

I have following 2 line in my GeneratedPluginRegistrant.register:
```
PurchasesFlutterPlugin.register(with: registry.registrar(forPlugin: "PurchasesFlutterPlugin"))
PurchasesUiFlutterPlugin.register(with: registry.registrar(forPlugin: "PurchasesUiFlutterPlugin"))
```

vegaro:
Hi @Albertbol 

[...]

PR #1041, opened 2024-04-11
Remove macos platform support in purchases_ui_flutter

vegaro:
#1038 reports that MacOS is not working properly. The PaywallView and PaywallFooterView are being wrapped with UIKit, but they need to use AppKit, which is what's available in MacOS.

This PR removes support for now until we figure out what we need to do

### DIFF
--- a/purchases_ui_flutter/pubspec.yaml
+++ b/purchases_ui_flutter/pubspec.yaml
@@ -50,8 +50,6 @@ flutter:
         pluginClass: PurchasesUiFlutterPlugin
       ios:
         pluginClass: PurchasesUiFlutterPlugin
-      macos:
-        pluginClass: PurchasesUiFlutterPlugin
 
   # To add assets to your plugin package, add an assets section, like this:
   # assets:


tonidero (approved):

dsvf:
Hi there! True to [xkcd 1172](https://xkcd.com/1172/) _(every change breaks someones workflow)_, this PR broke _my_ workflow.

I develop and publish an iOS/Android app, but I do the flutter development as a Mac OS "native" app, since for regular coding and testing the workflow is smoother and less resource-intensive than using Simulator/an emulator. The paywall part of my app is long-since validated and shipped, and the Mac OS build never leaves my device - it is debug only. So that the purchase does not work on Mac OS is not important while I work on other features.

However, since this version, I can no longer even build a local debug build, since there is a dependency resolution problem - there is no latest purchases_ui version for Mac OS.

Now, I could theoretically remove RevCat for my debug builds and re-add them when building for release, but the previous way was perfect as-is. I have as of now pinned purchases-flutter to 6.26.0 to re-enable my Mac OS debug build situation, but if there was a way to secretly re-add this, that would be great

PR #1053, opened 2024-04-25
Crash on Amazon paywalls after purchasing

### LINKED ISSUE #1049: Purchase doesn't work properly with Amazon's Live App Testing

## Environment ##
```
[√] Flutter (Channel stable, 3.19.5, on Microsoft
    Windows [Version 10.0.22631.3447], locale ja-JP)    
[√] Windows Version (Installed version of Windows is    
    version 10 or higher)
[√] Android toolchain - develop for Android devices
    (Android SDK version 34.0.0)
[√] Chrome - develop for the web
[√] Visual Studio - develop Windows apps (Visual Studio 
    Community 2022 17.8.3)
[√] Android Studio (version 2022.3)
[√] VS Code (version 1.88.1)
[√] Connected device (3 available)
[√] Network resources

• No issues found!
```
```
  purchases_flutter: ^6.26.0
  purchases_ui_flutter: ^6.26.0
```
Fire OS 8.3.2.3
in Amazon's Live App Testing
- [x] [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue
**Other information** (e.g. stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc.)

## Describe the bug ##
I implemented in-app purchases using `PaywallFooterView` and tested purchases with Amazon's Live App Testing.
Then, even though the Amazon Appstore UI shows that the purchase is complete, `PaywallFooterView`'s `onPurchaseCompleted` does not seem to be triggered.

When checking with Logcat, the following error occurred.
(This error was displayed after the log `RevenueCatUI D Dismissing paywall after purchase`.)

```
[ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: type 'Null' is not a subtype of type 'String' in type cast
#0      _$$StoreTransactionImplFromJson (package:purchases_flutter/models/store_transaction.g.dart:11)
#1      new _$StoreTransactionImpl.fromJson (package:purchases_flutter/models/store_transaction.freezed.dart:215)
#2      _$StoreTransactionFromJson (package:purchases_flutter/models/store_transaction.freezed.dart:18)
#3      new StoreTransaction.fromJson (package:purchases_flutter/models/store_transaction.dart:51)
#4      PaywallViewMethodHandler._handleOnPurchaseCompleted (package:purchases_ui_flutter/views/paywall_view_method_handler.dart:62)
#5      PaywallViewMethodHandler.handleMethodCall (package:purchases_ui_flutter/views/paywall_view_method_handler.dart:32)
#6      InternalPaywallFooterView._buildListenerChannel.<anonymous closure> (package:purchases_ui_flutter/views/internal_paywall_footer_view.dart:99)
#7      MethodChannel._handleAsMethodCall (package:flutter/src/services/platform_channel.dart:571)
#8      MethodChannel.setMethodCallHandler.<anonymous closure> (package:flutter/src/services/platform_channel.dart:564)
#9      _DefaultBinaryMessenger.setMessageHandler.<anonymous closure> (package:flutter/src/services/binding.dart:603)
#10     _invoke2 (dart:ui/hooks.dart:344)
#11     _ChannelCallbackRecord.invoke (dart:ui/channel_buffers.dart:45)
#12     _Channel.push (dart:ui/channel_buffers.dart:135)
#13     ChannelBuffers.push (dart:ui/channel_buffers.dart:343)
#14     PlatformDispatcher._dispatchPlatformMessage (dart:ui/platform_dispatcher.dart:737)
#15     _dispatchPlatformMessage (dart:ui/hooks.dart:257)
```

However, afterwards restoring purchases (`onRestoreCompleted`) worked fine.

I tried resetting the in-app items from the Amazon Developer page multiple times, but the result did not change.

## Additional context ##
I don't know if it is directly related to this issue, but another app that was published on the Amazon Appstore was suspended due to in-app purchases not working properly.
The reason for the suspension was "Even after purchasing IAP, it does not reflect in the app as app asks for purchase again for using other pro features but displays message as IAP is already purchased."
This app used `purchases_flutter: ^6.5.1` and did not use `purchases_ui_flutter`.
I then submitted an update for the app with `purchases_flutter: ^6.24.0`, but it was rejected because "Observed that IAP Item is not delivered as expected to the user post purchase."
However, in the case of this app, Purchase worked fine on my actual device (Fire 7 2022, FireOS 8.3.2.2).
Furthermore, perhaps because the app has been suspended, I can no longer install this app that I submitted to the live app test after uninstalling it, so I am currently unable to test this app on my device.
Therefore, I decided to try using `purchases_ui_flutter` with another app, and the issue I reported this time was confirmed.

vegaro:
In `PurchaseCompleted` we are passing a Flutter's `StoreTransaction` in the callback. This is very confusing but:

- In Android `StoreTransaction` is what's returned after a successful purchase. This class has a nullable `orderId`
- In Android `Transaction` is the transactions in the CustomerInfo. This class has a non-nullable `transactionIdentifier`

The issue we are having is that in Flutter we only have one `StoreTransaction` with a non-nullable `transactionId`, and we are using it in both places, the `PurchaseCompleted` and `CustomerInfo`.

I see several ways of fixing this:

- Making transactionIdentifier nullable. I don’t like this. It’s breaking
- Making transactionIdentifier an empty string. Fixing it in next major
- Making transactionIdentifier an empty string, deprecating it and creating orderId which is nullable and the same as transactionIdentifier. We’ll have nullable orderIds in the customer info transactions where this object is also used. In Android we have two classes, one for CustomerInfo transactions (Transaction with non-nullable transactionId ) and StoreTransaction which is used for completed purchases
- Making a copy of StoreTransaction and name it CompletedPurchase , but with a nullable transactionIdentifier. Also breaking since we need to modify the PurchaseCompleted callback to receive this object. Making a new callback won’t work because the broken `PurchaseCompleted will keep crashing.

In this PR I am doing number 2, since it's the least breaking. This only affects apps using Amazon paywalls, which is not many.

Will fix #1049

### DIFF
--- a/lib/models/introductory_price.g.dart
+++ b/lib/models/introductory_price.g.dart
@@ -11,10 +11,10 @@ _$IntroductoryPriceImpl _$$IntroductoryPriceImplFromJson(Map json) =>
       (json['price'] as num).toDouble(),
       json['priceString'] as String,
       json['period'] as String,
-      json['cycles'] as int,
+      (json['cycles'] as num).toInt(),
       $enumDecode(_$PeriodUnitEnumMap, json['periodUnit'],
           unknownValue: PeriodUnit.unknown),
-      json['periodNumberOfUnits'] as int,
+      (json['periodNumberOfUnits'] as num).toInt(),
     );
 
 Map<String, dynamic> _$$IntroductoryPriceImplToJson(
--- a/lib/models/store_transaction.dart
+++ b/lib/models/store_transaction.dart
@@ -11,12 +11,12 @@ class StoreTransaction with _$StoreTransaction {
   /// Experimental. This factory method is subject to changes without
   /// a major release.
   const factory StoreTransaction.create(
-    /// RevenueCat Id associated to the transaction.
+    /// RevenueCat Id associated to the transaction. Empty for Amazon.
     // ignore: invalid_annotation_target
     @JsonKey(readValue: _readTransactionIdentifier)
     String transactionIdentifier,
 
-    /// Deprecated: Use transactionIdentifier instead.
+    /// Deprecated: Use transactionIdentifier instead. Empty for Amazon.
     @Deprecated('Use transactionIdentifier instead.')
     // ignore: invalid_annotation_target
     @JsonKey(readValue: _readTransactionIdentifier)
--- a/lib/models/store_transaction.dart
+++ b/lib/models/store_transaction.dart
@@ -12,12 +12,14 @@ class StoreTransaction with _$StoreTransaction {
   /// a major release.
   const factory StoreTransaction.create(
     /// RevenueCat Id associated to the transaction.
+    // ignore: invalid_annotation_target
+    @JsonKey(readValue: _readTransactionIdentifier)
     String transactionIdentifier,
 
     /// Deprecated: Use transactionIdentifier instead.
     @Deprecated('Use transactionIdentifier instead.')
     // ignore: invalid_annotation_target
-    @JsonKey(readValue: _readRevenueCatIdentifier)
+    @JsonKey(readValue: _readTransactionIdentifier)
         String revenueCatIdentifier,
 
     /// Product Id associated with the transaction.
@@ -51,5 +53,6 @@ class StoreTransaction with _$StoreTransaction {
       _$StoreTransactionFromJson(json);
 }
 
-Object? _readRevenueCatIdentifier(Map json, String key) =>
-    json['transactionIdentifier'];
+// reads json or defaults to '' if not there (Amazon)
+Object _readTransactionIdentifier(Map json, String key) =>
+    json['transactionIdentifier'] ?? '';
--- a/lib/models/store_transaction.freezed.dart
+++ b/lib/models/store_transaction.freezed.dart
@@ -21,11 +21,13 @@ StoreTransaction _$StoreTransactionFromJson(Map<String, dynamic> json) {
 /// @nodoc
 mixin _$StoreTransaction {
   /// RevenueCat Id associated to the transaction.
+// ignore: invalid_annotation_target
+  @JsonKey(readValue: _readTransactionIdentifier)
   String get transactionIdentifier => throw _privateConstructorUsedError;
 
   /// Deprecated: Use transactionIdentifier instead.
   @Deprecated('Use transactionIdentifier instead.')
-  @JsonKey(readValue: _readRevenueCatIdentifier)
+  @JsonKey(readValue: _readTransactionIdentifier)
   String get revenueCatIdentifier => throw _privateConstructorUsedError;
 
   /// Product Id associated with the transaction.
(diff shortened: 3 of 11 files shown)


lib/models/store_transaction.dart
```diff
@@ -25,11 +27,17 @@ class StoreTransaction with _$StoreTransaction {
 
     /// Purchase date of the transaction in ISO 8601 format.
     String purchaseDate,
+  {
+    /// Deprecated: Use transactionIdentifier instead.
+    @Deprecated('Use orderId instead.')
+    /// RevenueCat Id associated to the transaction.
+    @Default('') String transactionIdentifier,
+  }
```
vegaro, line 35: This needs to be made optional in order to have a default, which it actually makes it a breaking change

lib/models/introductory_price.g.dart
```diff
@@ -11,10 +11,10 @@ _$IntroductoryPriceImpl _$$IntroductoryPriceImplFromJson(Map json) =>
       (json['price'] as num).toDouble(),
       json['priceString'] as String,
       json['period'] as String,
-      json['cycles'] as int,
+      (json['cycles'] as num).toInt(),
       $enumDecode(_$PeriodUnitEnumMap, json['periodUnit'],
           unknownValue: PeriodUnit.unknown),
-      json['periodNumberOfUnits'] as int,
+      (json['periodNumberOfUnits'] as num).toInt(),
```
tonidero, line 17: New freezed version?

lib/models/store_transaction.dart
```diff
@@ -11,13 +11,15 @@ class StoreTransaction with _$StoreTransaction {
   /// Experimental. This factory method is subject to changes without
   /// a major release.
   const factory StoreTransaction.create(
-    /// RevenueCat Id associated to the transaction.
+    /// RevenueCat Id associated to the transaction. Empty for Amazon.
```
tonidero, line 14: Maybe this to be more specific? Since it will not be empty for Amazon inapps I believe.
```suggestion
    /// RevenueCat Id associated to the transaction. Empty for Amazon subscriptions.
```
I believe this will only be empty there but lmk if I'm wrong.

tonidero (approved):
Thanks for taking care of this!

lib/models/introductory_price.g.dart
vegaro: yes, the build was failing because CI was using the newer version of freezed so I had to rerun the build of files. I could split it into another PR, but I think it's fine since they are autogenerated?

lib/models/store_transaction.freezed.dart
```diff
@@ -20,12 +20,14 @@ StoreTransaction _$StoreTransactionFromJson(Map<String, dynamic> json) {
 
 /// @nodoc
 mixin _$StoreTransaction {
-  /// RevenueCat Id associated to the transaction.
+  /// RevenueCat Id associated to the transaction. Empty for Amazon.
+// ignore: invalid_annotation_target
```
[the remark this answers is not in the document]
vegaro: this file is auto generated and for I think the non-autogenerated one has correct indentation right?

lib/models/introductory_price.g.dart
tonidero: Yeah I think it's fine

lib/models/store_transaction.freezed.dart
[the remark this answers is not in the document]
tonidero: Ohh right, ignore me 😅

lib/models/store_transaction.dart
tonidero: Ohh ok... I was thinking we would be using the `Transaction` model for inapps but maybe that's not correct for paywalls... I think this is ok then

PR #1062, opened 2024-05-06
Migration to applying Gradle plugins with the declarative plugins block in sample

vegaro:
I kept getting 

```
You are applying Flutter's main Gradle plugin imperatively using the apply script method, which is deprecated and will be removed in a future release. Migrate to applying Gradle plugins with the declarative plugins block: https://flutter.dev/go/flutter-gradle-plugin-apply
```

when running the sample. I created a new plugin project and made some changes over to our purchase tester app

### DIFF
--- a/revenuecat_examples/purchase_tester/android/app/src/main/java/com/revenuecat/purchases_sample/MainActivity.kt
+++ b/revenuecat_examples/purchase_tester/android/app/src/main/java/com/revenuecat/purchases_sample/MainActivity.kt
@@ -1,0 +1,5 @@
+package com.revenuecat.purchases_sample
+
+import io.flutter.embedding.android.FlutterFragmentActivity
+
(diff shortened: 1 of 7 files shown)


revenuecat_examples/purchase_tester/android/app/src/main/java/com/revenuecat/purchases_sample/MainActivity.kt
```diff
@@ -0,0 +1,5 @@
+package com.revenuecat.purchases_sample
+
+import io.flutter.embedding.android.FlutterFragmentActivity
+
+class MainActivity: FlutterFragmentActivity()
```
vegaro, line 5: migrated this to kotlin since that's the default when creating a new app

tonidero (approved):
Nice! Thanks for doing this!

PR #1098, opened 2024-06-13
Removes api key from integration test

JayShortway:
As the title says. Saw this while working on something else. 

Reading [.circleci/config.yml](https://github.com/RevenueCat/purchases-flutter/blob/main/.circleci/config.yml) it seems that the `replace-api-key` command should pick this up and replace it properly, but let me know if I'm overlooking something! 

I'm also happy to close this if this is not a problem, or if the fix is not as easy as this PR.


tonidero (approved):
Ups yeah, this is good. Thanks!

PR #1104, opened 2024-06-24
Adds .aiexclude

JayShortway:
Adds `.aiexclude` to exclude files with sensitive info from being accessed by Gemini in Android Studio. More info can be read [here](https://developer.android.com/studio/preview/gemini/aiexclude).

### DIFF
--- a/.aiexclude
+++ b/.aiexclude
@@ -1,0 +1,3 @@
+fastlane/.env
+local.properties
+constant.dart


tonidero (approved):

vegaro (approved):

.aiexclude
```diff
@@ -0,0 +1,3 @@
+fastlane/.env
+local.properties
+constant.dart
```
vegaro, line 3: does this look for this file in any subdirectory?

vegaro: I believe this should be

```
**/**/local.properties
**/**/constant.dart
```

JayShortway: If I understand the [example](https://developer.android.com/studio/preview/gemini/aiexclude#examples) correctly, it does: 
>The pattern KEYS blocks all files called "KEYS" with no file extension at or below the directory that contains the .aiexclude file.

vegaro: Makes sense! I just saw a big caution in https://developer.android.com/studio/preview/gemini/aiexclude#how-write with the same. 

I was confused by the main screenshot in that link because they do use `*.txt`. Sorry for the noise!

JayShortway: No worries, good to double check!

PR #1123, opened 2024-07-25
[PurchaseTester] Refactor upsell screen to display all offerings

tonidero:
I did this change to test some things and I think it's more useful. Basically moved the current UI under a collapsible section, so we can display all offerings, instead of only the `current` offering as we're doing now.

| Before | After |
| -- | -- |
| image | image |


JayShortway (approved):
Very useful indeed!

PR #1129, opened 2024-08-05
v7.0.0

fire-at-will:
This PR contains the necessary changes to consume Purchases Hybrid Commons 13.0.0.

### OS Version Requirements
Bumps required OS versions to:
- iOS 13.0
- tvOS 13.0
- watchOS 6.2
- macOS 10.15
- Android: SDK 21 (Android 5.0)

### API Additions
- Introduces `StoreKitVersion` enum
- Reworks the `PurchasesAreCompletedBy` enum to be a sealed class requiring a `storeKitVersion` when set to `PurchasesAreCompletedByMyApp`

### PurchasesConfiguration Adjustments
- Replaced the deprecated `useStoreKit2IfAvailable` bool with `storeKitVersion`, a StoreKitVersion enum
- Removed the deprecated `observerMode` bool

### Purchases Function Removals/Deprecations
- Removes the deprecated `setAutomaticAppleSearchAdsAttributionCollection` since it was removed in PHC 13.0.0
- Removes `setFinishTransactions()`
- Deprecates `syncObserverModeAmazonPurchase`, replaces it with `syncAmazonPurchase`

### Other
- Updates the MagicWeather/Purchase Tester apps
- Updates the automated tests and added a few new ones for `Purchases.configure()`
- Includes a v7 migration guide


ios/Classes/PurchasesFlutterPlugin.m
```diff
@@ -326,6 +314,24 @@ - (void)syncPurchasesWithResult:(FlutterResult)result {
...
 
+- (void)recordPurchaseForProductID:(NSString*)productId
+                            result:(FlutterResult)result {
+    if (@available(iOS 15.0, macOS 12.0, *)) {
+        [RCCommonFunctionality recordPurchaseForProductID:productId
+                                               completion:^(NSDictionary<NSString *,id> * _Nullable responseDictionary, RCErrorContainer * _Nullable error) {
+            if (error) {
+                [self rejectWithResult:result error:error];
+            } else {
+                result(responseDictionary);
```
fire-at-will, line 325: I think this is being returned correctly - let me know if I'm missing something!

lib/models/upgrade_info.dart
```diff
@@ -37,3 +37,21 @@ enum ProrationMode {
   /// plus remaining prorated time from the old plan.
   immediateAndChargeFullPrice
 }
+
+// ignore_for_file: deprecated_member_use_from_same_package
+extension ProrationModeExtension on ProrationMode {
```
fire-at-will, line 42: @tonidero Are these values correct? Just want to double check 😅

tonidero: These values look good! But I just noticed we are missing `deferred` as well, which should have a value of `6`. Thanks for double-checking! I see I forgot to add that in my base PR...

tonidero: Oh wait! There is also the `GoogleProductChange` that has `deferred` with the correct value. This one is deprecated, so no need to add `deferred` here. Nvm then! These look good :)

tonidero (approved):
Looks great!!

lib/models/purchases_completed_by.dart
```diff
@@ -0,0 +1,44 @@
...
+  const PurchasesAreCompletedByRevenueCat();
+}
+
+class PurchasesAreCompletedByMyApp extends PurchasesAreCompletedBy {
+  final StoreKitVersion storeKitVersion;
+
+  PurchasesAreCompletedByMyApp({
+    required this.storeKitVersion,
+  });
+}
```
MarkVillacampa, line 44: nice! 🙌

MarkVillacampa (approved):

MarkVillacampa:
🚀

PR #1137, opened 2024-08-08
Update iOS deployment target in readme

fire-at-will:
This PR updates the iOS deployment target note in the README to reflect the fact that the minimum iOS deployment target is iOS13 as of version 7.0.0.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@
 To use this plugin, add `purchases_flutter` as a [dependency in your pubspec.yaml file](https://flutter.io/platform-plugins/).
 
 ### Requirements
-*purchases_flutter* requires Xcode 14.0+ and minimum targets iOS 11.0+.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -24,7 +24,7 @@
 To use this plugin, add `purchases_flutter` as a [dependency in your pubspec.yaml file](https://flutter.io/platform-plugins/).
 
 ### Requirements
-*purchases_flutter* requires Xcode 14.0+ and minimum targets iOS 11.0+.
+*purchases_flutter* requires Xcode 14.0+ and minimum targets iOS 13.0+.
```
tonidero, line 27: Should we add the android min version as well while we're at it?

tonidero (approved):

README.md
fire-at-will: Sure, might as well!

fire-at-will: Added! :D

PR #1179, opened 2024-10-08
Fix area insets in PaywallView

vegaro:
This PR basically does the same we do in React Native to handle the safe area in the PaywallView

| Before | After |
|--------|--------|
| simulator_screenshot_C744866E-38EE-4081-973F-FEB94C903675 | simulator_screenshot_57C3C890-24D0-4EF9-8F7C-2866AC82A3DA |


JayShortway (approved):
Perfect is the enemy of good. Nice improvement!

vegaro:
Good news, I managed to fix them all

aboedo (approved):

tonidero (approved):
Nice! Many thanks for fixing this! 🙏

PR #1227, opened 2024-12-17
Support Web Redemption Links

tonidero:
This adds support for Web Redemption Links in our Flutter SDK.


lib/models/web_purchase_redemption_result.freezed.dart
```diff
@@ -0,0 +1,875 @@
+// coverage:ignore-file
```
tonidero, line 1: This is generated code by freezed, no need to review it!

revenuecat_examples/purchase_tester/pubspec.yaml
```diff
@@ -13,6 +13,8 @@ dependencies:
   # Use with the CupertinoIcons class for iOS style icons.
   cupertino_icons: ^1.0.4
 
+  app_links: ^6.3.3
```
tonidero, line 16: I added this library to read the deep links. Seemed one of the simpler ones.

tonidero (commented):
Sorry for the big diff, but more than half is actually generated code, so it should hopefully not be as bad as it looks 🙏

vegaro (approved):
looks good! I think this solution will just work fine with any deeplinking library

PR #1237, opened 2025-01-07
[Beta] Flutter Web support

tonidero:
This PR adds initial support for RevenueCat's Web SDK in the Flutter plugin. The implementation:

- Adds web platform support through `purchases_flutter_web.dart`
- Integrates purchases-js-hybrid-mappings (v0.0.7) for web functionality
- Implements core RevenueCat features for web:
  - Configuration and setup
  - Customer info retrieval
  - Offerings and packages
  - User management (login, logout)
  - Purchasing a package

### Breaking change
- Uses `package:web` and `dart:js_interop` for JavaScript interop with RevenueCat's Web SDK. This requires a min dart SDK version of 3.4.0, which is lower than our current min version.

### Unsupported Features
The following features are currently stubbed:
- Direct product operations (getProducts, purchaseProduct)
- Platform-specific features (iOS/Android specific methods)
- Subscriber attributes
- Device identifiers

This is based on @Jethro87's work in #1236


marcucio:
Are you looking for any help with this?

Yasinilhan61:
Any update on this?

davoutuk:
When is this due for release?

marcucio:
is this ready to use? Should I start testing?

ysfsl:
> is this ready to use? Should I start testing?

not ready yet, i have tested and login doesn't work

Jethro87:
Hi all, we don't have a timeline for when this will be available, but my team and I will continue working on it.

marcucio:
I have a flutter web app launching in the next week so I will have to branch and finish what I need myself for my use case, what is the current status of this so I know at what point I'm starting at?

lib/purchases_flutter.dart
```diff
@@ -1230,6 +1230,16 @@ class Purchases {
...
+
+  /// Web-only: Loads and caches some optional data in the Purchases SDK.
+  /// Currently only fetches branding information.
+  ///
+  /// You can call this method after configuring the SDK to speed up the first call to purchase.
+  /// This method has no effect on non-web platforms.
+  static Future<void> preload() async {
+    if (!kIsWeb) return;
+    await _channel.invokeMethod('preload');
+  }
```
tonidero, line 1242: I'm going to remove this for now, and add it later if needed... Seems like a very generic API that could be confused with it doing something else. We can add it back in the future if needed.

pubspec.yaml
```diff
@@ -13,8 +13,11 @@ environment:
 dependencies:
   flutter:
     sdk: flutter
+  flutter_web_plugins:
+    sdk: flutter
   freezed_annotation: ^2.0.1
   json_annotation: ^4.8.0
+  js: ^0.6.7
```
tonidero, line 20: This is deprecated now... But will keep using it for now and move to the `web` plugin in the future.

lib/web/purchases_flutter_web.dart
```diff
@@ -0,0 +1,872 @@
...
+      registrar,
+    );
+
+    final instance = PurchasesFlutterPlugin();
+    channel.setMethodCallHandler(instance.handleMethodCall);
+
+    void injectScript() {
+      final script = html.ScriptElement()
+        ..src =
+            'https://unpkg.com/@revenuecat/purchases-js-hybrid-mappings@0.0.7-alpha.1/dist/index.umd.js'
```
tonidero, line 46: This is the new library we are using that contains the mappings. This allowed to remove a lot of logic from here.

lib/web/purchases_flutter_web.dart
```diff
@@ -0,0 +1,872 @@
...
+        case 'presentCodeRedemptionSheet':
+          return _presentCodeRedemptionSheet();
+        case 'purchaseSubscriptionOption':
+          return _purchaseSubscriptionOption(call.arguments);
+        case 'purchaseDiscountedProduct':
+          return _purchaseDiscountedProduct(call.arguments);
+        case 'purchaseDiscountedPackage':
+          return _purchaseDiscountedPackage(call.arguments);
+        case 'getPromotionalOffer':
+          return _getPromotionalOffer(call.arguments);
```
tonidero, line 142: I'm thinking about joining all the cases that should no-op or return an unsupported error together here, instead of creating separate methods. I think that would improve visibility IMO.

pubspec.yaml
tonidero: You're correct! noticed that after my initial comment. Need to revisit this before shipping this in any stable form. Thanks for checking!!

revenuecat_examples/purchase_tester/macos/Flutter/GeneratedPluginRegistrant.swift
```diff
@@ -7,10 +7,8 @@ import Foundation
 
 import app_links
 import purchases_flutter
-import purchases_ui_flutter
 
 func RegisterGeneratedPlugins(registry: FlutterPluginRegistry) {
   AppLinksMacosPlugin.register(with: registry.registrar(forPlugin: "AppLinksMacosPlugin"))
   PurchasesFlutterPlugin.register(with: registry.registrar(forPlugin: "PurchasesFlutterPlugin"))
-  PurchasesUiFlutterPlugin.register(with: registry.registrar(forPlugin: "PurchasesUiFlutterPlugin"))
```
JayShortway, line 15: Is this intended?

lib/web/purchases_flutter_web.dart
```diff
@@ -0,0 +1,440 @@
...
+  static Completer<void>? _initCompleter;
+
+  static void _injectScriptIfNeeded() {
+    if (_initCompleter != null) {
+      return;
+    }
+    _initCompleter = Completer<void>();
+
+    // TODO: Host library ourselves.
+    const library = 'https://unpkg.com/@revenuecat/purchases-js-hybrid-mappings@0.0.7-alpha.2/dist/index.umd.js';
```
JayShortway, line 26: Should we define this URL somewhere where we can more easily automatically update it (either the entire URL or only the version number)?

lib/web/purchases_flutter_web.dart
```diff
@@ -0,0 +1,440 @@
...
+      'packageIdentifier': packageIdentifier,
+      'presentedOfferingContext': arguments['presentedOfferingContext'],
+    };
+    return await _getMapFromInstanceMethod('purchasePackage', [options]);
+  }
+
+  Future<Map<String, dynamic>> _restorePurchases() async =>
+    // Web SDK doesn't support restore purchases, but returns current customer info
+    // to match behavior with other platforms
+    await _getCustomerInfo();
```
JayShortway, line 242: Should we log anything here explaining that we're not actually restoring anything?

lib/web/purchases_flutter_web.dart
```diff
@@ -0,0 +1,440 @@
...
+
+  Future<void> _close() async {
+    // Web SDK's close() just unsets the instance
+    // TODO
+    // js.context['PurchasesHybridMappings'] = null;
```
JayShortway, line 247: Why is this a TODO? Does the commented-out code not work?

JayShortway (approved):
Amazing job! ❤️ Just some questions/suggestions.

lib/web/purchases_flutter_web.dart
tonidero: Good point... I don't think it's a great practice to log things on web (that's why in web, it's Silent by default), but maybe it's worth it 👍

revenuecat_examples/purchase_tester/macos/Flutter/GeneratedPluginRegistrant.swift
tonidero: That was automatically removed by flutter... However, I haven't tested the UI functions yet on this branch. Will confirm whether it's needed before actually reverting this change. Good catch!

lib/web/purchases_flutter_web.dart
JayShortway: Alright, you're the best judge of that! (If it's silent by default, we're not logging unless the dev wants to, so maybe that's okay?)

lib/web/purchases_flutter_web.dart
```diff
@@ -0,0 +1,440 @@
...
+    final script = HTMLScriptElement()
+      ..type = 'text/javascript'
+      ..text = '''
+        window.after_rc_load_callback = async (callback) => {
+          callback(await import("$library"));
+        };
+      ''';
+
+    assert(document.head != null);
+    document.head!.append(script);
```
n-bernat, line 37: `assert` followed by bang operator (postfix null assertion; `!`) would _in theory_ result in different debug and release mode errors (`_AssertionError` vs `_TypeError`). I would probably check for nullability and explicitly throw a relevant error, e.g.:

```suggestion
    final head = document.head;
    if (head == null) {
      throw StateError(
        "The current document doesn't have a head element which is required to insert a script.",
      );
    }
    
    head.append(script);
```

---
_that being said_, even if a developer removes the `head` section, Flutter would add it back automatically, so I guess it's nullable only to be compatible with the Web IDL standard 🤷‍♂️

pubspec.yaml
```diff
@@ -13,8 +13,11 @@ environment:
 dependencies:
   flutter:
     sdk: flutter
+  flutter_web_plugins:
+    sdk: flutter
   freezed_annotation: ^2.0.1
   json_annotation: ^4.8.0
+  web: ^1.1.1
```
n-bernat, line 20: Since you depend on `package:web`, the [min. required SDK version is 3.4.0](https://github.com/dart-lang/web/blob/222164b5010cb0324d4132666f1d43750ffc6448/web/pubspec.yaml#L7), AND since the min. Dart version is 3.4.0, you can bump both `dart` and `flutter` versions at the top of this file ([corresponding Flutter version for this Dart version is 3.22.0](https://docs.flutter.dev/release/archive)).

```yaml
environment:
  sdk: ">=3.4.0 <4.0.0"
  flutter: ">=3.22.0"
```

lib/web/purchases_flutter_web.dart
```diff
@@ -0,0 +1,440 @@
...
+        return 'DEBUG';
+      case 'info':
+        return 'INFO';
+      case 'warn':
+        return 'WARN';
+      case 'error':
+        return 'ERROR';
+      default:
+        return 'SILENT';
+    }
```
n-bernat, line 195: Once you bump SDK to >= 3.0 you get support for switch expressions, e.g.:
```suggestion
    return switch (logLevel.toLowerCase()) {
      'verbose' => 'VERBOSE',
      'debug' => 'DEBUG',
      'info' => 'INFO',
      'warn' => 'WARN',
      'error' => 'ERROR',
      _ => 'SILENT',
    };
```

lib/web/purchases_flutter_web.dart
```diff
@@ -0,0 +1,440 @@
...
+      } else if (arg == null) {
+        jsArgs.add(null);
+      } else {
+        throw ArgumentError(
+          'Unsupported argument type: ${arg.runtimeType}',
+        );
+      }
+    }
+    return jsArgs;
+  }
```
n-bernat, line 409: Dart 3 also has support for pattern matching, e.g.:

```suggestion
  List<JSAny?> _processArgs(List<dynamic> args) {
    return <JSAny?>[
      for (final arg in args)
        switch (arg) {
          final Map<String, dynamic> arg => arg.jsify(),
          final List arg => arg.jsify(),
          final String arg => arg.toJS,
          final int arg => arg.toJS,
          final bool arg => arg.toJS,
          null => null,
          _ => throw ArgumentError(
              'Unsupported argument type: ${arg.runtimeType}',
            ),
        },
    ];
  }
```

pubspec.lock
```diff
@@ -216,6 +216,11 @@ packages:
     description: flutter
     source: sdk
     version: "0.0.0"
+  flutter_web_plugins:
```
n-bernat, line 219: [This whole file should probably be `.gitignore`d](https://dart.dev/tools/pub/private-files#the-rules) and not in the repo 🙈

lib/web/purchases_flutter_web.dart
```diff
@@ -0,0 +1,440 @@
...
+
+  PlatformException _processError(dynamic error) {
+    if (error is JSObject && error.has('code')) {
+      final errorMap = _convertJsRecordToMap(error);
+      final code = errorMap['code'];
+      final message = errorMap['message'] as String?;
+      final underlyingErrorMessage = errorMap['underlyingErrorMessage'] as String?;
+      final finalMessage = '$message. $underlyingErrorMessage';
```
n-bernat, line 417: That `as String?` has no effect since you interpolate them anyway, for consistency with `code` you can drop it

```suggestion
      final message = errorMap['message'];
      final underlyingErrorMessage = errorMap['underlyingErrorMessage'];
      final finalMessage = '$message. $underlyingErrorMessage';
```

lib/web/purchases_flutter_web.dart
tonidero: That makes sense. Thanks for the suggestion!

lib/web/purchases_flutter_web.dart
tonidero: makes sense, thanks for the suggestion!

pubspec.yaml
tonidero: That's a very good point. Thanks for bringing it up!

lib/web/purchases_flutter_web.dart
```diff
@@ -0,0 +1,428 @@
...
+  Future<bool> _isSandbox() async {
+    final purchases = _getInstance();
+    return purchases.callMethod('isSandbox'.toJS).dartify() as bool;
+  }
+
+  Future<Map<String, dynamic>> _checkTrialOrIntroductoryPriceEligibility(
+    dynamic arguments,
+  ) async {
+    // Web SDK doesn't support trial/intro eligibility checking
+    // Return unknown status for all products
```
ajpallares, line 261: I wonder, would it be preferable to fail the function in this case? I don't have a clear opinion but I *think* it makes more sense

ajpallares (commented):
Amazing work Toni! While I don't think I have enough knowledge to give approval, I do have 2 small comments. Thanks for doing this!

pubspec.lock
```diff
@@ -297,10 +302,10 @@ packages:
     dependency: transitive
     description:
       name: js
-      sha256: c1b2e9b5ea78c45e1a0788d29606ba27dc5f71f019f32ca5140f61ef071838cf
+      sha256: f2c445dce49627136094980615a031419f7f3eb393237e4ecd97ac15dea343f3
       url: "https://pub.dev"
     source: hosted
-    version: "0.7.1"
+    version: "0.6.7"
```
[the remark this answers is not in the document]
tonidero: Actually we ended up removing the `js` dependency, so this shouldn't be here anymore. Planning to remove this entire file from being checked in though as suggested above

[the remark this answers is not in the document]
ajpallares: Oh! Makes sense. Thanks for the explanation!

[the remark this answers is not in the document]
n-bernat: fwiw, `package:js` will still be in your local `pubspec.lock` because `build_runner` (code generator) from `dev_dependencies` depends on it; since it's a dev dependency it won't be imported for devs who depend on this package though

(you can check dependency graph with `flutter pub deps`)

lib/web/purchases_flutter_web.dart
tonidero: Yeah we did it this way because it's how it behaves in Android so I think it's better to be consistent there, but I could totally see your point as well. Lmk if you feel strongly about it though!

JayShortway (approved):
Very nice! 🚀 💪

tonidero:
I'm tackling a possible approach to separate web functionality to its own package in #1326. It seems to be working so far. This would avoid the breaking change introduced by using the `web` package.

n-bernat:
@tonidero I just switched to the `main` branch of this repo, and checked whether everything works with different SDK versions, and there are lots of version conflicts:

<details>
<summary>⭐ SDK/Framework support ⭐</summary>

- Flutter 3.19 / Dart 3.3
  - `purchases_flutter` ❌ (version solving failed)
  - `purchases_ui_flutter` ✅
  - `magic_weather` ✅
  - `purchase_tester` ✅

- Flutter 3.16 / Dart 3.2
  - `purchases_flutter` ❌
  - `purchases_ui_flutter` ✅
  - `magic_weather` ✅
  - `purchase_tester` ❌

- Flutter 3.10 / Dart 3.0
  - `purchases_flutter` ❌
  - `purchases_ui_flutter` ❌
  - `magic_weather` ✅
  - `purchase_tester` ❌

- Flutter 3.0 / Dart 2.17
  - `purchases_flutter` ❌
  - `purchases_ui_flutter` ❌
  - `magic_weather` ❌
  - `purchase_tester` ❌

</details>

Packages can't resolve dependencies due to conflicts with `dev_dependencies`, but example apps also don't work within their declared version constraints either. 

In my packages, I usually check whether the build succeeds on: 
- the current Flutter stable channel, 
- the current Flutter beta channel,
- the oldest supported Dart/Flutter version. 

It may be something worth adding to this repo as well.

I don't know what kind of support you have to provide for older versions, but in the perfect world something that I would recommend in this case is to use this opportunity (adding support for a new platform) to do some minor cleanup, sync versions between subpackages, bump everything, set up CI to ensure that everything always works for min SDK constraints, and maybe separately, somewhere in the future use [Pub Workspaces](https://dart.dev/tools/pub/workspaces) to set SDK version in a single place, so this issue never happens again.

As for splitting web support into a separate package, it sounds kinda annoying for end users and messes with the whole point of a multiplatform framework, not a big fan of it tbh

Great job overall! Thanks a lot for working on it 🚀

tonidero:
To everyone following this PR, we've just released version [9.0.0-beta.1](https://github.com/RevenueCat/purchases-flutter/releases/tag/9.0.0-beta.1) with web support! Please check it out and let us know your feedback! It would be greatly appreciated. To use it, you should just create a new [Web Billing](https://www.revenuecat.com/docs/web/web-billing/overview) app and use the provided API key when using web targets.

PR #1268, opened 2025-02-17
feat: Introduce CustomerCenter iOS layer

facumenzella:
This PR introduces the CustomerCenter iOS layer. For now, we're only exposing a `present` function, but we might need to make RCCustomerCenterViewController public anyways for other platforms.


ajpallares (commented):
Changes look good to me. But I'll leave approval for someone who actually knows Flutter 😄

purchases_ui_flutter/lib/purchases_ui_flutter.dart
```diff
@@ -50,6 +49,13 @@ class RevenueCatUI {
     return _parseStringToResult(result);
   }
 
+  static presentCustomerCenter() async {
```
tonidero, line 52: I'm not totally sure if it would pick the correct type automatically, but should this be
```suggestion
  static Future<void> presentCustomerCenter() async {
```

purchases_ui_flutter/ios/Classes/PurchasesUiFlutterPlugin.swift
```diff
@@ -129,6 +153,22 @@ public class PurchasesUiFlutterPlugin: NSObject, FlutterPlugin {
         #endif
     }
 
+        private func presentCustomerCenter(
```
tonidero, line 156: Hmm the indentation here seems wrong?

purchases_ui_flutter/ios/Classes/PurchasesUiFlutterPlugin.swift
```diff
@@ -129,6 +153,22 @@ public class PurchasesUiFlutterPlugin: NSObject, FlutterPlugin {
...
+            _ result: @escaping FlutterResult
```
tonidero, line 157: Seems the result is never called... I don't remember what the behavior is in this case? As long as it doesn't get stuck waiting for a response it should be ok but we should confirm.

purchases_ui_flutter/ios/Classes/Helpers/UIView+.swift
tonidero, line 1: Hmm the name of the file seems wrong? Do we need to add `UIView+Extensions` or something like that?

tonidero (commented):
Just some comments. We also need to merge PHC before this so we can point to a stable version of PHC.

purchases_ui_flutter/lib/purchases_ui_flutter.dart
facumenzella: Surprised it worked without it 🤔

purchases_ui_flutter/ios/Classes/PurchasesUiFlutterPlugin.swift
facumenzella: I took this from the paywalls example that does not return either

```swift
    private func presentPaywall(
        _ result: @escaping FlutterResult,
        requiredEntitlementIdentifier: String?,
        offeringIdentifier: String?,
        displayCloseButton: Bool?
    ) {
#if os(iOS)
        if #available(iOS 15.0, *) {
            let displayCloseButton = displayCloseButton ?? false

            var options: [String:Any] = [
                PaywallProxy.PaywallOptionsKeys.displayCloseButton: displayCloseButton,
                // This is needed for: https://github.com/RevenueCat/purchases-flutter/issues/1023
                PaywallProxy.PaywallOptionsKeys.shouldBlockTouchEvents: true
            ]

            if let offeringIdentifier {
                options[PaywallProxy.PaywallOptionsKeys.offeringIdentifier] = offeringIdentifier
            }

            if let requiredEntitlementIdentifier {
                options[PaywallProxy.PaywallOptionsKeys.requiredEntitlementIdentifier] = requiredEntitlementIdentifier

                self.paywallProxy.presentPaywallIfNeeded(
                    options: options,
                    paywallResultHandler: result
                )
            } else {
                self.paywallProxy.presentPaywall(
                    options: options,
                    paywallResultHandler: result
                )
            }
        } else {
            NSLog("Presenting paywall requires iOS 15+")
        }
#else
        NSLog("Presenting paywall requires iOS")
#endif
    }
```

facumenzella: Would something like [this](https://github.com/RevenueCat/purchases-flutter/pull/1268/commits/4158a5bd6d21d1b15de3b7aa46aefb959d5d8f59) make sense? This way we call `result(nil)` for now, till we add proper support more for completion events; and return an error if you can't present it.

tonidero: Yeah, for the errors looks good. For the success case, I guess we are returning before the customer center is dismissed... TBH, considering the CustomerCenter doesn't return anything, I don't think it's that bad, but it does work differently than how the paywall works and I don't think it's ideal... I would say we should fix before shipping this, but can be done in follow-up PRs.

vegaro: That's what I did in Android as well, I think it makes sense to only complete the promise on dismiss, it makes more sense for a developer in my opinion

purchases_ui_flutter/lib/purchases_ui_flutter.dart
```diff
@@ -49,6 +49,13 @@ class RevenueCatUI {
     return _parseStringToResult(result);
   }
 
+  static Future<void> presentCustomerCenter() async {
+    // handling result will be implemented in upcoming PRs
```
vegaro, line 53: I actually don't think we need to handle anything, since the actions will be sent as a callback not as a result

vegaro (approved):

Issue #1273, opened 2025-02-20
Please add Proguard rules to purchases_flutter just like how you added them for purchases_ui_flutter in v8.5.1


Unity2Blender:
For purchases_flutter along with Flutterflow I face an issue where the app crashes on Android when shareFile from share_plus package is called, This issue is happening when Revenuecat's purchases_flutter package is also used.

I request you to please add the same Proguard line and changes in purchases_flutter package and rollout a new update before this weekend so that I can ship new updates for my app (Flutterflow users can't edit Proguard rules, It would be much quicker if you guys add Proguard configuration with purchases_flutter)

_Originally posted by @Unity2Blender in https://github.com/RevenueCat/purchases-flutter/pull/1270#issuecomment-2672622892_

RCGitBot:
👀 We've just linked this [issue to our internal tracker](https://revenuecat.zendesk.com/agent/tickets/53739) and notified the team. Thank you for reporting, we're checking this out!

Unity2Blender:
https://github.com/RevenueCat/purchases-flutter/blob/main/android%2Fproguard-rules.pro

In the Proguard rules file of purchases_flutter, If you could add 

`-keep class org.xmlpull.v1.** { *; }`

This would solve a part of my problem with conflicts of SDK as I can't edit this file from Flutterflow.

Unity2Blender:
@vegaro , I have referenced the stack trace of issue in Release 8.5.1's discussion, Kindly refer to it.

Adding `-keep class org.xmlpull.v1.** { *; }` to purchases_flutter/android/proguard-rules.pro may solve this issue for Flutterflow and other Flutter based app developers.

facumenzella:
Hey @Unity2Blender we've already merged this https://github.com/RevenueCat/purchases-flutter/pull/1274
It will be included in the next release :)

Unity2Blender:
Thanks for your contribution.

PR #1270, opened 2025-02-19
Add proguard-rules.pro to `purchases-ui-flutter`

vegaro:
Fixes:

- https://github.com/RevenueCat/purchases-flutter/issues/1245
- https://github.com/RevenueCat/purchases-flutter/issues/1269
- https://community.revenuecat.com/general-questions-7/app-crashing-when-paywall-is-open-on-flutter-5826?postid=20041#post20041

This is not really caused by us, as we don't have a direct dependency on `org.xmlpull` but we keep getting reports so we have decided to add it to our own proguard rules. This library is bundled by the OS, and if the library doesn't exist it won't have any effect, but if one of the dependencies of the app adding our plugin has a direct dependency and overrides it, it will the classes.

I think the key here (and why this doesn’t happen for purchases-android) is that in flutter apps, packages are compiled by the developers as Android modules. So if one of the flutter packages has a dependency on a different version of `org.xmlpull.v1` than the one the system bundles (and the one we compile purchases-android against), the app will be compiled using that version. Since R8 optimizes each app individually, it will analyze `org.xmlpull.v1` and strip out “unused” code from that direct dependency, and our code crashes because it was compiled to depend on those references that have been stripped.

PR #1274, opened 2025-02-21
chore: Update proguard-rules.pro like flutter_ui

facumenzella:
A follow-up of https://github.com/RevenueCat/purchases-flutter/pull/1270

Based on https://github.com/RevenueCat/purchases-flutter/issues/1273, we're adding the same rule like we did for `flutter_ui`


android/proguard-rules.pro
```diff
@@ -1,4 +1,9 @@
...
 -keep class com.amazon.** {*;}
 -keepattributes *Annotation*
--keep class androidx.lifecycle.DefaultLifecycleObserver
\ No newline at end of file
+-keep class androidx.lifecycle.DefaultLifecycleObserver
+
+# The org.xmlpull package is part of the Android framework, so the classes are always available.
+# However, some apps add it to their classpath, either explicitly or transitively. When
+# this happens, it becomes susceptible to be shrunk. The following rule avoids that.
+-keep class org.xmlpull.v1.** { *; }
```
tonidero, line 9: Since we're adding it here, we should remove it from the ui package. It would be unnecessary there, considering this is a dependency of that package.

vegaro (approved):

Unity2Blender:
Hi, thanks for updating the main branch but when would this release be available on Pub Dev, like v8.5.2?

I will be waiting for your pub dev release.

PR #1281, opened 2025-02-25
Increase android related versions in purchase_tester

vegaro:
- I got a compilation error because integration_test requires a higher compileSdk and ndkVersion 
- I also got a warning that kotlin 1.7.10 was going to be deprecated so I checked out a brand new project and updated the versions in settings.gradle to match that


JayShortway (approved):
Nice! We can always eagerly update `compileSdk` and AGP btw.

ajpallares (approved):

tonidero (approved):

PR #1289, opened 2025-03-10
Fix new Flutter version test issues

tonidero:
Looks like in the last version of flutter, api tests started failing with an error:
```
You are applying Flutter's app_plugin_loader Gradle plugin imperatively using the apply script method, 
which is not possible anymore. Migrate to applying Gradle plugins 
with the declarative plugins block: https://flutter.dev/to/flutter-gradle-plugin-apply
```

This PR updates the gradle files to new ones obtained from a new sample app, so we are actually using `gradle.kts` and the latest syntax. Considering it's only for the api-tests, this seems like a low-risk change.


.circleci/config.yml
```diff
@@ -230,6 +230,7 @@ jobs:
     description: "Build Android Flutter integration tests"
     docker:
       - image: ghcr.io/cirruslabs/flutter:stable
+    resource_class: large
```
tonidero, line 233: Not sure what changed since only a patch version of flutter was released but android integration tests were consistently failing due to the daemon process getting killed. This + increasing the memory for that seems to have fixed the issue.

vegaro (approved):

PR #1307, opened 2025-04-01
Fix API Tests

fire-at-will:
This PR fixes the broken API test job by increasing the resource class of the machine it runs on in CI.


tonidero (approved):
Is it enough with a large machine? In any case I think this is good!

fire-at-will:
The job passes, so I'll take it for now 🤪

PR #1326, opened 2025-04-24
Flutter Web: Separate functionality to separate package

tonidero:
This will separate the logic to support flutter web to a separate package. If it's present, it will be detected by Flutter and support web automatically. This allows us to avoid any breaking changes in our main `purchases_flutter` package


purchases_flutter_web/README.md
```diff
@@ -0,0 +1,7 @@
+## Installation
+To use this plugin, add `purchases_flutter` and `purchases_flutter_web` as a [dependency in your pubspec.yaml file](https://flutter.io/platform-plugins/).
```
JayShortway, line 2: Just curious: could we, at some point in the future, make `purchases_flutter_web` a transitive dependency of `purchases_flutter`?

JayShortway (approved):
Great job! 💪

purchases_flutter_web/README.md
tonidero: Possibly yeah, we could include it as a dependency of `purchases_flutter` and that would automatically add web support to the main SDK 👍

JayShortway: Nice!

n-bernat: Yep, you can turn it into something called [endorsed plugin](https://docs.flutter.dev/packages-and-plugins/developing-packages#endorsed-federated-plugin).

purchases_flutter_web/lib/purchases_flutter_web.dart
```diff
@@ -2,15 +2,14 @@ import 'dart:async';
...
 import 'package:flutter_web_plugins/flutter_web_plugins.dart';
 import 'package:web/web.dart';
 
-import '../purchases_flutter.dart';
-
 class PurchasesFlutterPlugin {
-  static final _unknownErrorCode = '${PurchasesErrorCode.unknownError.index}';
-  static final _configurationErrorCode = '${PurchasesErrorCode.configurationError.index}';
+  static final _unknownErrorCode = '0';
+  static final _configurationErrorCode = '23';
```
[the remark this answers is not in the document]
n-bernat: Yes, it's usually done by creating a package with platform interface (see: [url_launcher_web](https://github.com/flutter/packages/blob/main/packages/url_launcher/url_launcher_web/pubspec.yaml#L24)), but that whole federated plugin system is an overkill in my opinion, so that's why I was sceptical of separating web support it into a package

tonidero:
After talking through it with the team, we are planning to make a beta of the next major with web included in the `purchases_flutter` plugin instead of separating to a different plugin. We can gather feedback and reconsider before stable release if needed. Will close this for now.

PR #1328, opened 2025-04-25
Bump min supported versions and add tests for min SDK compatibility

tonidero:
This bumps the min Flutter/Dart versions to the actual mins we support in the SDKs and adds some tests to verify we are still compatible with the min SDK version we say we are compatible.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -223,7 +223,7 @@ jobs:
     steps:
         - checkout
         - flutter/install_sdk_and_pub:
-            version: 3.3.0
+            version: 3.16.3
         - flutter-get-dependencies:
             project: purchases_ui_flutter
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -203,6 +203,32 @@ jobs:
           name: Run tests for purchases_ui_flutter
           command: cd purchases_ui_flutter && flutter test
 
+  test_main_sdk_min_version_compatibility:
+    description: "Test min version compatibility for purchases_flutter SDK"
+    docker:
+      - image: ghcr.io/cirruslabs/flutter:3.0.0
+    steps:
+      - checkout
+      - build-flutter-project
+      - flutter-get-dependencies:
+          project: revenuecat_examples/MagicWeather
+      - flutter-get-dependencies:
+          project: revenuecat_examples/purchase_tester
+      - flutter-get-dependencies:
+          project: api_tester
+      - flutter-get-dependencies:
+          project: purchases_ui_flutter
+
+  test_ui_sdk_min_version_compatibility:
+    description: "Tet UI min version compatibility for purchases_ui_flutter SDK"
+    docker:
+      - image: ghcr.io/cirruslabs/flutter:3.0.0
+    steps:
+        - checkout
+        - build-flutter-project
+        - flutter-get-dependencies:
+            project: purchases_ui_flutter
+
   api_test:
     description: "Run API tests for Flutter"
     docker:
@@ -412,6 +438,8 @@ workflows:
       - ios-integration-test
       - macos-integration-test
       - android-integration-test-build
+      - test_main_sdk_min_version_compatibility
+      - test_ui_sdk_min_version_compatibility
 
   deploy:
     when:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,13 +19,19 @@ aliases:
         ignore: /.*/
       branches:
         only: main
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -19,13 +19,19 @@ aliases:
...
+      tag: 2024.04.1
 
 orbs:
-  android: circleci/android@0.2.1
+  android: circleci/android@2.5.0
   gcp-cli: circleci/gcp-cli@2.1.0
   swissknife: roopakv/swissknife@0.65.0
   revenuecat: revenuecat/sdks-common-config@2.1.0
   macos: circleci/macos@2.3.2
+  flutter: circleci/flutter@2.1.0
```
tonidero, line 34: I think we probably should move more of our jobs to use this orb... But for now, I wanted to minimize the changes.

pubspec.yaml
```diff
@@ -7,8 +7,8 @@ issue_tracker: https://github.com/RevenueCat/purchases-flutter/issues
 documentation: https://docs.revenuecat.com/docs
 
 environment:
-  sdk: ">=2.15.0 <4.0.0"
-  flutter: ">=1.10.0"
+  sdk: ">=2.18.0 <4.0.0"
+  flutter: ">=3.3.0"
```
JayShortway, line 11: Just double checking: this is not "breaking" as it was already broken, right?

.circleci/config.yml
```diff
@@ -203,6 +209,24 @@ jobs:
...
+
+  test_ui_sdk_min_version_compatibility:
+    description: "Tet UI min version compatibility for purchases_ui_flutter SDK"
+    <<: *android-machine-emulator
+    steps:
+        - checkout
+        - flutter/install_sdk_and_pub:
+            version: 3.16.3
+        - flutter-get-dependencies:
+            project: purchases_ui_flutter
```
JayShortway, line 228: Curious why we need an emulator for these jobs?

.circleci/config.yml
```diff
@@ -203,6 +209,24 @@ jobs:
           name: Run tests for purchases_ui_flutter
           command: cd purchases_ui_flutter && flutter test
 
+  test_main_sdk_min_version_compatibility:
+    description: "Test min version compatibility for purchases_flutter SDK"
+    <<: *android-machine-emulator
+    steps:
+      - checkout
+      - flutter/install_sdk_and_pub:
+          version: 3.3.0
```
JayShortway, line 218: Not necessarily for this PR, but it would be nice to define the min version (for both purchases and revenuecatui) in a central place. It's now defined twice, here and in the `pubspec.yaml`.

.circleci/config.yml
```diff
@@ -19,13 +19,19 @@ aliases:
         ignore: /.*/
       branches:
         only: main
+  android-machine-emulator: &android-machine-emulator
+    executor:
+      name: android/android-machine
+      resource-class: large
+      tag: 2024.04.1
```
JayShortway, line 26: Any reason not to use the latest, or at least a newer version?

JayShortway (approved):
Thanks a lot for taking care of this! 💪 Just some questions.

pubspec.yaml
tonidero: That's the point yeah... Until now it would fail while resolving dependencies for this package. Now, it would just fail slightly earlier. To verify, you can see that our current dependency, `json_annotation` to 4.8.0 requires dart 2.18.0:

image

That dart version corresponds to flutter 3.3.0, so it wouldn't work anyway before these versions.

.circleci/config.yml
tonidero: We don't really need an emulator, that machine name was poorly named... (a copy paste from Android). I will rename the machine so there is no confusion. Thanks!

.circleci/config.yml
tonidero: Not really, should work on the latest as well. Will bump 👍

.circleci/config.yml
JayShortway: In that case I think we should use the `android/android-docker` executor, instead of the `android/android-machine` one.

.circleci/config.yml
tonidero: Agreed... I'm not sure off the top of my head to get this from those different places, but we can check in future PRs.

pubspec.yaml
JayShortway: Thanks for confirming, makes sense!

PR #1335, opened 2025-05-04
Add `storefront` API to obtain the country of the store account

tonidero:
This adds a new async API: `Purchases.storefront`, which will return storefront information about the store account used. Currently only the country code. It can be used like this:
```
Storefront? storefront = await Purchases.storefront;
```

Will be null when the storefront information can't be obtained (For example, in simulators without a storekit config file).

### DIFF
--- a/lib/models/storefront.dart
+++ b/lib/models/storefront.dart
@@ -1,0 +1,13 @@
+/// Contains the information about the current store account.
+class Storefront {
+  /// Country code of the current store account.
+  String countryCode;
+
+  Storefront({
+    required this.countryCode,
+  });
+
+  factory Storefront.fromJson(Map<String, dynamic> json) => Storefront(
+    countryCode: json['countryCode'] as String,
+  );
+}
--- a/lib/purchases_flutter.dart
+++ b/lib/purchases_flutter.dart
@@ -562,6 +562,15 @@ class Purchases {
   static Future<String> get appUserID async =>
       await _channel.invokeMethod('getAppUserID') as String;
 
+  /// Gets the current storefront for the store account.
+  static Future<Storefront?> get storefront async {
+    final storefrontJson = await _channel.invokeMethod('getStorefront');
+    if (storefrontJson == null) {
+      return null;
+    }
+    return Storefront.fromJson(Map<String, dynamic>.from(storefrontJson));
+  }
+
   /// This function will logIn the current user with an appUserID.
   /// Typically this would be used after logging in a user to identify them without
   /// calling configure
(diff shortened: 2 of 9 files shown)


lib/models/storefront.dart
```diff
@@ -0,0 +1,13 @@
+/// Contains the information about the current store account.
+class Storefront {
```
tonidero, line 2: Please note how this is not using `freezed`. We wanted to start moving models off freezed, and this seemed like an easy start. Note that this means we won't have some convenience functionality we got with freezed like equals/toString for free... But I think that's ok and we can add that later if needed.

lib/purchases_flutter.dart
```diff
@@ -562,6 +562,15 @@ class Purchases {
   static Future<String> get appUserID async =>
       await _channel.invokeMethod('getAppUserID') as String;
 
+  /// Gets the current storefront for the store account.
+  static Future<Storefront?> get storefront async {
+    final storefrontJson = await _channel.invokeMethod('getStorefront');
```
ajpallares, line 567: Question (since I'm not experienced on this): are all map/dictionary objects in the native transformed into JSON by Flutter?

ajpallares (approved):
Makes sense! 🚀

lib/models/storefront.dart
aboedo: yeah equals and toString aren't exactly advanced features lol we can add if we ever see the need

aboedo (approved):

lib/purchases_flutter.dart
tonidero: To be fair, it's not actually json (as in not a json string or an object representing a Json), but just a simple Dart Map yeah. The MethodChannel performs the transformation for those basic types. You can see those here: https://docs.flutter.dev/platform-integration/platform-channels#codec

ajpallares: Oh, I see. Then, in the line below
```
return Storefront.fromJson(Map<String, dynamic>.from(storefrontJson));
```
if `storefrontJson` is already a Dart Map, why is
```
Map<String, dynamic>.from(storefrontJson)
```
needed?

(Sorry for the newbie questions 🙏)

tonidero: Not a newbie question at all!! So what we get from the channel is actually a “Map<Object?, Object?>”. We could try to parse this directly but since we know the type of this map, it’s easier to convert it to the expected types like this before actually parsing the map I think. There might be better ways to do this but this was the easiest IMO

ajpallares: Ohh this makes sense. Thanks so much for explaining!

PR #1357, opened 2025-06-06
Update CHANGELOG with 9.0.0 betas

tonidero:
Updating the CHANGELOG to include 9.0.0-beta.1 and 9.0.0-beta.2


tonidero:
Bump @RevenueCat/coresdk

ajpallares (approved):
Looks good! One question though, I'm guessing that the versions in the changelog appear sorted by creation date. Is it ok to leave `9.0.0-beta.1` between `8.7.4` and `8.7.5` then?

tonidero:
> Is it ok to leave 9.0.0-beta.1 between 8.7.4 and 8.7.5 then?

It's when it was released I believe... So I think that's fine yeah... As in, it included changes from 8.7.4 but not 8.7.5+

PR #1331, opened 2025-05-01
Chore: Migrate to Freezed 3.0

davidpryor:
NOTE: This bumps the min SDK version to 3.0 but this seems to be planned already for the 9.0.0.

Description: Upgraded the freezed package to 3.0 and performed code migrations
Related Issue: #1295 

Summary:
- bumped package sdk min version to 3.0 for sealed classes and class modifiers features
- bumped example sdk min version to 3.0
- upgraded packages
- changed `WebPurchaseRedemptionResult` redirected constructors to be public so they may be used with `switch` expressions
- migrated from using `when` method to `switch` expressions
- fixed tests

Issue #1295, opened 2025-03-16
upgrade to latest freezed_annotation


g-apparence:
- [X] I have updated Purchases SDK to the latest version
- [X] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-flutter/blob/main/CONTRIBUTING.md)
- [x] I have searched the [Community](https://community.revenuecat.com)
- [x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [x] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-flutter/issues)

## Describe the bug ##
freezed_annotation is currently on 2.X version while the latest version is now 3.X
Please upgrade to latest version of freezed (as many of us are using it)

RCGitBot:
👀 We've just linked this [issue to our internal tracker](https://revenuecat.zendesk.com/agent/tickets/55127) and notified the team. Thank you for reporting, we're checking this out!

fatherOfLegends:
Need this as well.

hantrungkien:
Please help me to update:

```
Because purchases_flutter >=4.0.0-beta.1 depends on freezed_annotation ^2.0.1 and mm_app depends on freezed_annotation ^3.0.0,
  purchases_flutter >=4.0.0-beta.1 is forbidden.
So, because mm_app depends on purchases_flutter ^8.6.1, version solving failed.
```

eggnstone:
As a temporary workaround, overriding works fine for me:
```
dependency_overrides:
  freezed: ^3.0.4
  freezed_annotation: ^3.0.0
```

davidpryor:
Could I grab this issue?

With the upgrade to the sdk minimum version (2.18.0 -> 3.4.0) in the 9.0.0 beta1 branch, this upgrade is now possible.

If so, I have a local branch working off of main. I am not sure if you'd rather me branch from elsewhere since 9.0.0 will have the required sdk version bump.

I have upgraded freezed, bumped json_annotation one minor version, upgraded the freezed models to use the abstract/sealed keywords, and migrated to using switch expressions instead of the "when" method.

normidar:
What is the milestone for this issue?

appregion:
Any chance this could be prioritized soon? We're blocked by the freezed version.

yelkamel:
I did the override thank you, but guys please update the freeze package in your project it's necessary to use riverpod.

HarmonyV:
[...]

PR #1365, opened 2025-06-17
[EXTERNAL] Chore: Migrate to Freezed 3.0 (#1331) by @davidpryor

tonidero:
Description: Upgraded the freezed package to 3.0 and performed code migrations
Related Issue: #1295 

Summary:
- bumped package sdk min version to 3.0 for sealed classes and class modifiers features
- bumped example sdk min version to 3.0
- upgraded packages
- changed `WebPurchaseRedemptionResult` redirected constructors to be public so they may be used with `switch` expressions
- migrated from using `when` method to `switch` expressions
- fixed tests

Contributed by @davidpryor


pubspec.yaml
```diff
@@ -15,7 +15,7 @@ dependencies:
     sdk: flutter
   flutter_web_plugins:
     sdk: flutter
-  freezed_annotation: ^2.0.1
+  freezed_annotation: ^3.0.0
```
tonidero, line 18: ~~Note that I removed the dev_dependency of freezed but left the dependency. This is to avoid forcing us to support dart 3.6+ which is a much recent version of the Dart SDK. `freezed_annotation` does support Dart 3.0+. We are planning to slowly move away from freezed, so we can remove this dependency as well in the future.~~ Doing this in a followup PR

lib/models/customer_info_wrapper.dart
```diff
@@ -9,7 +9,7 @@ part 'customer_info_wrapper.g.dart';
 @freezed
 
 /// Class containing all information regarding the customer
-class CustomerInfo with _$CustomerInfo {
+abstract class CustomerInfo with _$CustomerInfo {
```
JayShortway, line 12: I suppose all the classes made abstract is needed for Freezed. These models aren't really designed to be extensible. Do you think it's possible (and desired) to make them final/closed again in https://github.com/RevenueCat/purchases-flutter/pull/1366? Maybe with a smart Search and Replace? 😄

pubspec.yaml
JayShortway, line 1: (This comment is unrelated to this file.)

I wonder what the changes in this PR do for our [generated SDK reference](https://pub.dev/documentation/purchases_flutter/latest/index.html). Might be good to regenerate and double check, as a form of api tests?

lib/models/web_purchase_redemption_result.dart
```diff
@@ -7,25 +6,36 @@ import 'purchases_error.dart';
 part 'web_purchase_redemption_result.freezed.dart';
 
 @freezed
-abstract class WebPurchaseRedemptionResult with _$WebPurchaseRedemptionResult {
+sealed class WebPurchaseRedemptionResult with _$WebPurchaseRedemptionResult {
```
JayShortway, line 9: I like `sealed` no matter the language 😄

purchases_ui_flutter/pubspec.yaml
```diff
@@ -7,8 +7,8 @@ documentation: https://docs.revenuecat.com/docs
 
 environment:
   # TODO: Check constraints are ok
```
JayShortway, line 9: Haha how long has this been here?

lib/models/customer_info_wrapper.dart
tonidero: It might be tricky... the way freezed models are generated, these are basically interfaces, with just the factory methods, and the actual implementations are done as part of the mixins generated by freezed. We can't mark it final with the provided implementation, so we would need to rewrite the models now... I think it might be fine to keep these models as interfaces? It might indeed make moving away from freezed models later in the future, but I don't think it would be impossible either...

JayShortway: Yea alright, I think we can keep it as interfaces now indeed, if the alternative is rewriting the models. We might do that in the _next_ major haha. Thanks for the context!

pubspec.yaml
tonidero: I did a quick test generating these locally, and with [dart_apitool](https://pub.dev/packages/dart_apitool) (thanks for the suggestion!) and the changes are mostly expected, after the follow-up PRs #1366  and #1368

tonidero:
Will start merging this + #1366 + #1368 into the 9.0.0 dev branch

PR #1366, opened 2025-06-17
Remove freezed dev dependency

tonidero:
This reverts some changes done in #1365 in order to support freezed and we remove freezed so we don't have issues causing us to have to increase our min sdk version. Note that this also means we can't regenerate freezed models anymore, but we plan to move away from freezed moving forward.


lib/models/customer_info_wrapper.dart
```diff
@@ -1,4 +1,5 @@
-import 'package:freezed_annotation/freezed_annotation.dart';
+
+import '../freezed_annotation/freezed_annotation.dart';
```
tonidero: Note I had to copy this runtime dependency in our SDK... I thought of keeping that dependency for now, but wanted to avoid incompatibilities with the freezed models down the line, considering we won't be updating those in the future. 

Plan is to slowly move away our models from freezed.

pubspec.yaml
```diff
@@ -7,17 +7,18 @@ issue_tracker: https://github.com/RevenueCat/purchases-flutter/issues
...
 dependencies:
   flutter:
     sdk: flutter
   flutter_web_plugins:
     sdk: flutter
-  freezed_annotation: ^3.0.0
   json_annotation: ^4.8.0
   web: ^1.1.1
+  collection: ^1.15.0
+  meta: ^1.7.0
```
tonidero: These are dependencies of freezed_annotation. After removing them, we still need them at runtime. We can remove them later on, as we remove freezed from our models.

pubspec.yaml
```diff
@@ -7,26 +7,24 @@ issue_tracker: https://github.com/RevenueCat/purchases-flutter/issues
...
+  sdk: ">=3.4.0 <4.0.0"
+  flutter: ">=3.22.0"
 
 dependencies:
   flutter:
     sdk: flutter
   flutter_web_plugins:
     sdk: flutter
-  freezed_annotation: ^3.0.0
+  freezed_annotation: 3.0.0
```
tonidero, line 18: Note I'm locking the dependency to the current version to avoid changes inadvertently causing issues, now that we won't be regenerating the models anymore.

pubspec.yaml
```diff
@@ -7,26 +7,24 @@ issue_tracker: https://github.com/RevenueCat/purchases-flutter/issues
...
   web: ^1.1.1
 
 dev_dependencies:
-  build_runner: ^2.1.10
   flutter_driver:
     sdk: flutter
   flutter_lints: ^2.0.1
   flutter_test:
     sdk: flutter
-  freezed: ^3.0.0
```
tonidero, line 29: With this, we won't be able to regenerate the models anymore.

pubspec.yaml
```diff
@@ -7,26 +7,24 @@ issue_tracker: https://github.com/RevenueCat/purchases-flutter/issues
 documentation: https://docs.revenuecat.com/docs
 
 environment:
-  sdk: ">=3.6.0 <4.0.0"
-  flutter: ">=3.27.0"
+  sdk: ">=3.4.0 <4.0.0"
```
JayShortway: Just curious what this is based on? (I noticed we came from 2.x.)

pubspec.yaml
JayShortway: Locking the version won't cause conflicts with other dependencies in people's apps, that might depend on a different version?

JayShortway (approved):
Good stuff! Only questions.

pubspec.yaml
tonidero: The increase to 3.4.0 min was to use the web package (was done previously on the 9.0.0-dev branch). The freezed update required going to 3.6.0, but since we're removing it, we can keep it at 3.4.0.

pubspec.yaml
tonidero: Not a problem here anymore since we are actually removing this dependency in #1368

PR #1368, opened 2025-06-18
Remove freezed generated code in favor of manually written mappings

tonidero:
This follows up the work in #1365 and #1366. But this ends up removing freezed completely in favor of writing our own mapping implementation. A few points to note:
- While I tried to minimize breaking changes, there are a couple that I wasn't sure how to fix.
- I had to add the Equatable package to be able to keep an equals implementation same way we had it with freezed. We could also write these manually (equals + hashcode) but felt like a lot of boilerplate, and this package seems simpler... Lmk if you think otherwise!


api_tester/lib/api_tests/models/customer_info_wrapper_api_test.dart
```diff
@@ -7,10 +7,6 @@ class _CustomerInfoApiTest {
     CustomerInfo customerInfo = CustomerInfo.fromJson(json);
   }
 
-  void _checkToJson(CustomerInfo customerInfo) {
-    Map<String, dynamic> json = customerInfo.toJson();
```
tonidero, line 11: This is one of the breaking changes. I didn't think we need to keep the `toJson` method in all our models.

api_tester/lib/api_tests/models/store_transaction_api_test.dart
```diff
@@ -8,29 +8,22 @@ class _StoreTransactionApiTest {
...
-    Map<String, dynamic> json = transaction.toJson();
-  }
-
   void _checkConstructor(
       String transactionIdentifier,
-      String revenueCatIdentifier,
       String productIdentifier,
       String purchaseDate) {
     StoreTransaction transaction =
-        StoreTransaction(revenueCatIdentifier, productIdentifier, purchaseDate);
```
tonidero, line 21: This is another breaking change... I could have kept `revenueCatIdentifier` which was deprecated, but would mean additional mapping code, and I felt it would be ok to remove it now.

api_tester/lib/api_tests/models/store_transaction_api_test.dart
```diff
@@ -8,29 +8,22 @@ class _StoreTransactionApiTest {
...
+        StoreTransaction(transactionIdentifier, productIdentifier, purchaseDate);
     StoreTransaction transaction2 =
-        StoreTransaction.create(
```
tonidero, line 23: Another breaking change. Seems we had a `create` function for StoreTransaction, but I think it's better to keep the same patter we have everywhere else.

api_tester/lib/api_tests/models/web_purchase_redemption_result_api_test.dart
```diff
@@ -14,20 +14,20 @@ class _WebPurchaseRedemptionResultApiTest {
       String obfuscatedEmail,
   ) {
     WebPurchaseRedemptionResult result =
-        WebPurchaseRedemptionResult.success(customerInfo: customerInfo);
-    result = WebPurchaseRedemptionResult.error(error: error);
-    result = const WebPurchaseRedemptionResult.purchaseBelongsToOtherUser();
-    result = const WebPurchaseRedemptionResult.invalidToken();
-    result = WebPurchaseRedemptionResult.expired(obfuscatedEmail: obfuscatedEmail);
+        WebRedemptionSuccess(customerInfo: customerInfo);
```
tonidero: This is the biggest breaking change. I wasn't sure how to namespace the results to WebPurchaseRedemptionResult, so I change the name of each value of the sealed class so it's at least more concrete... In any case, this was already a breaking change from #1365

build.yaml
```diff
@@ -1,13 +0,0 @@
-targets:
```
tonidero, line 1: This could have been removed in the previous PR, but I removed it here.

lib/models/customer_info_wrapper.dart
```diff
@@ -1,67 +1,106 @@
...
+    this.allExpirationDates,
+    this.requestDate, {
+    this.latestExpirationDate,
+    this.originalPurchaseDate,
+    this.originalApplicationVersion,
+    this.managementURL,
+  });
+
+  factory CustomerInfo.fromJson(Map<String, dynamic> json) => CustomerInfo(
+      EntitlementInfos.fromJson(Map<String, dynamic>.from(json['entitlements'])),
```
tonidero, line 75: I guess it's error prone to have all this mappings done manually... I would like to write more tests for all these mappings before merging this.

lib/models/customer_info_wrapper.dart
```diff
@@ -1,67 +1,106 @@
...
+      Map<String, dynamic>.from(json['allExpirationDates']).map((k, v) => MapEntry(k, v as String?)),
+      json['requestDate'] as String,
+      latestExpirationDate: json['latestExpirationDate'] as String?,
+      originalPurchaseDate: json['originalPurchaseDate'] as String?,
+      originalApplicationVersion: json['originalApplicationVersion'] as String?,
+      managementURL: json['managementURL'] as String?,
+    );
+
+  @override
+  List<Object?> get props => [
```
tonidero, line 91: This is how `Equatable` knows what fields to check. This means we need to remember to add new fields here, which I don't like... But I think the same problem exists by writing custom equals + hashCode and this removes a lot of the boilerplate

JayShortway: Oh that's unfortunate. I would have hoped it would analyze the model's AST and figure out what its fields are at build time. But if there's no tool like that, I agree that this is a bit better than writing custom equals and hashCode functions.

tonidero: Yeah... it doesn't have any sort of code generation, like freezed does, which requires extra steps to generate that code. Not ideal, but I'm not sure of good alternatives right now.

api_tester/lib/api_tests/models/customer_info_wrapper_api_test.dart
JayShortway: Agreed.

pubspec.yaml
```diff
@@ -15,17 +15,15 @@ dependencies:
     sdk: flutter
   flutter_web_plugins:
     sdk: flutter
-  freezed_annotation: 3.0.0
-  json_annotation: ^4.8.0
   web: ^1.1.1
+  equatable: ^2.0.0
```
JayShortway, line 19: What's your take on this giving us similar "lock-ins" as freezed did? (Talking about people getting blocked if we don't update soon enough, and having to do a major release when we do update.)

lib/models/web_purchase_redemption_result.dart
```diff
@@ -1,52 +1,58 @@
...
+}
+
+class WebRedemptionInvalidToken extends WebPurchaseRedemptionResult {
+  const WebRedemptionInvalidToken();
+}
+
+class WebRedemptionExpired extends WebPurchaseRedemptionResult {
+  final String obfuscatedEmail;
+  const WebRedemptionExpired({required this.obfuscatedEmail});
+}
```
JayShortway: I think we should keep the word "Purchase" in these subclasses, otherwise we lose some meaning. I wouldn't even be opposed to `WebPurchaseRedemptionResultSuccess`, but `WebPurchaseRedemptionSuccess` works too.

lib/models/customer_info_wrapper.dart
JayShortway, line 1: I noticed all of these model classes are in files named `_wrapper`, but they're not really wrappers (anymore) right? Not the biggest deal, but maybe we can rename the files. (Can also happen separately.)

JayShortway (approved):
Love this, thanks for doing it! This is a big migration, so our migration guide should be thorough, but overall I think it's a good improvement! Only minor stuff.

lib/models/web_purchase_redemption_result.dart
tonidero: Yeah... I tried to make it shorter, but I do think some meaning is lost... Will add that, thanks!

pubspec.yaml
tonidero: I thought about it... I believe it should be less of an issue since Equatable is a smaller library... But there is indeed a chance for them having a breaking change and causing us issues. In that case, we would basically just need to remove it and write the equals/hashcode ourselves (or write something similar to what they do to make it simpler). Either case it shouldn't be complicated, so I don't think we're locking ourselves as bad as with freezed.

JayShortway: Yea, and I think we would be able to maintain API stability in that case right? So we wouldn't need to release a major if we were to remove Equatable?

tonidero: Mostly... I guess the superclass change would be a breaking change... We could write our own `Equatable` superclass to minify it, but if someone expects these classes to inherit from this package's `Equatable` it would break them.

lib/models/customer_info_wrapper.dart
tonidero: This is actually another breaking change, since some users could potentially be referencing the specific files in their imports... I'm kinda leaning towards leaving them, to avoid the breaking change. Lmk if you think otherwise!

JayShortway: Ah yes, let's leave it then. We can file this under "For Historical Reasons" 😄

lib/models/web_purchase_redemption_result.dart
tonidero: Done in https://github.com/RevenueCat/purchases-flutter/pull/1368/commits/613a3d79ad5135ce606314e35d720d98f5d57013

PR #1397, opened 2025-07-08
Add `toJson()` to various classes

devsdocs:

  - [x] A description about what and why you are contributing, even if it's trivial.

  - [x] The issue number(s) or PR number(s) in the description if you are contributing in response to those.

  - [x] If applicable, unit tests.

Add `toJson()` methods manually related to #1368, minimal unit test is added and passed.


tonidero:
Hi @devsdocs, thank you very much for your contribution! On a first look, your code seems good! But I had a few questions on the usage in order to understand it better. Do you need this type of serialization for a specific functionality? If so could you let us know more? 

For context, keeping a smaller API surface helps with maintaining and usability for SDKs like ours so we want to make sure of the use cases for this and make sure it's needed before adding them back. 

Thanks again!

devsdocs:
Hi @tonidero, thank you for the response. I use `.toJson()` mainly for custom caching. While the SDKs also handles caching, having JSON access helps with offline UX, support tools, and reproducing issues. Also support for additional information being shared to external server/database.

tonidero:
Hi again @devsdocs,

So just to make sure, you have some custom caching that stores the RC models by yourself and you are also sharing these objects with your severs? If that's correct could you explain which models you are doing this for and the reason for this? Also, do you need to send the entire objects? Or could specific pieces of information be enough?

We are uncertain whether this is a frequent enough use case for us to maintain these utility functions moving forward (for people that need specific data, they can write their own conversion methods as extension functions with the pieces of data they are interested in), so we want to make sure we're not missing anything.

devsdocs:
Ahh, I see your concern @tonidero, to be more specific, I only use `CustomerInfo` object but store and use the information of the whole object. But if that is your concern, I think it is ok to ignore this PR (or review it later if many user demand it LOL). I'll write custom extension later after next major release.

tonidero:
Sounds good! And yeah we will certainly reconsider this if it's needed by many developers, so it's great to have this :) 

Will close this for now then, but we will keep an eye open in case more people ask for it. Thanks again!

PR #1408, opened 2025-07-17
Change result of purchase methods to PurchaseResult

tonidero:
This changes the result type of the purchase methods. Instead of returning just a `CustomerInfo`, now we will return a `PurchaseResult`, which includes a `CustomerInfo` and a `StoreTransaction`

### DIFF
--- a/api_tester/lib/api_tests/models/purchase_result_api_test.dart
+++ b/api_tester/lib/api_tests/models/purchase_result_api_test.dart
@@ -1,0 +1,22 @@
+import 'package:purchases_flutter/object_wrappers.dart';
+
+// ignore_for_file: unused_element
+// ignore_for_file: unused_local_variable
+// ignore_for_file: deprecated_member_use
+class _PurchaseResultApiTest {
+  void _checkFromJsonFactory(Map<String, dynamic> json) {
+    PurchaseResult transaction = PurchaseResult.fromJson(json);
+  }
+
+  void _checkConstructor(
+      CustomerInfo customerInfo,
+      StoreTransaction storeTransaction) {
+    PurchaseResult purchaseResult =
+    PurchaseResult(customerInfo, storeTransaction);
+  }
+
+  void _checkProperties(PurchaseResult purchaseResult) {
+    CustomerInfo customerInfo = purchaseResult.customerInfo;
+    StoreTransaction storeTransaction = purchaseResult.storeTransaction;
+  }
+}
--- a/lib/models/purchase_result.dart
+++ b/lib/models/purchase_result.dart
@@ -1,0 +1,29 @@
+import 'package:equatable/equatable.dart';
+
+import 'customer_info_wrapper.dart';
+import 'store_transaction.dart';
+
+/// Represents the successful result of a purchase operation.
+class PurchaseResult extends Equatable {
+  /// The updated [CustomerInfo] after the purchase has been synced with
+  /// RevenueCat's servers.
+  final CustomerInfo customerInfo;
+  /// The [StoreTransaction] for this purchase.
+  final StoreTransaction storeTransaction;
+
+  const PurchaseResult(
+    this.customerInfo,
+    this.storeTransaction,
+  );
+
+  factory PurchaseResult.fromJson(Map<String, dynamic> json) => PurchaseResult(
+    CustomerInfo.fromJson(Map<String, dynamic>.from(json['customerInfo'])),
+    StoreTransaction.fromJson(Map<String, dynamic>.from(json['transaction'])),
+  );
+
+  @override
+  List<Object> get props => [
+    customerInfo,
+    storeTransaction,
+  ];
+}
--- a/lib/purchases_flutter.dart
+++ b/lib/purchases_flutter.dart
@@ -341,7 +341,7 @@ class Purchases {
   /// It is now recommended to use [Purchases.purchaseStoreProduct]
   /// to make a purchase with a [StoreProduct] if you can.
   ///
-  /// Makes a purchase. Returns a [CustomerInfo] object. Throws a
+  /// Makes a purchase. Returns a [PurchaseResult] object. Throws a
   /// [PlatformException] if the purchase is unsuccessful.
   /// Check if [PurchasesErrorHelper.getErrorCode] is
   /// [PurchasesErrorCode.purchaseCancelledError] to check if the user cancelled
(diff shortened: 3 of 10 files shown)


lib/models/purchase_result.dart
```diff
@@ -0,0 +1,29 @@
...
+import 'customer_info_wrapper.dart';
+import 'store_transaction.dart';
+
+/// Represents the successful result of a purchase operation.
+class PurchaseResult extends Equatable {
+  /// The updated [CustomerInfo] after the purchase has been synced with
+  /// RevenueCat's servers.
+  final CustomerInfo customerInfo;
+  /// The [StoreTransaction] for this purchase.
+  final StoreTransaction storeTransaction;
```
tonidero, line 12: I see that in the web platforms (RN and capacitor) we also have the `productIdentifier` here... but seemed redundant since it's already part of the `PurchaseResult`.

lib/web/purchases_flutter_web.dart
```diff
@@ -12,7 +12,7 @@ import '../purchases_flutter.dart';
 class PurchasesFlutterPlugin {
   static final _unknownErrorCode = '${PurchasesErrorCode.unknownError.index}';
   static final _configurationErrorCode = '${PurchasesErrorCode.configurationError.index}';
-  static const _purchasesHybridMappingsVersion = '13.29.0';
+  static const _purchasesHybridMappingsVersion = '15.0.0';
```
tonidero, line 15: This was super outdated 😬

lib/web/purchases_flutter_web.dart
```diff
@@ -162,14 +162,16 @@ class PurchasesFlutterPlugin {
...
     final options = {
       'apiKey': apiKey,
-      'appUserId': appUserId,
       'flavor': _platformName,
       'flavorVersion': _pluginVersion,
     };
+    final appUserId = arguments['appUserId'] as String?;
+
+    if(appUserId != null && appUserId.isNotEmpty) {
+      options['appUserId'] = appUserId;
```
tonidero, line 173: Had to change this to be added like this since some changes in the underlying JS libraries made it break without this because we were passing null as the appUserId. A proper fix might be better in `purchases-js-hybrid-mappings`... but I thought to fix it here for now to avoid another release.

api_tester/lib/api_tests/models/purchase_result_api_test.dart
```diff
@@ -0,0 +1,22 @@
...
+  void _checkFromJsonFactory(Map<String, dynamic> json) {
+    PurchaseResult transaction = PurchaseResult.fromJson(json);
+  }
+
+  void _checkConstructor(
+      CustomerInfo customerInfo,
+      StoreTransaction storeTransaction) {
+    PurchaseResult purchaseResult =
+    PurchaseResult(customerInfo, storeTransaction);
+  }
```
ajpallares, line 16: I don't quite understand why we need to test the constructor's API. 
Is the constructor public? If so, if we ever need to add an extra property to `PurchaseResult`, it would be a breaking change since the constructor would also change. Or am I missing something? (note that I know `null` about Dart 😅)

tonidero: That's a good question! So the constructors for all our classes are currently public in dart (they were public before removing freezed as well). If we need to add new fields in the future, we would need to add multiple constructors, falling back to some default when no value is passed. This can indeed be problematic if no good default can be set though...

We could try to make the constructors package private... we haven't really been doing that until now, but I guess that now that we have removed freezed, we have more control. I might just want to leave it like this for now so we dedicate more time to study our options moving forward. Wdyt @RevenueCat/coresdk?

ajpallares: Thanks for the explanation! Exactly, it might be problematic if we need to add new fields in the future.
> We could try to make the constructors package private...

Yes, I think this should be our end goal. But I agree that, for now, we can leave it like this and think about this in the future 👍

JayShortway: Agreed, I like non-public constructors, but doing that now would be a breaking change right?

tonidero: Yup... though having said that, it would be "ok" since we're releasing this on a major anyway. My main concern is over delaying the release for this, since I'm not sure of the best way to do this.

JayShortway: Hah I'm not sure either. It's not just adding visibility modifiers? 😅

tonidero: I think for package private you need to define each file as part of a library, and prepend any method with an `_`: https://medium.com/@simplemanderson/dart-libraries-and-visibility-and-how-they-caught-me-by-surprise-d98ba7868fde.

I just need to test that it works 😅  (I'm focusing on returning the StoreTransaction in purchases-unity after a purchase but I can test this in a bit)

tonidero: I was just testing this for a bit and it's not straightforward... We would need to define a library and make all models a `part of` that library... I wouldn't make that change right now without more testing, so I would prefer to keep it as is for this major. Wdyt @JayShortway @ajpallares ?

JayShortway: Thanks for explaining and testing! 🙏 In that case I'd be okay keeping it as is.

ajpallares: Thank you for the investigation Toni! Sure, let's go with what we have right now, as it involves way deeper changes 👍

lib/models/purchase_result.dart
ajpallares: I'm guessing you mean that it's part of the `StoreTransaction`, right? So yeah, totally not needed to add the redundancy 👍

lib/purchases_flutter.dart
ajpallares, line 1: I see there are two methods that have not been migrated to return `PurchaseResult`. Is that intended?
The two methods are:
```dart
static Future<CustomerInfo> purchaseProductWithWinBackOffer(
  StoreProduct product,
  WinBackOffer winBackOffer,
)

static Future<CustomerInfo> purchasePackageWithWinBackOffer(
  Package package,
  WinBackOffer winBackOffer,
)
```

tonidero: Thanks for the catch!! I missed those 🤦 Changed in https://github.com/RevenueCat/purchases-flutter/pull/1408/commits/649d72a0d457bb263e0b7652d99439d713d8ae5d

revenuecat_examples/purchase_tester/lib/src/upsell.dart
```diff
@@ -285,9 +285,10 @@ class _PurchaseButton extends StatelessWidget {
 
   Future<void> _purchasePackage(BuildContext context, Package package) async {
     try {
-      final customerInfo = await Purchases.purchasePackage(package);
-      final isPro =
-          customerInfo.entitlements.active.containsKey(entitlementKey);
+      final purchaseResult = await Purchases.purchasePackage(package);
+      final isPro = purchaseResult.customerInfo.entitlements
+        .active.containsKey(entitlementKey);
+      print("StoreTransaction: ${purchaseResult.storeTransaction}");
```
ajpallares, line 291: Just to confirm. Is this print intended to be commited?

ajpallares (approved):
Great job Toni!!!

revenuecat_examples/purchase_tester/lib/src/upsell.dart
tonidero: I would like to leave it as I think it can be useful... This is the test app in any case so I think it's ok

PR #1411, opened 2025-07-18
Fix wrong SDK version on web for analytics

tonidero:
After merging the v9 base branch in #1407, we didn't update the version of the SDK, so it wasn't caught by our automations. This mismatch can cause some of our analytics to not be set correctly. This fixes the number so it works moving forward.


fire-at-will (approved):
Looks good! Is there anything we need to update in our release automations to prevent this from happening again?

tonidero:
> Is there anything we need to update in our release automations to prevent this from happening again?

We should change to do the replacement of the strings even if the previous string doesn't match the previous version. Can try to work on that next week.

Issue #1433, opened 2025-08-16
[Web] Type 'double' is not a subtype of type 'int' in type cast when build in wasm in Flutter Web


westlinkin:
- [x] I have updated Purchases SDK to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-flutter/blob/main/CONTRIBUTING.md)
- [x] I have searched the [Community](https://community.revenuecat.com)
- [x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [x] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-flutter/issues)

# ‼️ Required data ‼️ #
**Do not remove any of the steps from the template below. If a step is not applicable to your issue, please leave that step empty.**

There are a lot of things that can contribute to things not working. Having a very basic understanding of your environment will help us understand your issue faster!

## Environment ##
- [x] Output of `flutter doctor`
```
Doctor summary (to see all details, run flutter doctor -v):
[!] Flutter (Channel stable, 3.32.8, on macOS 15.6 24G84 darwin-arm64, locale en-US)
    ! The flutter binary is not on your path. Consider adding /Users/Linkin/Study/flutter/bin to your path.
    ! The dart binary is not on your path. Consider adding /Users/Linkin/Study/flutter/bin to your path.
[!] Android toolchain - develop for Android devices (Android SDK version 36.0.0)
    ! Some Android licenses not accepted. To resolve this, run: flutter doctor --android-licenses
[✓] Xcode - develop for iOS and macOS (Xcode 16.4)
[✓] Chrome - develop for the web
[✓] Android Studio (version 2025.1)
[✓] IntelliJ IDEA Community Edition (version 2025.2)
[✓] VS Code (version 1.103.0)
[✓] Connected device (2 available)
    ! Error: Browsing on the local area network for lyric的iPad. Ensure the device is unlocked and attached with a cable or associated with the same
      local area network as this Mac.
      The device must be opted into Developer Mode to connect wirelessly. (code -27)
[✓] Network resources
```

- [x] Version of `purchases-flutter`
9.2.0
- [x] Testing device version e.g.: iOS 15.5, Android API 30, etc.
Chrome
- [x] How often the issue occurs- every one of your customers is impacted? Only in dev?
When build in wasm mode
- [x] [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue
```
[...]

PR #1434, opened 2025-08-18
Fix WASM parsing

tonidero:
There were some parsing issues when using WASM that made it impossible to obtain offerings and CustomerInfo related to parsing double numbers that were meant to be integers. This fixes those by just rounding to the nearest integer. This should be fine since these values are meant to be integer originally.

This should help dealing with #1433 

Still investigating another issue that makes the purchase flow get stuck loading on WASM which will be addressed in future PRs

### DIFF
--- a/revenuecat_examples/purchase_tester/pubspec.yaml
+++ b/revenuecat_examples/purchase_tester/pubspec.yaml
@@ -15,6 +15,11 @@ dependencies:
 
   app_links: ^6.3.3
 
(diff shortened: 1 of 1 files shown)


tonidero:
Right it seems that WASM converts all numbers in the json to double when parsing them actually, since from my tests, they already should be integers... Not sure why it does that though.

tonidero:
@vegaro, So this seems to be an issue in WASM, See this issue: https://github.com/flutter/flutter/issues/159400

PR #1443, opened 2025-08-28
Virtual Currency Support for Flutter Web

fire-at-will:
## Description
This PR implements the `getVirtualCurrencies`, `getCachedVirtualCurrencies`, and `invalidateVirtualCurrencies` functions for the Flutter web platform.

It also adds a few print statements to the VC testing screen to help with debugging.

## Testing
- Manually tested the new VC functions in the purchase tester app in Chrome


lib/web/purchases_flutter_web.dart
```diff
@@ -314,6 +317,21 @@ class PurchasesFlutterPlugin {
...
+  Future<void> _invalidateVirtualCurrenciesCache() async {
+    _getInstance().callMethod('invalidateVirtualCurrenciesCache'.toJS);
+  }
+
+  Future<Map<String, dynamic>?> _getCachedVirtualCurrencies() async {
+    final cachedVirtualCurrencies = _getInstance().callMethod('getCachedVirtualCurrencies'.toJS);
+    if (cachedVirtualCurrencies == null) {
+      return null;
+    }
+    return _convertJsRecordToMap(cachedVirtualCurrencies);
```
tonidero: Hmm can we use `_getNullableMapFromInstanceMethod` here?

tonidero (approved):
Just a possible cleanup but looks good to me! Great job!

lib/web/purchases_flutter_web.dart
fire-at-will: Using `_getNullableMapFromInstanceMethod` here actually causes an error when it gets executed. This is because `_getNullableMapFromInstanceMethod` assumes that the JS function will return a promise, but the `getCachedVirtualCurrencies` and `invalidateVirtualCurrenciesCache` functions in PurchaseCommon from PHC are actually synchronous methods that return a nullable record or void.

This approach works here, but can you think of a reason that maybe we should change the PHC functions to return promises, since the other functions there return promises?

tonidero: Ohh right! Hmm I think not TBH... most of our functions are asynchronous, so it makes sense for them to be promises. So yeah, I think this is good then, thanks for explaining!

MarkVillacampa (approved):

lib/web/purchases_flutter_web.dart
fire-at-will: At the very least, I think I might extract this logic to a common `_getNullableMapFromInstanceSyncMethod` so it can be reused in the future :)

fire-at-will: @tonidero I've extracted that logic in 5dbcda35dfd429fa01b0339a2157898bcead8dea, which will let us re-use this in any future sync function calls that return nullable maps 🙂

Issue #1219, opened 2024-12-03
Add Swift Package Manager support


loic-sharma:
Hello,

Flutter is migrating to [Swift Package Manager][]. Please [add Swift Package Manager support][] to your plugin.

Flutter will eventually deprecate and then remove support for CocoaPods. Adding Swift Package Manager support future-proofs your plugins.

[Swift Package Manager]: https://www.swift.org/documentation/package-manager/
[add Swift Package Manager support]: https://docs.flutter.dev/packages-and-plugins/swift-package-manager/for-plugin-authors#how-to-add-swift-package-manager-support-to-an-existing-flutter-plugin

### How to add Swift Package Manager support to your plugins

You can find the migration guide here: https://docs.flutter.dev/packages-and-plugins/swift-package-manager/for-plugin-authors#how-to-add-swift-package-manager-support-to-an-existing-flutter-plugin

If you run into issues or have questions, please reach out to the Flutter team. You can ping me at @loic-sharma, send a message to the #hackers-ios channel on [Flutter’s discord][], or [open a GitHub issue][].

Thank you for your wonderful contributions to the Flutter ecosystem!

[Flutter’s discord]: https://github.com/flutter/flutter/blob/master/docs/contributing/Chat.md
[open a GitHub issue]: https://github.com/flutter/flutter/issues/new/choose

### Why is Flutter migrating to Swift Package Manager?

[CocoaPods is now in maintenance mode][].

Swift Package Manager support is one of Flutter’s most requested features: [flutter#33850][].

Flutter's Swift Package Manager integration has several benefits:

1. **Access to the Swift package ecosystem**. Flutter plugins can use the growing ecosystem of [Swift packages][].
2. **Simplifies Flutter installation**. Swift Package Manager is bundled with Xcode. In the future, you won’t need to install Ruby and CocoaPods to target iOS or macOS.

[CocoaPods is now in maintenance mode]: https://blog.cocoapods.org/CocoaPods-Support-Plans/
[flutter#33850]: https://github.com/flutter/flutter/issues/33850
[Swift packages]: https://swiftpackageindex.com/

RCGitBot:
👀 We've just linked this [issue to our internal tracker](https://revenuecat.zendesk.com/agent/tickets/49767) and notified the team. Thank you for reporting, we're checking this out!

kmurphy-rc:
Hi @loic-sharma,
[...]

PR #1452, opened 2025-09-03
Support integration via Swift Package Manager

ajpallares:
As mentioned in the [official Flutter documentation](https://docs.flutter.dev/packages-and-plugins/swift-package-manager/for-app-developers)
> Flutter is migrating to [Swift Package Manager](https://www.swift.org/documentation/package-manager/) to manage iOS and macOS native dependencies

This PR adds some Package.swift files, allowing purchases-flutter to be integrated via SPM. 

Also requested in #1219


.circleci/config.yml
```diff
@@ -335,6 +351,50 @@ jobs:
...
+    macos:
+      xcode: 15.2
+    shell: /bin/bash --login -o pipefail
+    steps:
+      - checkout
+      - macos/preboot-simulator:
+          device: iPhone 14
+          version: '16.4'
+      - setup-flutter
+      - enable-spm-feature
```
tonidero: Hmm just wondering, what's the default state? Is it enabled by default, or disabled? Just wondering if we need both an enable and disable methods, or just one.

ios/purchases_flutter.podspec
```diff
@@ -12,8 +12,8 @@ Pod::Spec.new do |s|
   s.license          = { :file => '../LICENSE' }
   s.author           = { 'RevenueCat' => 'support@revenuecat.com' }
   s.source           = { :path => '.' }
-  s.source_files = 'Classes/**/*'
-  s.public_header_files = 'Classes/**/*.h'
+  s.source_files = 'purchases_flutter/Sources/purchases_flutter/**/*.{h,m}'
```
tonidero, line 15: Hmm is moving the files so nested needed? Just wondering if it could just be something like `purchases_flutter/Sources/**/*.{h,m}`...

purchases_ui_flutter/ios/purchases_ui_flutter.podspec
```diff
@@ -13,7 +13,7 @@ Flutter plugin that integrates RevenueCat Paywalls
   s.license          = { :file => '../LICENSE' }
   s.author           = { 'RevenueCat' => 'support@revenuecat.com' }
   s.source           = { :path => '.' }
-  s.source_files = 'Classes/**/*'
+  s.source_files = 'purchases_ui_flutter/Sources/purchases_ui_flutter/**/*.swift'
```
tonidero, line 16: Same question regarding the new location of files as in the main SDK.

.circleci/config.yml
ajpallares: Good point. ~I think that Swift Package Manager is still disabled by default~, but I'm not 100% sure. In any case, I did this intentionally for two reasons:
1. To make it explicit whether or not SPM is used
2. To guard against Flutter changing its default state from Cocoapods to SPM (which will happen eventually). This way, integration tests do not depend on the default state.

ios/purchases_flutter.podspec
ajpallares: Right. Probably it's not strictly needed. I did it this way to respect the default folder structure of Swift Package Manager. That said, I'm thinking that perhaps I could simplify it with
* Simpler folder structure (less nesting)
* Unique Package.swift file shared for iOS and macOS
Let me try moving things around.
Thanks for the suggestion!

ajpallares: I tried to move things around to simplify the Package.swift but couldn't make it work. Originally I followed the official migration guide [How to add Swift Package Manager support to an existing Flutter plugin](https://docs.flutter.dev/packages-and-plugins/swift-package-manager/for-plugin-authors#how-to-add-swift-package-manager-support-to-an-existing-flutter-plugin) and there it specifies the exact same nesting of files I did

> Your plugin should look like:
> 
> ```
> plugin_name/ios/
> ├── ...
> └── plugin_name/
>    ├── Package.swift
>    └── Sources/plugin_name/include/plugin_name/
>       └── .gitkeep
> ```

I agree that it's a bit convoluted but perhaps it's better to be safe in this regard. FWIW, Firebase's [flutterfire](https://github.com/firebase/flutterfire) also has a similar nesting (e.g. https://github.com/firebase/flutterfire/blob/main/packages/firebase_auth/firebase_auth/ios/firebase_auth/Sources/firebase_auth/include/Public/FLTFirebaseAuthPlugin.h)

tonidero: That makes total sense, and yeah totally agree we should follow documentation. had not read it before reviewing 🥴. But yeah, this seems good then!

.circleci/config.yml
```diff
@@ -304,6 +318,7 @@ jobs:
           device: iPhone 14
           version: '16.4'
       - setup-flutter
+      - disable-spm-feature
```
tonidero: Do we need to enable/disable SPM on the example project as well? Just checking since maybe the sample app is still using cocoapods always right? Not sure if we need to have a separate sample app to test using SPM though 🤔

tonidero (commented):
Just a last question. But the changes to the SDK make sense!

.circleci/config.yml
ajpallares: Great eye!! 👀 
Yes, we need to enable/disable SPM on the example project to make sure that the correct integration method is being used. 
That said, I've tried removing the two Package.swift files, force SPM with `flutter config --enable-swift-package-manager` and then
```
flutter build ios  
```
The result is that it uses SPM only for those dependencies that it can find. Since I removed the two Package.swift files, it falls back to CocoaPods for purchases_flutter and purchases_ui_flutter (❗).

It looks like I need to either have a separate sample app or, perhaps, include in `disable-spm-feature` a step to remove the `podfile`. Thanks for the suggestion!

ajpallares: @tonidero I finally implemented a simpler solution in https://github.com/RevenueCat/purchases-flutter/pull/1452/commits/fdbfb9472268a7da14100806a70fbea9f87619d6. The idea is to remove all `.podspec` files for SPM integration tests and remove all `Package.swift` files for CocoaPods integration tests. This way, we make sure that the local dependency can only be resolved via the desired package manager (thank you @aboedo for the idea!).

.circleci/config.yml
```diff
@@ -179,6 +179,26 @@ commands:
...
 
+  prepare-for-spm-integration-test:
+    description: Enable Swift Package Manager support in Flutter + remove local .podspec files
+    steps:
+      - run:
+          name: Enable SPM feature
+          command: flutter config --enable-swift-package-manager
+      - run:
+          name: Remove all local .podspec files recursively
+          command: find . -name "*.podspec" -type f -delete
```
tonidero, line 190: I wonder if we should have all `podspec` files renamed to `podspec.SAMPLE` or something, and then we rename those on CI for tests and when releasing... My logic is to make sure that, when developing locally, we use SPM, which seems like the direction we should follow... Otherwise, when developing locally, we might not be using SPM but cocoapods, possibly causing us to miss issues.

On the other hand, if we do this, I guess we would be dropping support for cocoapods when developing locally, so we wouldn't be testing the exact same thing developers are seeing... So... after all this brain dump, I think maybe it's ok as is 😅

tonidero (approved):
Just a bit of a brain dump, but I think this looks great! Amazing to get this!

.circleci/config.yml
ajpallares: That's also a very good point 🤔 

Firstly, I've tested in a VM, and I _think_ that Flutter does default to SPM, so in theory we would be relying on SPM when developing locally. So we could theoretically remove all traces of CocoaPods from the purchase tester sample (in particular, the Podfile). If it ever uses CocoaPods (e.g. due to a SPM issue/misconfiguration), we would see the Podfile in the uncommitted changes. As a last resort, the integration tests should probably catch any potential errors we might introduce regarding SPM integration.

And if we want use CocoaPods locally to develop/test, we can always run
```
flutter config --no-enable-swift-package-manager
```

or [temporarily turn off SPM for the project](https://docs.flutter.dev/packages-and-plugins/swift-package-manager/for-app-developers#turn-off-for-a-single-project).

WDYT?

tonidero: Hmm I think that could be good to do... but not a strong opinion, if it uses SPM by default, I'm ok leaving it as is as well.

ajpallares: It does use SPM by default, but since there are "traces" of CocoaPods, it also installs the pods, even though the purchases_flutter dependency gets added via SPM.

I agree that this improvement (removing all traces of CocoaPods from the example) would be good to do. I think it can go in a separate PR though.

ajpallares: Done in this PR https://github.com/RevenueCat/purchases-flutter/pull/1454

PR #1454, opened 2025-09-05
Default to Swift Package Manager in Purchase Tester (instead of CocoaPods)

ajpallares:
This PR removes all the traces of CocoaPods from the Purchase Tester example app of the repo. This way, purchases_flutter is added as a local dependency via Swift Package Manager.

The Purchase Tester project is also ready to seamlessly work with CocoaPods for development when needed. To do that, just run
```bash
flutter config --no-enable-swift-package-manager
```
at the beginning of the session.

To re-enable Swift Package Manager, run 
```bash
flutter config --enable-swift-package-manager
```

## Notes for reviewer
The changes in this PR are:
* Remove all traces of CocoaPods from the Purchase Tester example app
* Commit all file changes generated by flutter when doing `flutter pub get` and `flutter build ios (macos)` when using SPM


revenuecat_examples/purchase_tester/ios/RunnerTests/RunnerTests.swift
ajpallares, line 1: To keep supporting CocoaPods integration without errors, I had to add a dummy RunnerTests target to the Xcode project. The reason is that the Podfile flutter autogenerates expects this target to exist. Otherwise, an error would raise when running `flutter build ios` if using CocoaPods

revenuecat_examples/purchase_tester/ios/RunnerTests/RunnerTests.swift
```diff
@@ -0,0 +1,17 @@
...
+//  Created by Antonio Pallares on 5/9/25.
+//  Copyright © 2025 The Chromium Authors. All rights reserved.
+//
+
+import Testing
+
+struct RunnerTests {
+
+    @Test func example() async throws {
+        // Write your test here and use APIs like `#expect(...)` to check expected conditions.
```
tonidero, line 14: Should we add your comment above here so we don't delete this accidentally?

revenuecat_examples/purchase_tester/macos/Flutter/GeneratedPluginRegistrant.swift
```diff
@@ -7,10 +7,8 @@ import Foundation
 
 import app_links
 import purchases_flutter
-import purchases_ui_flutter
 
 func RegisterGeneratedPlugins(registry: FlutterPluginRegistry) {
   AppLinksMacosPlugin.register(with: registry.registrar(forPlugin: "AppLinksMacosPlugin"))
   PurchasesFlutterPlugin.register(with: registry.registrar(forPlugin: "PurchasesFlutterPlugin"))
-  PurchasesUiFlutterPlugin.register(with: registry.registrar(forPlugin: "PurchasesUiFlutterPlugin"))
```
tonidero, line 15: This happens locally for me as well but I wasn't sure whether to check it in github since st seems like it should be added right? I think it might be due to the UI plugin not working on macos?

tonidero (approved):
Looks great! Thanks for doing this!

PR #1456, opened 2025-09-05
Fix release scripts

tonidero:
We didn't update the path in #1454. This fixes that


ajpallares (approved):
Oh! Great catch! Sorry I missed this and thank you for fixing it!

PR #1460, opened 2025-09-07
Pass `presentedOfferingContext` to natives when presenting paywall or displaying PaywallView

tonidero:
We were not passing the presentedOfferingContext to the natives when the developer chose a non-current offering. This could cause lose of data. This fixes that.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -3,14 +3,14 @@ version '9.5.0'
 
 buildscript {
     ext.kotlin_version = '1.8.22'
-    ext.common_version = '17.5.0'
+    ext.common_version = '17.5.1'
     repositories {
         google()
         mavenCentral()
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:8.12.2'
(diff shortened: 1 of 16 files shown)


ios/purchases_flutter/Package.swift
```diff
@@ -12,7 +12,7 @@ let package = Package(
         .library(name: "purchases-flutter", targets: ["purchases_flutter"])
     ],
     dependencies: [
-        .package(url: "https://github.com/RevenueCat/purchases-hybrid-common.git", exact: "17.3.0")
+        .package(url: "https://github.com/RevenueCat/purchases-hybrid-common.git", exact: "17.5.0")
```
tonidero, line 15: Seems this one got mismatched at some point and not updated 😬

tonidero:
Bump @RevenueCat/coresdk

ajpallares (approved):
Sorry, I missed this one!

I don't have a lot of context (🥁) on this one, but I took a look and it makes sense!

PR #1461, opened 2025-09-08
Add `purchase` method with `PurchaseParams` object

tonidero:
This adds a new class `PurchaseParams` that allows to specify the different possible parameters that you can pass to the purchase methods.

This pattern allows to add more parameters to the purchase methods, making it more flexible. No new functionality is added in this PR, just deprecates the old APIs in favor of the new one.

- [ ] Update docs

### DIFF
--- a/test/purchases_flutter_test.dart
+++ b/test/purchases_flutter_test.dart
@@ -597,7 +597,7 @@ void main() {
           1234567890,
       );
       final purchaseParams = PurchaseParams.package(
-        package: mockPackage,
+        mockPackage,
         googleProductChangeInfo: GoogleProductChangeInfo(
           'old_product_id',
           prorationMode: GoogleProrationMode.immediateAndChargeFullPrice,
@@ -659,7 +659,7 @@ void main() {
         PresentedOfferingContext('main', null, null),
       );
       const purchaseParams = PurchaseParams.package(
-          package: mockPackage,
+          mockPackage,
       );
       final purchasePackageResult =
       await Purchases.purchase(purchaseParams);
@@ -719,7 +719,7 @@ void main() {
         1234567890,
       );
       final purchaseParams = PurchaseParams.storeProduct(
-          storeProduct: mockStoreProduct,
+          mockStoreProduct,
           googleProductChangeInfo: GoogleProductChangeInfo(
             'old_product_id',
             prorationMode: GoogleProrationMode.immediateAndChargeFullPrice,
@@ -777,7 +777,7 @@ void main() {
         productCategory: ProductCategory.nonSubscription,
       );
       const purchaseParams = PurchaseParams.storeProduct(
-          storeProduct: mockStoreProduct,
+          mockStoreProduct,
       );
       final purchaseProductResult =
       await Purchases.purchase(purchaseParams);
@@ -832,7 +832,7 @@ void main() {
         null,
       );
       final purchaseParams = PurchaseParams.subscriptionOption(
-          subscriptionOption: mockSubscriptionOption,
+          mockSubscriptionOption,
           googleProductChangeInfo: GoogleProductChangeInfo(
             'old_product_id',
             prorationMode: GoogleProrationMode.immediateAndChargeFullPrice,
@@ -895,7 +895,7 @@ void main() {
         null,
       );
       const purchaseParams = PurchaseParams.subscriptionOption(
-          subscriptionOption: mockSubscriptionOption,
+          mockSubscriptionOption,
       );
       final purchaseSubscriptionOptionResult =
       await Purchases.purchase(purchaseParams);


lib/models/purchase_params.dart
```diff
@@ -0,0 +1,38 @@
...
+import 'purchasable.dart';
+import 'store_product_wrapper.dart';
+import 'subscription_option_wrapper.dart';
+
+/// Parameters for making a purchase.
+class PurchaseParams {
+  final Purchasable purchasableItem;
+  final GoogleProductChangeInfo? googleProductChangeInfo;
+  final bool? googleIsPersonalizedPrice;
+  final PromotionalOffer? promotionalOffer;
```
tonidero, line 14: Right now this allows potentially passing both a `googleProductChangeInfo` and a `promotionalOffer` at the same time, which doesn't make much sense... But I think it's ok? The platform where the other parameter doesn't work will be ignored.

lib/models/purchasable.dart
```diff
@@ -0,0 +1,7 @@
+
+import 'presented_offering_context_wrapper.dart';
+
+/// Interface for items that can be purchased
+abstract class Purchasable {
```
tonidero: I thought about the best way of doing this in dart... The other option I thought was to use sealed classes, but we're trying to avoid sealed classes to make it extensible in the future without causing breaking changes.

Also thought about creating multiple constructors of the `PurchaseParams` object, each receiving a Package/StoreProduct/SubscriptionOption, but that caused more duplication IMO...

lib/purchases_flutter.dart
```diff
@@ -555,6 +560,47 @@ class Purchases {
...
+    final presentedOfferingContext = purchaseParams.purchasableItem.presentedOfferingContext?.toJson();
+    final purchaseArgs = <String, dynamic>{
+      'googleOldProductIdentifier': googleProductChangeInfo?.oldProductIdentifier,
+      'googleProrationMode': prorationMode,
+      'googleIsPersonalizedPrice': googleIsPersonalizedPrice,
+      'signedDiscountTimestamp': signedDiscountTimestamp,
+      'presentedOfferingContext': presentedOfferingContext,
+    };
+    if (purchasableItem is Package) {
+      return await _invokeReturningPurchaseResult('purchasePackage', {
```
tonidero, line 585: This still uses the same native code to handle the purchases. I just changed the dart API and how we process these there. We could handle this logic in the native layer and just pass the entire purchaseParams... but that was more complex, and I think it would lead to extra duplication of logic in iOS/Android/web

lib/models/purchasable.dart
ajpallares: One question. This would open up the possibility of someone implementing their own class that implements `Purchasable`, right?
I guess it's OK as it's dealt with as an error in the `purchase` method

lib/purchases_flutter.dart
ajpallares: Yes, I think this is fine. And I think it also makes it easier to discover the supported `Purchasable` items (following the implementation is simple and gives you the clue)

lib/models/purchasable.dart
tonidero: That's correct, but as you said, it would fail later when trying to purchase it

ajpallares: Great! I think it's fine 👍

tonidero: One other thing, I also thought about following the same pattern we do with `PurchasesConfiguration`... but felt to me this is better because it ends up with an immutable object. In `PurchasesConfiguration`, all parameters are mutable, which I didn't fully like... on the other hand, it's not as consistent to do something different. Lmk if you have any concerns over this @RevenueCat/coresdk

ajpallares (approved):
Nice! I really like this approach. Thank you for doing this!

tonidero:
@Jethro87 I would love your thoughts on the new APIs as well :)

lib/models/purchase_params.dart
```diff
@@ -0,0 +1,110 @@
...
+  ///
+  /// [package] The package to purchase.
+  ///
+  /// [googleProductChangeInfo] Android and Google Play only. Optional [GoogleProductChangeInfo] you wish to
+  /// change from containing the googleOldProductIdentifer and the optional prorationMode.
+  ///
+  /// [googleIsPersonalizedPrice] Android and Google Play only. Optional boolean indicates
+  /// personalized pricing on products available for purchase in the EU.
+  /// For compliance with EU regulations. User will see "This price has been
+  /// customize for you" in the purchase dialog when true.
```
Jethro87:
```suggestion
  /// customized for you" in the purchase dialog when true.
```

Jethro87 (changes requested):

Jethro87:
This looks great!

I haven't requested the following in a review yet, but what do you think about changing the named constructors to accept a required positional param (package, product, subscriptionOption) instead of a required optional param? @tonidero @ajpallares. 

So instead of the developer doing this:
```
await Purchases.purchase(PurchaseParams.package(package: package, googleProductChangeInfo: info)); // or
await Purchases.purchase(PurchaseParams.storeProduct(product: product, promotionalOffer: offer)); // or
await Purchases.purchase(PurchaseParams.subscriptionOption(subscriptionOption: option, googleIsPersonalizedPrice: true));
```
they'll do this:
```
await Purchases.purchase(PurchaseParams.package(package, googleProductChangeInfo: info)); // or
await Purchases.purchase(PurchaseParams.storeProduct(product, promotionalOffer: offer)); // or
await Purchases.purchase(PurchaseParams.subscriptionOption(option, googleIsPersonalizedPrice: true));
```

The constructors would then look like:
```
const PurchaseParams.package(
  Package package, {
  GoogleProductChangeInfo? googleProductChangeInfo,
  bool? googleIsPersonalizedPrice,
  PromotionalOffer? promotionalOffer,
}) : this._(
        package,
        null,
        null,
        googleProductChangeInfo,
        googleIsPersonalizedPrice,
        promotionalOffer,
      );

const PurchaseParams.storeProduct(
  StoreProduct storeProduct, {
  GoogleProductChangeInfo? googleProductChangeInfo,
  bool? googleIsPersonalizedPrice,
  PromotionalOffer? promotionalOffer,
}) : this._(
        null,
        storeProduct,
        null,
        googleProductChangeInfo,
        googleIsPersonalizedPrice,
        promotionalOffer,
      );

const PurchaseParams.subscriptionOption(
  SubscriptionOption subscriptionOption, {
  GoogleProductChangeInfo? googleProductChangeInfo,
  bool? googleIsPersonalizedPrice,
}) : this._(
        null,
        null,
        subscriptionOption,
        googleProductChangeInfo,
        googleIsPersonalizedPrice,
        null,
      );
```

tonidero:
Thanks for the feedback @Jethro87! I have indeed seen this pattern used a lot so I think it does make sense. I've applied the changes in https://github.com/RevenueCat/purchases-flutter/pull/1461/commits/c0ba7a52f4240b8e7349db3b3da76b2965af4ccb.

Thinking about it, I think a rule of thumb could be that, if there is a single required parameter, positional makes more sense, but if there are multiple, named parameters might make more sense... Lmk if you think otherwise!

lib/models/purchase_params.dart
tonidero: Applied fix in https://github.com/RevenueCat/purchases-flutter/pull/1461/commits/c0ba7a52f4240b8e7349db3b3da76b2965af4ccb

Issue #1471, opened 2025-09-15
Crash on new purchase() API on iOS


apizon:
- [x] I have updated Purchases SDK to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-flutter/blob/main/CONTRIBUTING.md)
- [x] I have searched the [Community](https://community.revenuecat.com)
- [x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [x] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-flutter/issues)

# ‼️ Required data ‼️ #
**Do not remove any of the steps from the template below. If a step is not applicable to your issue, please leave that step empty.**

There are a lot of things that can contribute to things not working. Having a very basic understanding of your environment will help us understand your issue faster!

## Environment ##
- [x] Output of `flutter doctor`

[✓] Flutter (Channel stable, 3.35.3, on macOS 15.6 24G84 darwin-arm64, locale en-US)
[✗] Android toolchain - develop for Android devices
    ✗ Unable to locate Android SDK.
      Install Android Studio from: https://developer.android.com/studio/index.html
      On first launch it will assist you in installing the Android SDK components.
      (or visit https://flutter.dev/to/macos-android-setup for detailed instructions).
      If the Android SDK has been installed to a custom location, please use
      `flutter config --android-sdk` to update to that location.

[✓] Xcode - develop for iOS and macOS (Xcode 16.4)
[✗] Chrome - develop for the web (Cannot find Chrome executable at /Applications/Google Chrome.app/Contents/MacOS/Google Chrome)
    ! Cannot find Chrome. Try setting CHROME_EXECUTABLE to a Chrome executable.
[!] Android Studio (not installed)
[✓] Connected device (2 available)
[✓] Network resources

- [x] Version of `purchases-flutter`: 9.6.0
- [x] Testing device version e.g.: iOS 18.6.2
- [x] How often the issue occurs: occurs every time, found during development
- [x] [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue
[crash.log](https://github.com/user-attachments/files/22348865/crash.log)

- [x] Steps to reproduce, with a description of expected vs. actual behavior

To reproduce:
```
final offerings = await Purchases.getOfferings()
final monthly = offerings.current?.monthly;
await Purchases.purchase(PurchaseParams.package(monthly));
```

Expected: purchase menu appears
[...]

PR #1472, opened 2025-09-15
Fix crash when calling new purchase w/ purchase params function

fire-at-will:
Addresses https://github.com/RevenueCat/purchases-flutter/issues/1471 by passing `nil` to `RCCommonFunctionality.purchase` for `signedDiscountTimestamp` instead of `NSNull`, which appears to be causing issues when bridging to Swift. 

It accomplishes this by porting over the `mappingNSNullToNil` function from the React Native SDK.

### Explanation
The crash occurs between when `PurchasesFlutterPlugin.m` calls `RCCommonFunctionality`'s purchase function, so the issue appears to be with the Objective-C/Swift bridging. After narrowing it down, we found that passing `nil` instead of `NSNull` for the `signedDiscountTimestamp` parameter resolved the issue.

From https://www.andyibanez.com/posts/nil-null-mess-objective-c-and-swift/ :
> In Objective-C, you can send messages to nil, but when you are sending messages to a real object, the object needs to implement the method you want to call - in other words, you want the object to respond to a selector. If you send a message to an object and the object does not respond to it, you will crash with a message similar to “unrecognized selector sent to instance 0xb4df00d”.

The stacktrace for our crash shows: `reason: '-[NSNull length]: unrecognized selector sent to instance`, so it appears that for some reason, the objc/swift bridging appears to be attempting to call length() on NSNull. NSNull is a Foundation object, which inherits from NSObject, but doesn’t have a length() function, hence the crash.

Since messages can be sent to `nil`, the bridging code can call nil.length() and the app can continue executing without an issue, explaining why this PR works.

### Next Steps
While this PR addresses this issue in this one particular spot, we should audit other Obj-C/Swift boundaries throughout the hybrid SDKs to ensure that this issue isn't present anywhere else.

### DIFF
--- a/ios/purchases_flutter/Sources/purchases_flutter/PurchasesFlutterPlugin.m
+++ b/ios/purchases_flutter/Sources/purchases_flutter/PurchasesFlutterPlugin.m
@@ -344,8 +344,8 @@ signedDiscountTimestamp:(nullable NSString *)discountTimestamp
 presentedOfferingContext:(NSDictionary *)presentedOfferingContext
 signedDiscountTimestamp:(nullable NSString *)discountTimestamp
                  result:(FlutterResult)result {
-    [RCCommonFunctionality purchasePackage:packageIdentifier.mappingNSNullToNil
-                  presentedOfferingContext:presentedOfferingContext.mappingNSNullToNil
+    [RCCommonFunctionality purchasePackage:packageIdentifier
+                  presentedOfferingContext:presentedOfferingContext
                    signedDiscountTimestamp:discountTimestamp.mappingNSNullToNil
                            completionBlock:[self getResponseCompletionBlock:result]];
 }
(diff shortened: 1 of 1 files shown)


joshdholtz (approved):
Looks good to me! :shipit:

ios/purchases_flutter/Sources/purchases_flutter/PurchasesFlutterPlugin.m
```diff
@@ -330,17 +336,17 @@ - (void)purchaseProduct:(NSString *)productIdentifier
...
 }
 
 - (void)purchasePackage:(NSString *)packageIdentifier
 presentedOfferingContext:(NSDictionary *)presentedOfferingContext
 signedDiscountTimestamp:(nullable NSString *)discountTimestamp
                  result:(FlutterResult)result {
-    [RCCommonFunctionality purchasePackage:packageIdentifier
-                  presentedOfferingContext:presentedOfferingContext
-                   signedDiscountTimestamp:discountTimestamp
+    [RCCommonFunctionality purchasePackage:packageIdentifier.mappingNSNullToNil
```
tonidero: This shouldn't be nullable so shouldn't ever happen I think?

fire-at-will: probably not, just wanted to be safe. I'll remove it from `packageIdentifier` and `presentedOfferingContext`

PR #1474, opened 2025-09-16
Add `winBackOffer` parameter to `PurchaseParams`

tonidero:
This also deprecates the existing methods for purchasing win back offers.


lib/purchases_flutter.dart
```diff
@@ -582,17 +583,27 @@ class Purchases {
...
       'customerEmail': customerEmail,
+      'winBackOfferIdentifier': winBackOffer?.identifier,
     };
+    final isWinBackOfferPurchase = (defaultTargetPlatform == TargetPlatform.iOS
+        || defaultTargetPlatform == TargetPlatform.macOS)
+        && winBackOffer != null;
     if (package != null) {
-      return await _invokeReturningPurchaseResult('purchasePackage', {
+      final methodName = isWinBackOfferPurchase
+          ? 'purchasePackageWithWinBackOffer'
```
tonidero, line 593: I thought of changing the native side to reuse the same `purchasePackage` function... But it would just mean moving the logic of which PHC purchase method to use there... Didn't have a strong opinion, but this way I keep the logic shared

lib/purchases_flutter.dart
```diff
@@ -582,17 +583,27 @@ class Purchases {
       'signedDiscountTimestamp': signedDiscountTimestamp,
       'presentedOfferingContext': presentedOfferingContextJson,
       'customerEmail': customerEmail,
+      'winBackOfferIdentifier': winBackOffer?.identifier,
```
tonidero, line 586: I thought we might need to check for nullability in the native side as we did for the signedDiscountTimestamp, but when we use the native purchase methods with winback offers, we already check that we have a value, so it should never be null (the native method also doesn't support null)

lib/models/purchase_params.dart
```diff
@@ -39,7 +42,11 @@ class PurchaseParams {
   /// for more info.
   ///
   /// [promotionalOffer] iOS only. Promotional offer that will be applied to the product.
-  /// Retrieve this offer using [Purchases.getPromotionalOffer].
+  /// Retrieve this offer using [Purchases.getEligibleWinBackOffersForPackage].
```
ajpallares: I think this is incorrect?
```suggestion
  /// Retrieve this offer using [Purchases.getPromotionalOffer].
```

ajpallares: In addition, perhaps I would add
> Ignored if a win-back offer is provided

lib/models/purchase_params.dart
```diff
@@ -39,7 +42,11 @@ class PurchaseParams {
...
+  ///
+  /// [winBackOffer] iOS only. Win-back offer that will be applied to the product.
+  /// Retrieve this offer using [Purchases.getWinBackOffer]. If both a promotional offer
```
ajpallares: I think `Purchases.getWinBackOffer` does not exist. Did you mean `Purchases.getEligibleWinBackOffersForPackage`?
```suggestion
  /// Retrieve this offer using [Purchases.getEligibleWinBackOffersForPackage]. If both a promotional offer
```

lib/models/purchase_params.dart
ajpallares: If so, we should probably add the same phrase to the other `PurchaseParams` constructors docs

ajpallares (commented):
Looks great! Just some comments about documentation

lib/models/purchase_params.dart
tonidero: Totally right, fixed in https://github.com/RevenueCat/purchases-flutter/pull/1474/commits/f92b28fcb863caa01491e9f386efe07b810416ad

lib/models/purchase_params.dart
tonidero: Fixed in https://github.com/RevenueCat/purchases-flutter/pull/1474/commits/f92b28fcb863caa01491e9f386efe07b810416ad

ajpallares (approved):
Great!

lib/purchases_flutter.dart
fire-at-will: Do you mean exposing a `purchase(purchaseParams: PurchaseParams)` function in `RCCommonFunctionality`? That could allow us to eventually get rid of the `purchasePackageWithWinBackOffer` and `purchaseProductWithWinBackOffer` functions in PHC's RCCommonFunctionality and get rid of the logic in the hybrid SDKs that determine which purchase function to call based on the contents of the purchase params.

fire-at-will:
Note: Since we're deprecating the old win-back purchase functions, we should be sure to update the code samples in the docs here: https://www.revenuecat.com/docs/subscription-guidance/subscription-offers/ios-subscription-offers#redeeming-a-win-back-offer-on-your-paywall

Issue #1297, opened 2025-03-18
[Customer Center] presentCustomerCenter() does not work on ios


MaximilianFlechtner:
- [x] I have updated Purchases SDK to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-flutter/blob/main/CONTRIBUTING.md)
- [x] I have searched the [Community](https://community.revenuecat.com)
- [x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [x] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-flutter/issues)

# ‼️ Required data ‼️ #
**Do not remove any of the steps from the template below. If a step is not applicable to your issue, please leave that step empty.**

There are a lot of things that can contribute to things not working. Having a very basic understanding of your environment will help us understand your issue faster!

## Environment ##
- [x] Output of `flutter doctor`
```
[✓] Flutter (Channel stable, 3.27.1, on macOS 15.3.1 24D70 darwin-arm64, locale en-US)
[✗] Android toolchain - develop for Android devices
    ✗ Unable to locate Android SDK.
      Install Android Studio from: https://developer.android.com/studio/index.html
      On first launch it will assist you in installing the Android SDK components.
      (or visit https://flutter.dev/to/macos-android-setup for detailed instructions).
      If the Android SDK has been installed to a custom location, please use
      `flutter config --android-sdk` to update to that location.

[✓] Xcode - develop for iOS and macOS (Xcode 16.2)
[✗] Chrome - develop for the web (Cannot find Chrome executable at /Applications/Google Chrome.app/Contents/MacOS/Google Chrome)
    ! Cannot find Chrome. Try setting CHROME_EXECUTABLE to a Chrome executable.
[!] Android Studio (not installed)
[✓] Connected device (4 available)
[✓] Network resources

! Doctor found issues in 3 categories.
```

- [x] Version of `purchases-flutter` -> `purchases_flutter: ^8.6.1`
- [x] Testing device version e.g.: iOS 15.5, Android API 30, etc. -> `IOS Simulator IOS 18.3`
- [ ] How often the issue occurs- every one of your customers is impacted? Only in dev?
- [ ] [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue
- [ ] Steps to reproduce, with a description of expected vs. actual behavior
[...]

PR #1476, opened 2025-09-22
Add support for customer center view without `presentCustomerCenter` helper

facumenzella:
## Motivation

This PR adds a variant to address https://github.com/RevenueCat/purchases-flutter/issues/1297

Following existing patterns, this PR adds the necessary code to display `CustomerCenterView` without using the helper.

- Register the Customer Center platform view on both iOS and Android
- Adjust the purchase tester sample

| Header | Header |
|--------|--------|
| Screenshot 2025-09-22 at 12 44 04 | Screenshot 2025-09-22 at 12 44 16 |

### DIFF
--- a/purchases_ui_flutter/android/src/main/kotlin/com/revenuecat/purchases_ui_flutter/PurchasesUiFlutterPlugin.kt
+++ b/purchases_ui_flutter/android/src/main/kotlin/com/revenuecat/purchases_ui_flutter/PurchasesUiFlutterPlugin.kt
@@ -10,6 +10,7 @@ import com.revenuecat.purchases.hybridcommon.ui.PaywallSource
 import com.revenuecat.purchases.hybridcommon.ui.PresentPaywallOptions
 import com.revenuecat.purchases.hybridcommon.ui.presentPaywallFromFragment
 import com.revenuecat.purchases.ui.revenuecatui.customercenter.ShowCustomerCenter
+import com.revenuecat.purchases_ui_flutter.views.CustomerCenterViewFactory
 import com.revenuecat.purchases_ui_flutter.views.PaywallFooterViewFactory
 import com.revenuecat.purchases_ui_flutter.views.PaywallViewFactory
 import io.flutter.embedding.android.FlutterFragmentActivity
@@ -44,6 +45,10 @@ class PurchasesUiFlutterPlugin: FlutterPlugin, MethodCallHandler, ActivityAware,
         flutterPluginBinding.platformViewRegistry.registerViewFactory(
             "com.revenuecat.purchasesui/PaywallFooterView",
             PaywallFooterViewFactory(flutterPluginBinding.binaryMessenger)
+        )
+        flutterPluginBinding.platformViewRegistry.registerViewFactory(
+            "com.revenuecat.purchasesui/CustomerCenterView",
+            CustomerCenterViewFactory(flutterPluginBinding.binaryMessenger)
         )
         channel = MethodChannel(flutterPluginBinding.binaryMessenger, "purchases_ui_flutter")
         channel.setMethodCallHandler(this)
--- a/purchases_ui_flutter/android/src/main/kotlin/com/revenuecat/purchases_ui_flutter/views/CustomerCenterView.kt
+++ b/purchases_ui_flutter/android/src/main/kotlin/com/revenuecat/purchases_ui_flutter/views/CustomerCenterView.kt
@@ -2,6 +2,8 @@ package com.revenuecat.purchases_ui_flutter.views
 
 import android.content.Context
 import android.view.View
+import com.revenuecat.purchases.Purchases
+import com.revenuecat.purchases.hybridcommon.ui.CustomerCenterListenerWrapper
 import io.flutter.plugin.common.BinaryMessenger
 import io.flutter.plugin.common.MethodCall
 import io.flutter.plugin.common.MethodChannel
@@ -18,10 +20,14 @@ internal class CustomerCenterView(
 
     private val methodChannel: MethodChannel
     private val nativeCustomerCenterView: NativeCustomerCenterView
+    private val customerCenterListener: CustomerCenterListenerWrapper
 
     init {
         methodChannel = MethodChannel(messenger, "com.revenuecat.purchasesui/CustomerCenterView/$id")
         methodChannel.setMethodCallHandler(this)
+
+        customerCenterListener = createCustomerCenterListener()
+        registerCustomerCenterListener()
 
         nativeCustomerCenterView = NativeCustomerCenterView(context) {
             methodChannel.invokeMethod("onDismiss", null)
@@ -33,10 +39,79 @@ internal class CustomerCenterView(
 
     override fun dispose() {
         methodChannel.setMethodCallHandler(null)
+        unregisterCustomerCenterListener()
     }
 
     override fun onMethodCall(methodCall: MethodCall, result: MethodChannel.Result) {
         result.notImplemented()
+    }
+
+    private fun createCustomerCenterListener(): CustomerCenterListenerWrapper {
+        return object : CustomerCenterListenerWrapper() {
+            override fun onRestoreStartedWrapper() {
+                methodChannel.invokeMethod("onRestoreStarted", null)
+            }
+
+            override fun onRestoreCompletedWrapper(customerInfo: Map<String, Any?>) {
+                methodChannel.invokeMethod("onRestoreCompleted", customerInfo)
+            }
+
+            override fun onRestoreFailedWrapper(error: Map<String, Any?>) {
+                methodChannel.invokeMethod("onRestoreFailed", error)
+            }
+
+            override fun onShowingManageSubscriptionsWrapper() {
+                methodChannel.invokeMethod("onShowingManageSubscriptions", null)
+            }
+
+            override fun onFeedbackSurveyCompletedWrapper(feedbackSurveyOptionId: String) {
+                methodChannel.invokeMethod("onFeedbackSurveyCompleted", feedbackSurveyOptionId)
+            }
+
+            override fun onManagementOptionSelectedWrapper(option: String, url: String?) {
+                methodChannel.invokeMethod(
+                    "onManagementOptionSelected",
+                    mapOf(
+                        "optionId" to option,
+                        "url" to url
+                    )
+                )
+            }
+
+            override fun onManagementOptionSelectedWrapper(
+                option: String,
+                actionIdentifier: String?,
+                purchaseIdentifier: String?
+            ) {
+                methodChannel.invokeMethod(
+                    "onManagementOptionSelected",
+                    mapOf(
+                        "optionId" to option,
+                        "url" to null
+                    )
+                )
+                if (actionIdentifier != null) {
+                    methodChannel.invokeMethod(
+                        "onCustomActionSelected",
+                        mapOf(
+                            "actionId" to actionIdentifier,
+                            "purchaseIdentifier" to purchaseIdentifier
+                        )
+                    )
+                }
+            }
+        }
+    }
+
+    private fun registerCustomerCenterListener() {
+        Purchases.sharedInstance.customerCenterListener = customerCenterListener
+    }
+
+    private fun unregisterCustomerCenterListener() {
+        val purchases = Purchases.sharedInstance
+        if (purchases.customerCenterListener === customerCenterListener) {
+            purchases.customerCenterListener = null
+        }
     }
 }
 
--- a/purchases_ui_flutter/android/src/main/kotlin/com/revenuecat/purchases_ui_flutter/views/CustomerCenterView.kt
+++ b/purchases_ui_flutter/android/src/main/kotlin/com/revenuecat/purchases_ui_flutter/views/CustomerCenterView.kt
@@ -1,0 +1,47 @@
+package com.revenuecat.purchases_ui_flutter.views
+
+import android.content.Context
+import android.view.View
+import io.flutter.plugin.common.BinaryMessenger
+import io.flutter.plugin.common.MethodCall
+import io.flutter.plugin.common.MethodChannel
+import io.flutter.plugin.common.MethodChannel.MethodCallHandler
+import io.flutter.plugin.platform.PlatformView
+import com.revenuecat.purchases.ui.revenuecatui.views.CustomerCenterView as NativeCustomerCenterView
+
+internal class CustomerCenterView(
+    context: Context,
+    id: Int,
+    messenger: BinaryMessenger,
+    creationParams: Map<String?, Any?>,
+) : PlatformView, MethodCallHandler {
+
+    private val methodChannel: MethodChannel
+    private val nativeCustomerCenterView: NativeCustomerCenterView
+
+    init {
+        methodChannel = MethodChannel(messenger, "com.revenuecat.purchasesui/CustomerCenterView/$id")
+        methodChannel.setMethodCallHandler(this)
+
+        nativeCustomerCenterView = NativeCustomerCenterView(context) {
+            methodChannel.invokeMethod("onDismiss", null)
+        }
+        nativeCustomerCenterView.applyCreationParams(creationParams)
+    }
+
+    override fun getView(): View = nativeCustomerCenterView
+
+    override fun dispose() {
+        methodChannel.setMethodCallHandler(null)
+    }
+
+    override fun onMethodCall(methodCall: MethodCall, result: MethodChannel.Result) {
+        result.notImplemented()
+    }
+}
+
+@Suppress("UNUSED_PARAMETER")
+private fun NativeCustomerCenterView.applyCreationParams(creationParams: Map<String?, Any?>) {
+    // Currently there are no Android specific configuration options.
+    // This extension is reserved for future parameters to keep init tidy.
+}
--- a/purchases_ui_flutter/ios/purchases_ui_flutter/Sources/purchases_ui_flutter/PurchasesUiCustomerCenterView.swift
+++ b/purchases_ui_flutter/ios/purchases_ui_flutter/Sources/purchases_ui_flutter/PurchasesUiCustomerCenterView.swift
@@ -1,0 +1,171 @@
+import Flutter
+import UIKit
+import PurchasesHybridCommonUI
+@_spi(Internal) import RevenueCatUI
+import RevenueCat
+
+final class PurchasesUiCustomerCenterViewFactory: NSObject, FlutterPlatformViewFactory {
+    private let messenger: FlutterBinaryMessenger
+
+    init(messenger: FlutterBinaryMessenger) {
+        self.messenger = messenger
+        super.init()
+    }
+
+    func create(
+        withFrame frame: CGRect,
+        viewIdentifier viewId: Int64,
+        arguments args: Any?
+    ) -> FlutterPlatformView {
+        if #available(iOS 15.0, *) {
+            return PurchasesUiCustomerCenterView(
+                frame: frame,
+                viewIdentifier: viewId,
+                arguments: args,
+                binaryMessenger: messenger
+            )
+        } else {
+            print("Error: attempted to present customer center on unsupported iOS version.")
+            return UnsupportedPlatformView()
+        }
+    }
+
+    func createArgsCodec() -> FlutterMessageCodec & NSObjectProtocol {
+        FlutterStandardMessageCodec.sharedInstance()
+    }
+}
+
+@available(iOS 15.0, *)
+final class PurchasesUiCustomerCenterView: NSObject, FlutterPlatformView {
+    private let methodChannel: FlutterMethodChannel
+    private let viewController: CustomerCenterUIViewController
+    private let wrapperView: ViewControllerWrapper<CustomerCenterUIViewController>
+
+    init(
+        frame: CGRect,
+        viewIdentifier viewId: Int64,
+        arguments args: Any?,
+        binaryMessenger messenger: FlutterBinaryMessenger
+    ) {
+        methodChannel = FlutterMethodChannel(
+            name: "com.revenuecat.purchasesui/CustomerCenterView/\(viewId)",
+            binaryMessenger: messenger
+        )
+
+        viewController = CustomerCenterUIViewController()
+        wrapperView = ViewControllerWrapper(viewController: viewController)
+
+        super.init()
+
+        setupMethodChannel()
+        configureViewController(arguments: args)
+    }
+
+    func view() -> UIView {
+        wrapperView
+    }
+
+    private func setupMethodChannel() {
+        methodChannel.setMethodCallHandler { _, result in
+            result(FlutterMethodNotImplemented)
+        }
+    }
+
+    private func configureViewController(arguments args: Any?) {
+        viewController.delegate = self
+
+        if let arguments = args as? [String: Any],
+           let shouldShowCloseButton = arguments["shouldShowCloseButton"] as? Bool {
+            viewController.shouldShowCloseButton = shouldShowCloseButton
+        }
+
+        viewController.onCloseHandler = { [weak self] in
+            self?.methodChannel.invokeMethod("onDismiss", arguments: nil)
+        }
+    }
+}
+
+@available(iOS 15.0, *)
+extension PurchasesUiCustomerCenterView: CustomerCenterViewControllerDelegateWrapper {
+    func customerCenterViewControllerWasDismissed(_ controller: CustomerCenterUIViewController) {
+        methodChannel.invokeMethod("onDismiss", arguments: nil)
+    }
+
+    func customerCenterViewControllerDidStartRestore(_ controller: CustomerCenterUIViewController) {
+        methodChannel.invokeMethod("onRestoreStarted", arguments: nil)
+    }
+
+    func customerCenterViewController(
+        _ controller: CustomerCenterUIViewController,
+        didFinishRestoringWith customerInfoDictionary: [String : Any]
+    ) {
+        methodChannel.invokeMethod("onRestoreCompleted", arguments: customerInfoDictionary)
+    }
+
+    func customerCenterViewController(
+        _ controller: CustomerCenterUIViewController,
+        didFailRestoringWith errorDictionary: [String : Any]
+    ) {
+        methodChannel.invokeMethod("onRestoreFailed", arguments: errorDictionary)
+    }
+
+    func customerCenterViewControllerDidShowManageSubscriptions(_ controller: CustomerCenterUIViewController) {
+        methodChannel.invokeMethod("onShowingManageSubscriptions", arguments: nil)
+    }
+
+    func customerCenterViewController(
+        _ controller: CustomerCenterUIViewController,
+        didStartRefundRequestForProductWithID productID: String
+    ) {
+        methodChannel.invokeMethod("onRefundRequestStarted", arguments: productID)
+    }
+
+    func customerCenterViewController(
+        _ controller: CustomerCenterUIViewController,
+        didCompleteRefundRequestForProductWithID productId: String,
+        withStatus status: String
+    ) {
+        methodChannel.invokeMethod(
+            "onRefundRequestCompleted",
+            arguments: [
+                "productId": productId,
+                "status": status
+            ]
+        )
+    }
+
+    func customerCenterViewController(
+        _ controller: CustomerCenterUIViewController,
+        didSelectCustomerCenterManagementOption optionID: String,
+        withURL url: String?
+    ) {
+        methodChannel.invokeMethod(
+            "onManagementOptionSelected",
+            arguments: [
+                "optionId": optionID,
+                "url": url as Any
+            ]
+        )
+    }
+
+    func customerCenterViewController(
+        _ controller: CustomerCenterUIViewController,
+        didCompleteFeedbackSurveyWithOptionID optionID: String
+    ) {
+        methodChannel.invokeMethod("onFeedbackSurveyCompleted", arguments: optionID)
+    }
+
+    func customerCenterViewController(
+        _ controller: CustomerCenterUIViewController,
+        didSelectCustomAction actionID: String,
+        withPurchaseIdentifier purchaseIdentifier: String?
+    ) {
+        methodChannel.invokeMethod(
+            "onCustomActionSelected",
+            arguments: [
+                "actionId": actionID,
+                "purchaseIdentifier": purchaseIdentifier as Any
+            ]
+        )
+    }
+}
--- a/purchases_ui_flutter/ios/purchases_ui_flutter/Sources/purchases_ui_flutter/PurchasesUiFlutterPlugin.swift
+++ b/purchases_ui_flutter/ios/purchases_ui_flutter/Sources/purchases_ui_flutter/PurchasesUiFlutterPlugin.swift
@@ -19,9 +19,11 @@ public class PurchasesUiFlutterPlugin: NSObject, FlutterPlugin {
         let messenger = registrar.messenger()
         let factory = PurchasesUiPaywallViewFactory(messenger: messenger)
         let footerFactory = PurchasesUiPaywallFooterViewFactory(messenger: messenger)
+        let customerCenterFactory = PurchasesUiCustomerCenterViewFactory(messenger: messenger)
 
         registrar.register(factory, withId: "com.revenuecat.purchasesui/PaywallView")
         registrar.register(footerFactory, withId: "com.revenuecat.purchasesui/PaywallFooterView")
+        registrar.register(customerCenterFactory, withId: "com.revenuecat.purchasesui/CustomerCenterView")
 
 #endif
         let channel = FlutterMethodChannel(name: "purchases_ui_flutter", binaryMessenger: messenger)
--- a/purchases_ui_flutter/lib/purchases_ui_flutter.dart
+++ b/purchases_ui_flutter/lib/purchases_ui_flutter.dart
@@ -6,6 +6,7 @@ import 'paywall_result.dart';
 export 'paywall_result.dart';
 export 'views/paywall_footer_view.dart';
 export 'views/paywall_view.dart';
+export 'views/customer_center_view.dart';
 
 class RevenueCatUI {
   static const _methodChannel = MethodChannel('purchases_ui_flutter');
--- a/purchases_ui_flutter/lib/views/customer_center_view.dart
+++ b/purchases_ui_flutter/lib/views/customer_center_view.dart
@@ -1,0 +1,107 @@
+import 'dart:io';
+
+import 'package:flutter/foundation.dart';
+import 'package:flutter/gestures.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter/rendering.dart';
+import 'package:flutter/services.dart';
+
+import 'customer_center_view_method_handler.dart';
+
+class CustomerCenterView extends StatelessWidget {
+  const CustomerCenterView({
+    super.key,
+    this.shouldShowCloseButton = true,
+    this.onDismiss,
+    this.onRestoreStarted,
+    this.onRestoreCompleted,
+    this.onRestoreFailed,
+    this.onShowingManageSubscriptions,
+    this.onRefundRequestStarted,
+    this.onRefundRequestCompleted,
+    this.onFeedbackSurveyCompleted,
+    this.onManagementOptionSelected,
+    this.onCustomActionSelected,
+  });
+
+  final bool shouldShowCloseButton;
+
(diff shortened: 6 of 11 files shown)


revenuecat_examples/purchase_tester/lib/main.dart
```diff
@@ -17,14 +17,14 @@ void main() async {
   } else if (Platform.isIOS || Platform.isMacOS) {
     StoreConfig(
       store: Store.appStore,
-      apiKey: appleApiKey,
+      apiKey: "appl_KFpRDwauDyfIGIETKyVymslMcjX",
```
facumenzella:
```suggestion
      apiKey: appleApiKey,
```

revenuecat_examples/purchase_tester/lib/main.dart
```diff
@@ -17,14 +17,14 @@ void main() async {
...
     );
   } else if (Platform.isAndroid) {
     // Run the app passing --dart-define=AMAZON=true
     const useAmazon = bool.fromEnvironment("amazon");
     StoreConfig(
       store: useAmazon ? Store.amazon : Store.playStore,
-      apiKey: useAmazon ? amazonApiKey : googleApiKey,
+      apiKey: useAmazon ? amazonApiKey : "goog_cRtaFpfwfcQYPXkIEqFUkVszUfN",
```
facumenzella:
```suggestion
      apiKey: useAmazon ? amazonApiKey : googleApiKey,
```

vegaro:
[butitcanbebetter](https://app.graphite.dev/settings/meme-library?org=RevenueCat)

purchases_ui_flutter/android/src/main/kotlin/com/revenuecat/purchases_ui_flutter/views/CustomerCenterView.kt
```diff
@@ -0,0 +1,122 @@
...
+
+    private val methodChannel: MethodChannel
+    private val nativeCustomerCenterView: NativeCustomerCenterView
+    private val customerCenterListener: CustomerCenterListenerWrapper
+
+    init {
+        methodChannel = MethodChannel(messenger, "com.revenuecat.purchasesui/CustomerCenterView/$id")
+        methodChannel.setMethodCallHandler(this)
+
+        customerCenterListener = createCustomerCenterListener()
```
tonidero: Hmm can we keep the same pattern we have in the native SDK and keep the listener as a global instance on `Purchases`? I would try to avoid linking the listener to the instance of CustomerCenterView. Then you also need to deal with edge cases like having multiple instances of the CustomerCenterView or possible memory leaks.

tonidero: Alternatively, we could try to add a new parameter to the `CustomerCenterView` for the listener, so we can pass the listener to the instance of the view, instead of a global instance as we're using right now.

facumenzella: I've been reading about this, and apparently *Flutter can create and dispose platform views*. 

We can have a new CustomerCenterView install its listener before the previous instance’s dispose() runs.  If we null the global property in dispose() we’ll delete the listener the new instance just set, breaking callbacks.

facumenzella: If we use the same id for `methodChannel` it could break the communication, because the first view would be closing the channel after the first one is created 🤔

facumenzella: I guess the main issue is that we don't expose a per-view listener, like paywalls

tonidero: > I guess the main issue is that we don't expose a per-view listener, like paywalls

I agree, I think that's the main issue... And I would say it shouldn't be too hard to add right? We can add an optional parameter in the `CustomerCenterView` [here](https://github.com/RevenueCat/purchases-android/blob/a664003c7e3d25720f6b8b938c1b9a0a0b972230/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/views/CustomerCenterView.kt#L28) that is a listener that can be passed in to Compose's `CustomerCenterOptions`. Then we would be using that for this new API. Wdyt?

facumenzella: I agree. I'll jump to that then 🐇

vegaro: This can be removed now then?

purchases_ui_flutter/android/src/main/kotlin/com/revenuecat/purchases_ui_flutter/views/CustomerCenterView.kt
```diff
@@ -0,0 +1,116 @@
...
+
+    override fun getView(): View = nativeCustomerCenterView
+
+    override fun dispose() {
+        methodChannel.setMethodCallHandler(null)
+        // Only clear if it's still our listener
+        val purchases = Purchases.sharedInstance
+        if (purchases.customerCenterListener === customerCenterListener) {
+            purchases.customerCenterListener = null
+        }
```
vegaro: same, this can be removed now I believe

purchases_ui_flutter/android/src/main/kotlin/com/revenuecat/purchases_ui_flutter/views/CustomerCenterView.kt
facumenzella: It should be removable once we merge https://github.com/RevenueCat/purchases-flutter/pull/1502

purchases_ui_flutter/android/src/main/kotlin/com/revenuecat/purchases_ui_flutter/views/CustomerCenterView.kt
```diff
@@ -0,0 +1,116 @@
...
+                    )
+                }
+            }
+        }
+    }
+
+}
+
+@Suppress("UNUSED_PARAMETER")
+private fun NativeCustomerCenterView.applyCreationParams(creationParams: Map<String?, Any?>) {
```
vegaro: should we remove this if it's empty?

api_tester/lib/api_tests/purchases_ui_flutter_api_test.dart
```diff
@@ -196,4 +196,108 @@ class _PurchasesFlutterApiTest {
...
+  }
+
+  void _checkCustomerCenterEventStream() {
+    Stream<CustomerCenterEvent> stream = RevenueCatUI.customerCenterEventStream;
+    
+    // Test event stream subscription
+    stream.listen((CustomerCenterEvent event) {
+      // Test event type checking
+      switch (event.runtimeType) {
+        case CustomerCenterOnDismissEvent:
```
vegaro: I am unsure if these are correct. I don't see `CustomerCenterOnDismissEvent`, `CustomerCenterOnRestoreStartedEvent`, etc anywhere in the code

purchases_ui_flutter/lib/purchases_ui_flutter.dart
```diff
@@ -1,14 +1,120 @@
...
+// Default empty callbacks
+void _emptyCallback() {}
+void _emptyStringCallback(String _) {}
+void _emptyDataCallback(Map<String, dynamic> _) {}
+
+/// Container for CustomerCenter callbacks
+class CustomerCenterCallbacks {
+  final VoidCallback onDismiss;
+  final VoidCallback onRestoreStarted;
+  final Function(Map<String, dynamic>) onRestoreCompleted;
```
vegaro: Is there a way we can type what these callbacks return? For example, could we have here `Function<CustomerInfo>`?

To give an example, in the `customerInfoUpdateListener` we convert from Map to CustomerInfo https://github.com/RevenueCat/purchases-flutter/blob/e00f2fd7c1e08bb12998e2e6102c824846ec49ec/lib/purchases_flutter.dart#L63

purchases_ui_flutter/lib/purchases_ui_flutter.dart
```diff
@@ -75,4 +219,90 @@ class RevenueCatUI {
         return PaywallResult.error;
     }
   }
+
+  static void _handleCustomerCenterMethodCall(MethodCall call) {
```
vegaro: I think we could join `_handleCustomerCenterMethodCall` and `_parseCustomerCenterMethodCall` since they do a switch on the same thing and have the joined function trigger the callback but also return the constructed event for the `customerCenterEventStream`

purchases_ui_flutter/lib/purchases_ui_flutter.dart
vegaro: Actually, I see you are doing that for the view in `customer_center_view_method_handler.dart`, I think you can be doing the same here.

Also, I believe that would allow you to just use the same callbacks you use for the View here (example `CustomerCenterRestoreCompletedCallback`) here right?

purchases_ui_flutter/lib/purchases_ui_flutter.dart
```diff
@@ -75,4 +219,90 @@ class RevenueCatUI {
...
+        return CustomerCenterEvent(
+          CustomerCenterEventType.restoreFailed, 
+          call.arguments as Map<String, dynamic>?,
+        );
+      case 'onShowingManageSubscriptions':
+        return const CustomerCenterEvent(CustomerCenterEventType.showingManageSubscriptions);
+      case 'onRefundRequestStarted':
+        return CustomerCenterEvent(
+          CustomerCenterEventType.refundRequestStarted, 
+          {'productIdentifier': call.arguments},
```
vegaro: The fact that some come already as a dictionary and some don't seems a bit problematic. I wonder if we should be sending this map directly from the native layers, so we can always assume the dictionaries are prepared in Swift or Android and not have to build it here

facumenzella: Indeed. Moving this to the native layer would simplify the work on this side for sure.

purchases_ui_flutter/lib/purchases_ui_flutter.dart
```diff
@@ -75,4 +152,56 @@ class RevenueCatUI {
...
+        callbacks?.onShowingManageSubscriptions?.call();
+        break;
+      case 'onRefundRequestStarted':
+        final productIdentifier = call.arguments as String? ?? '';
+        callbacks?.onRefundRequestStarted?.call(productIdentifier);
+        break;
+      case 'onRefundRequestCompleted':
+        final data = call.arguments as Map<String, dynamic>? ?? {};
+        final productIdentifier = data['productId'] as String? ?? '';
+        final status = data['status'] as String? ?? '';
```
vegaro: could these parameters be casted to non optional `String` to avoid the `''`? This applies to the other cases as well where you are casting to optional

purchases_ui_flutter/lib/purchases_ui_flutter.dart
```diff
@@ -52,11 +91,49 @@ class RevenueCatUI {
...
+    await _methodChannel.invokeMethod('presentCustomerCenter');
+  }
+
+  static void _ensureMethodChannelHandler() {
+    if (!_methodChannelHandlerSet) {
+      _methodChannel.setMethodCallHandler((call) async {
+        try {
+          _handleCustomerCenterMethodCall(call);
+        } catch (e) {
+          // Silently ignore callback errors
```
vegaro: Maybe we should log actually?

revenuecat_examples/purchase_tester/lib/src/upsell.dart
```diff
@@ -32,6 +33,7 @@ class _UpsellScreenState extends State<UpsellScreen> {
     _fetchData();
   }
 
+
```
vegaro: nitpick but I think this is an extra space

vegaro:
```suggestion
```

facumenzella:
@tonidero @vegaro would you mind taking another look :)

api_tester/lib/api_tests/purchases_ui_flutter_api_test.dart
```diff
@@ -196,4 +196,47 @@ class _PurchasesFlutterApiTest {
...
+      onDismiss: () {},
+      onRestoreStarted: () {},
+      onRestoreCompleted: (CustomerInfo customerInfo) {},
+      onRestoreFailed: (PurchasesError error) {},
+      onShowingManageSubscriptions: () {},
+      onRefundRequestStarted: (String productIdentifier) {},
+      onRefundRequestCompleted: (String productIdentifier, String status) {},
+      onFeedbackSurveyCompleted: (String optionIdentifier) {},
+      onManagementOptionSelected: (String optionIdentifier, String? url) {},
+      onCustomActionSelected: (String actionIdentifier, String? purchaseIdentifier) {},
```
tonidero, line 213: Hmm would it be worth encapsulating all these listeners into another "CustomerCenterListener" parameter? I'm afraid of the top level parameters getting too confusing with so many parameters.

purchases_ui_flutter/android/src/main/kotlin/com/revenuecat/purchases_ui_flutter/PurchasesUiFlutterPlugin.kt
```diff
@@ -149,6 +156,10 @@ class PurchasesUiFlutterPlugin: FlutterPlugin, MethodCallHandler, ActivityAware,
     }
 
     private fun presentCustomerCenterFromActivity(activity: Activity) {
+        // Set up customer center listener for callbacks before presenting
+        val customerCenterListener = createCustomerCenterListener()
+        Purchases.sharedInstance.customerCenterListener = customerCenterListener
```
tonidero: Hmm I wonder, would it make sense to add this same API to dart, so we follow the same API/architecture at that level? It could avoid potential issues of the customerCenterListener not getting deallocated correctly... This would affect iOS as well though, so we might need to add  a similar API there.

purchases_ui_flutter/ios/purchases_ui_flutter/Sources/purchases_ui_flutter/PurchasesUiCustomerCenterView.swift
```diff
@@ -0,0 +1,176 @@
...
+
+    func customerCenterViewControllerDidShowManageSubscriptions(_ controller: CustomerCenterUIViewController) {
+        methodChannel.invokeMethod("onShowingManageSubscriptions", arguments: nil)
+    }
+
+    func customerCenterViewController(
+        _ controller: CustomerCenterUIViewController,
+        didStartRefundRequestForProductWithID productID: String
+    ) {
+        methodChannel.invokeMethod("onRefundRequestStarted", arguments: productID)
```
tonidero: Hmm we're passing a string here instead of a dictionary. Should we pass a dictionary to provide more flexibility in the future if needed? I guess this is internal so we can change it later, but I think it's probably a good thing to do

purchases_ui_flutter/ios/purchases_ui_flutter/Sources/purchases_ui_flutter/PurchasesUiCustomerCenterView.swift
```diff
@@ -0,0 +1,176 @@
...
+                "url": url as Any
+            ]
+        )
+    }
+
+    func customerCenterViewController(
+        _ controller: CustomerCenterUIViewController,
+        didCompleteFeedbackSurveyWithOptionID optionID: String
+    ) {
+        methodChannel.invokeMethod("onFeedbackSurveyCompleted", arguments: optionID)
```
tonidero: Same here about passing a dictionary.

purchases_ui_flutter/ios/purchases_ui_flutter/Sources/purchases_ui_flutter/PurchasesUiFlutterPlugin.swift
```diff
@@ -199,3 +219,69 @@ private extension PurchasesUiFlutterPlugin {
...
+    func customerCenterViewController(_ controller: CustomerCenterUIViewController, didFailRestoringWith errorDictionary: [String: Any]) {
+        methodChannel?.invokeMethod("onRestoreFailed", arguments: errorDictionary)
+    }
+    
+    func customerCenterViewControllerDidShowManageSubscriptions(_ controller: CustomerCenterUIViewController) {
+        methodChannel?.invokeMethod("onShowingManageSubscriptions", arguments: nil)
+    }
+    
+    func customerCenterViewController(_ controller: CustomerCenterUIViewController, didStartRefundRequestForProductWithID productID: String) {
+        methodChannel?.invokeMethod("onRefundRequestStarted", arguments: productID)
```
tonidero: Same here about passing a dictionary

purchases_ui_flutter/test/purchases_ui_flutter_test.dart
```diff
@@ -195,4 +211,306 @@ void main() {
...
+    const viewWithCloseButton = CustomerCenterView(shouldShowCloseButton: true);
+    const viewWithoutCloseButton = CustomerCenterView(shouldShowCloseButton: false);
+    
+    expect(viewWithCloseButton.shouldShowCloseButton, true);
+    expect(viewWithoutCloseButton.shouldShowCloseButton, false);
+    
+    // Note: Actual behavior differs by platform:
+    // - iOS: respects the shouldShowCloseButton parameter
+    // - Android: always shows close button regardless of parameter value
+    // This is documented in the shouldShowCloseButton property documentation
```
tonidero, line 357: Hmm feels adding this documentation to the test doesn't help as much... maybe we can remove it? NABD though.

facumenzella: Well, I figured this won't hurt when revisiting the test 😓

api_tester/lib/api_tests/purchases_ui_flutter_api_test.dart
facumenzella: I've added a listener API instead of standalone callbacks :)

purchases_ui_flutter/android/src/main/kotlin/com/revenuecat/purchases_ui_flutter/PurchasesUiFlutterPlugin.kt
```diff
@@ -184,10 +205,87 @@ class PurchasesUiFlutterPlugin: FlutterPlugin, MethodCallHandler, ActivityAware,
...
+                    mapOf(
+                        "actionId" to actionId,
+                        "purchaseIdentifier" to purchaseIdentifier
+                    )
+                )
+            }
+        }
+    }
+
+    private fun ensureCustomerCenterListener() {
```
tonidero: Hmm the name for this one is weird... maybe `setCustomerCenterListenerIfNeeded`?

purchases_ui_flutter/lib/purchases_ui_flutter.dart
```diff
@@ -52,11 +91,57 @@ class RevenueCatUI {
...
+    CustomerCenterDismissed? onDismiss,
+    CustomerCenterRestoreStarted? onRestoreStarted,
+    CustomerCenterRestoreCompleted? onRestoreCompleted,
+    CustomerCenterRestoreFailed? onRestoreFailed,
+    CustomerCenterManageSubscriptions? onShowingManageSubscriptions,
+    CustomerCenterRefundRequestStarted? onRefundRequestStarted,
+    CustomerCenterRefundRequestCompleted? onRefundRequestCompleted,
+    CustomerCenterFeedbackSurveyCompleted? onFeedbackSurveyCompleted,
+    CustomerCenterManagementOptionSelected? onManagementOptionSelected,
+    CustomerCenterCustomActionSelected? onCustomActionSelected,
```
tonidero, line 85: Oh sorry... What I meant in my previous comment was to add a new global API in Flutter like `RevenueCatUI.setCustomerCenterListener` that keeps a reference to the listener globally, using Android's `Purchases.sharedInstance.customerCenterListener` (for iOS we would need to implement it), instead of passing the listeners to the instance like this. The idea is to have the same API in hybrids for the `presentPaywall` option like we do in Android, since there exist this limitation. It's a bigger change, but I think it might be safer. Wdyt @facumenzella @vegaro ?

purchases_ui_flutter/lib/purchases_ui_flutter.dart
```diff
@@ -75,4 +160,191 @@ class RevenueCatUI {
...
+        listener?.onDismiss?.call();
+        // Clear the listener registration to avoid retaining the plugin
+        await _clearCustomerCenterCallbacks();
+        break;
+      case 'onRestoreStarted':
+        listener?.onRestoreStarted?.call();
+        break;
+      case 'onRestoreCompleted':
+        final rawArguments = call.arguments;
+        if (rawArguments is! Map) {
```
JZDesign, line 198: New to dart, but I believe this check is essentially `if rawArguments.type == Map? && rawArguments != null` but I'm wondering if that is what you meant. Based on what comes next, it looks like you meant `if rawArguments.type != Map`

If so, I think you want to write this differently.

facumenzella: Potentially yes, let me revisit this. 😅

JZDesign: NVM

JZDesign: I read the docs wrong. this is right.

JZDesign: Found a table that made it clear after I posted this comment https://dart.dev/language/operators#type-test-operators

purchases_ui_flutter/lib/purchases_ui_flutter.dart
facumenzella: No problem :) 

Just to be clear on this:
- `RevenueCatUI.presentCustomerCenter` does not receive a listener. This must be set using the static API
- The standalone view does receive a listener.

Is this what you mean?

facumenzella: I can do it, it's just that in RN we've just merged a different implementation. I think that I could offer the static setter as well, and if people want to use it they can just pass nill for the per-view listener?

vegaro: > The idea is to have the same API in hybrids for the presentPaywall option like we do in Android, since there exist this limitation.

In Android we did it that way because of how the Activity communicates back the events, but here in the hybrids we could camouflage that and actually have them as in iOS, right? At least that's what we did in React Native

tonidero: >  in the hybrids we could camouflage that 

I mean, we can certainly camouflage it, but we could also have camouflaged it in Android itself. By using these APIs in this way we again run into the same issue that we could run in android of possibly causing memory leaks by holding objects statically (kinda, as part of the purchases singleton instance) related to something that has a specific lifetime, like presenting CustomerCenter. As in, the listener might not be reset to null after presenting the customer center ends.

TBH, I don't have specific repro steps for this, but in my opinion, it's not great to use static holders like this, which is why we designed the Android API like that in the first place.

Having said all this, I don't have repro steps for this issue, and I didn't notice we had already done it like this in RN... so I could be convinced to keep it this way. But we should be aware, we will have to maintain this API moving forward and monitor for possible issues related to this.

tonidero: > RevenueCatUI.presentCustomerCenter does not receive a listener. This must be set using the static API
The standalone view does receive a listener.

That's exactly what I meant, but happy to discuss if we think we should keep it the current way.

facumenzella: Ok I'll summarize here our conversation:
- The CustomerCenter callback are mixed with UI callbacks (onDismiss, onCustomAction) and some that are not necessarily UI (onRestoreStarted). The non-UI callbacks naturally could be static and global, but the other ones can't, but the other ones can't
- We'll stick with callbacks for now.
- We should be extra careful with memory leaks
We'll document this to keep it in mind for future iterations

purchases_ui_flutter/ios/purchases_ui_flutter/Sources/purchases_ui_flutter/PurchasesUiFlutterPlugin.swift
```diff
@@ -114,6 +136,37 @@ public class PurchasesUiFlutterPlugin: NSObject, FlutterPlugin {
         }
     }
 
+    private func setCustomerCenterCallbacks(
+        _ result: @escaping FlutterResult
+    ) {
+#if os(iOS)
+        if #available(iOS 15.0, *) {
+            self.ensureCustomerCenterDelegateForwarder()
```
facumenzella: This could be renamed to `setIfNeeded` too

purchases_ui_flutter/lib/purchases_ui_flutter.dart
```diff
@@ -52,11 +69,61 @@ class RevenueCatUI {
...
+        onShowingManageSubscriptions: onShowingManageSubscriptions,
+        onRefundRequestStarted: onRefundRequestStarted,
+        onRefundRequestCompleted: onRefundRequestCompleted,
+        onFeedbackSurveyCompleted: onFeedbackSurveyCompleted,
+        onManagementOptionSelected: onManagementOptionSelected,
+        onCustomActionSelected: onCustomActionSelected,
+      );
+    } else if (_hasRegisteredCustomerCenterCallbacks) {
+      await _clearCustomerCenterCallbacks();
+    }
```
vegaro, line 112: I think it would be easier to understand to always `_clearCustomerCenterCallbacks` at the beginning of this function, and then register

purchases_ui_flutter/lib/purchases_ui_flutter.dart
```diff
@@ -75,4 +142,207 @@ class RevenueCatUI {
...
+    _storeCustomerCenterCallbacks(
+      onRestoreStarted: onRestoreStarted,
+      onRestoreCompleted: onRestoreCompleted,
+      onRestoreFailed: onRestoreFailed,
+      onShowingManageSubscriptions: onShowingManageSubscriptions,
+      onRefundRequestStarted: onRefundRequestStarted,
+      onRefundRequestCompleted: onRefundRequestCompleted,
+      onFeedbackSurveyCompleted: onFeedbackSurveyCompleted,
+      onManagementOptionSelected: onManagementOptionSelected,
+      onCustomActionSelected: onCustomActionSelected,
```
vegaro: I think that we can remove `_storeCustomerCenterCallbacks` and inline its content here, since `_storeCustomerCenterCallbacks` is only called here. Also, removing `_storeCustomerCenterCallbacks` can make this file a bit shorter

purchases_ui_flutter/lib/purchases_ui_flutter.dart
```diff
@@ -75,4 +142,207 @@ class RevenueCatUI {
...
+    );
+    await _methodChannel.invokeMethod('setCustomerCenterCallbacks');
+  }
+
+  static Future<void> _clearCustomerCenterCallbacks() async {
+    final hadRegisteredCallbacks = _hasRegisteredCustomerCenterCallbacks;
```
vegaro: You do `    _ensureMethodChannelHandler();` in `_storeCustomerCenterCallbacks`, but not in here

purchases_ui_flutter/android/src/main/kotlin/com/revenuecat/purchases_ui_flutter/PurchasesUiFlutterPlugin.kt
```diff
@@ -184,10 +205,87 @@ class PurchasesUiFlutterPlugin: FlutterPlugin, MethodCallHandler, ActivityAware,
...
+        }
+    }
+
     override fun onActivityResult(requestCode: Int, resultCode: Int, data: Intent?): Boolean {
         if (requestCode == REQUEST_CODE_CUSTOMER_CENTER) {
             if (resultCode == Activity.RESULT_OK) {
                 Log.d(TAG, "Customer Center closed successfully")
+                // Send onDismiss callback
+                channel.invokeMethod("onDismiss", null)
                 pendingResult?.success("Customer Center closed successfully")
```
vegaro, line 289: I might be wrong here, I am not 100% sure, but will this success fire and forget, and continue and `clearCustomerCenterListener` or will it complete the Dart promise, meaning that the developer's code will execute and the `clearCustomerCenterListener` will possible never get called?

facumenzella: Just confirmed that it works as expected. The listener is always torn down even if Dart handlers take time.

revenuecat_examples/purchase_tester/lib/src/upsell.dart
```diff
@@ -170,9 +171,39 @@ class _UpsellScreenState extends State<UpsellScreen> {
...
+                            debugPrint('[CustomerCenter Modal] Refund request started for product: $productId'),
+                        onRefundRequestCompleted: (productId, status) =>
+                            debugPrint('[CustomerCenter Modal] Refund request completed for product $productId with status $status'),
+                        onFeedbackSurveyCompleted: (optionId) =>
+                            debugPrint('[CustomerCenter Modal] Feedback survey completed with option: $optionId'),
+                        onManagementOptionSelected: (optionId, url) =>
+                            debugPrint('[CustomerCenter Modal] Management option selected: $optionId (url: ${url ?? 'none'})'),
+                        onCustomActionSelected: (actionId, purchaseIdentifier) =>
+                            debugPrint('[CustomerCenter Modal] Custom action selected: $actionId (purchase: ${purchaseIdentifier ?? 'none'})'),
+                      );
```
vegaro, line 206: super nit but maybe adding a "debugPrint('[CustomerCenter] Dismissed');" after the await would be nice to make sure that's behaving as expected :)

purchases_ui_flutter/android/src/main/kotlin/com/revenuecat/purchases_ui_flutter/PurchasesUiFlutterPlugin.kt
```diff
@@ -184,10 +205,87 @@ class PurchasesUiFlutterPlugin: FlutterPlugin, MethodCallHandler, ActivityAware,
...
+                    mapOf(
+                        "actionId" to actionId,
+                        "purchaseIdentifier" to purchaseIdentifier
+                    )
+                )
+            }
+        }
+    }
+
+    private fun setCustomerCenterListenerIfNeeded() {
```
vegaro: The `ifNeeded` part is a bit confusing, because it's actually always setting it right? I think it would be more like 
`createCustomerCenterListenerIfNeeded` or maybe we could do the `Purchases.sharedInstance.customerCenterListener = customerCenterListener` in the `if` as well. Why is it done after the if check?

purchases_ui_flutter/android/src/main/kotlin/com/revenuecat/purchases_ui_flutter/PurchasesUiFlutterPlugin.kt
```diff
@@ -184,10 +205,87 @@ class PurchasesUiFlutterPlugin: FlutterPlugin, MethodCallHandler, ActivityAware,
...
+            Purchases.sharedInstance.customerCenterListener = null
+        }
+    }
+
     override fun onActivityResult(requestCode: Int, resultCode: Int, data: Intent?): Boolean {
         if (requestCode == REQUEST_CODE_CUSTOMER_CENTER) {
             if (resultCode == Activity.RESULT_OK) {
                 Log.d(TAG, "Customer Center closed successfully")
+                // Send onDismiss callback
+                channel.invokeMethod("onDismiss", null)
```
vegaro, line 288: what do you think of changing `onDismiss` to `clearCustomerCenterCallbacks`? so it's less confusing

purchases_ui_flutter/ios/purchases_ui_flutter/Sources/purchases_ui_flutter/PurchasesUiCustomerCenterView.swift
```diff
@@ -0,0 +1,186 @@
...
+    private func configureViewController(arguments args: Any?) {
+        viewController.delegate = self
+
+        if let arguments = args as? [String: Any],
+           let shouldShowCloseButton = arguments["shouldShowCloseButton"] as? Bool {
+            viewController.shouldShowCloseButton = shouldShowCloseButton
+        }
+
+        viewController.onCloseHandler = { [weak self] in
+            self?.methodChannel.invokeMethod("onDismiss", arguments: nil)
```
vegaro: will this and `customerCenterViewControllerWasDismissed` trigger `onDismiss` twice?

purchases_ui_flutter/lib/views/customer_center_view.dart
```diff
@@ -0,0 +1,118 @@
...
+  /// Whether to show a close button in the customer center.
+  /// 
+  /// **Platform Support:**
+  /// - **iOS**: Fully supported. When `false`, no close button is shown.
+  /// - **Android**: Not supported. Close button is always shown regardless of this setting.
+  ///
+  /// Defaults to `true`.
+  final bool shouldShowCloseButton;
+
+  final CustomerCenterDismissed? onDismiss;
```
vegaro: yeah I think this can lead to confusion, the `onDismiss` might seem like the button was pressed and the `Dismissed` like it was already dismissed

purchases_ui_flutter/lib/views/customer_center_view.dart
vegaro, line 1: One thing to test here is rotating the device and see if it resizes correctly. We had issues with paywalls around that

vegaro (commented):
I think this makes sense! Just couple of things that I mention in the comments:
- Does rotation work correctly?
- Does iOS trigger the `onDismiss` twice in the view? Becuase both the viewController and the delegate send the event
- Some naming suggestions

Other than that, I think this is almost ready to go!

purchases_ui_flutter/lib/views/customer_center_view.dart
facumenzella: Works as expected!

purchases_ui_flutter/ios/purchases_ui_flutter/Sources/purchases_ui_flutter/PurchasesUiCustomerCenterView.swift
facumenzella: Nice! Just debugged it and only happens for the standalone view (of course). Removed one

purchases_ui_flutter/android/src/main/kotlin/com/revenuecat/purchases_ui_flutter/PurchasesUiFlutterPlugin.kt
facumenzella: Do you mean renaming the mthod-channel event?

purchases_ui_flutter/android/src/main/kotlin/com/revenuecat/purchases_ui_flutter/PurchasesUiFlutterPlugin.kt
facumenzella: Ok I see what you mean. It felt natural to me, but I can do it "guard" style, and only set it if created

purchases_ui_flutter/android/src/main/kotlin/com/revenuecat/purchases_ui_flutter/PurchasesUiFlutterPlugin.kt
vegaro: yes. But up to you. It's just that at first this confused me and thought this was triggering a callback. Then I went to the `onDismiss` and saw it was doing something else (clearing the callbacks)

Feel free to ignore though. Coming to think about it I think it's valid either way

facumenzella: I been thinking about this. I was about to change it, but keeping it is consistent with the other platforms.

PR #1483, opened 2025-09-26
[V8] Prepare for next release

ajpallares:
This PR makes the changes needed for the next release of v8.

* Bump [fastlane-plugin-revenuecat_internal](https://github.com/RevenueCat/fastlane-plugin-revenuecat_internal) from `05ef095 ` to `1593f78`. See full diff in <a href="https://github.com/RevenueCat/fastlane-plugin-revenuecat_internal/compare/05ef0952ecb3b92398a0185a3f74019abd0c1d95...1593f78d0b9b24b48238337666183e3ba82f848e">compare view</a>.
* Pass current version in automatic bump.
* [CI] `bump` pipeline action won't trigger `deploy` workflow
* Fix dependencies declaration in Purchase Tester to fix failing CI jobs (iOS & macOS)

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -3,7 +3,7 @@ version '8.10.6'
 
 buildscript {
     ext.kotlin_version = '1.8.22'
-    ext.common_version = '14.3.0'
+    ext.common_version = '14.2.0'
     repositories {
         google()
         mavenCentral()
--- a/ios/purchases_flutter.podspec
+++ b/ios/purchases_flutter.podspec
@@ -15,7 +15,7 @@ Pod::Spec.new do |s|
   s.source_files = 'Classes/**/*'
   s.public_header_files = 'Classes/**/*.h'
   s.dependency 'Flutter'
-  s.dependency 'PurchasesHybridCommon', '14.3.0'
+  s.dependency 'PurchasesHybridCommon', '14.2.0'
   s.ios.deployment_target = '13.0'
   s.swift_version         = '5.0'
 
--- a/macos/purchases_flutter.podspec
+++ b/macos/purchases_flutter.podspec
@@ -16,7 +16,7 @@ Pod::Spec.new do |s|
   s.source_files     = 'Classes/**/*'
   s.public_header_files = 'Classes/**/*.h'
   s.dependency 'FlutterMacOS'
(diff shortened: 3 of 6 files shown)


vegaro (approved):

revenuecat_examples/purchase_tester/pubspec.yaml
```diff
@@ -15,6 +15,9 @@ dependencies:
 
   app_links: ^6.3.3
 
+  purchases_flutter:
+  purchases_ui_flutter:
```
ajpallares, line 19: I had to declare these as explicit dependencies of the Purchase Tester app because, otherwise, the autogenerated file `revenuecat_examples/purchase_tester/macos/Flutter/GeneratedPluginRegistrant.swift` would not include the registration of `PurchasesFlutterPlugin`. This was making the tests fail in iOS and macOS (e.g. https://app.circleci.com/pipelines/github/RevenueCat/purchases-flutter/4731).

I honestly don't know how this didn't fail some months ago when 8.10.6 was released... 🤷‍♂️

revenuecat_examples/purchase_tester/macos/Flutter/GeneratedPluginRegistrant.swift
```diff
@@ -7,10 +7,8 @@ import Foundation
 
 import app_links
 import purchases_flutter
-import purchases_ui_flutter
```
ajpallares, line 10: this file is autogenerated and does not include `import purchases_ui_flutter` because `purchases_ui_flutter` does not support macOS

ajpallares:
I'm making the PHC bump a separate PR so that it get calls out in the Release notes, as it will only b e the only developer-facing change in the next v8 release

revenuecat_examples/purchase_tester/pubspec.yaml
tonidero: Hmm I'm indeed a bit confused... But I think it should indeed work, since we're just overriding the dependencies later

ajpallares: Yeah, I honestly don't know why. Maybe the newer flutter version requires dependencies to be explicitly declare in order to register them... 
Note that the error is reproduced locally if the `purchases_flutter` dependency is not explicitly declared

PR #1489, opened 2025-09-29
Add CODEOWNERS

ajpallares:
### Motivation
It's useful to be notified when some changes happen (e.g. automatic PRs opened from CI)


.github/CODEOWNERS
```diff
@@ -0,0 +1,4 @@
+# CODEOWNERS for purchases-flutter
+
+# Default owners
+* @RevenueCat/coresdk
```
tonidero, line 4: Hmm I guess we could add the paywalls/customer center paths to ping monetization/customer center teams... But I guess we can iterate on this later as well

ajpallares: Yeah, it makes sense. I intentionally kept it only for CoreSDK because my initial purpose was to prevent us from missing some automatic PRs (e.g. the ones that upload the fastlane plugin). Also, it was not very obvious to me the paths that should be assigned to monetization/customer center teams.

Anyhow, I've pushed a change that adds monetization to certain paths (https://github.com/RevenueCat/purchases-flutter/pull/1489/commits/092ee615bd7999242610b1b112b54a798fb34036). I couldn't find any paths specific to Customer Center though. I'd love your input on this. Thanks Toni! 🙌

tonidero: Oh right... we only have customer center with the "present" functions, which are mixed with those for paywalls... I didn't look into it, but considering that, I'm worrying it might get noisy for those teams... So maybe it's ok to leave it at coresdk default owner and that's it... 

Might be good if we split the code better in the future, but I think the false pings could get annoying otherwise

ajpallares: Exactly, yes. I agree 👍 
Reverted at https://github.com/RevenueCat/purchases-flutter/pull/1489/commits/15ed228df8d20262980c06429b52208f160cd9f3

PR #1495, opened 2025-10-03
Update PHC to 17.9.0 in Package.swift of our SDKs

tonidero:
Looks like these versions haven't been updated for a while when using SPM. This fixes to use the latest version, which I think should fix things moving forward.


ajpallares (approved):
Nice catch! I totally missed this when adding SPM support 😓 
Thank you!

PR #1520, opened 2025-11-04
Add DEVELOPMENT instructions and fix `enableLocalBuild` task

vegaro:
I had to deal with enabling local dependencis yesterday and noticed our instructions were old and some even broken, for example `enableLocalBuild` was not doing anything.


android/build.gradle
```diff
@@ -49,7 +49,7 @@ android {
         // Gradle doesn't know which variant of purchases-hybrid-common to use when using
         // includeBuild and working with a local copy of purchases-hybrid-common
         // https://developer.android.com/studio/build/build-variants#resolve_matching_errors
-        missingDimensionStrategy 'dependencyVersions', 'latestDependencies', 'unityIAP'
+        missingDimensionStrategy 'apis', 'defaults'
```
tonidero, line 52: The comment above seems outdated as well? We only have variants right now in `purchases-android`.

DEVELOPMENT.md
```diff
@@ -1,18 +1,425 @@
...
+      ```bash
+      cd /path/to/purchases-hybrid-common/android
+      ./gradlew publishToMavenLocal
+      ```
+   
+   2. Add mavenLocal() to your sample app's `android/build.gradle`:
+      ```gradle
+      allprojects {
+          repositories {
+              mavenLocal()  // Add this first
```
tonidero, line 352: I've always wondered if we should leave this in, at least in purchase tester... But that's a separate thing.

DEVELOPMENT.md
```diff
@@ -1,18 +1,425 @@
...
+  - Ensure your local dependency paths are correct in podspecs and Podfiles
+
+- **General iOS Issues**:
+  - Clear Xcode derived data if builds fail
+  - Verify code signing settings for your development team
+
+### Android Issues
+- **Gradle Version Compatibility**: Ensure all projects use compatible AGP versions. If you get "Using multiple versions of the Android Gradle plugin" errors:
+  - Check that purchases-flutter, purchases-hybrid-common, and purchases-android all use the same AGP version
+  - Currently requires AGP 8.13.0 and Gradle 8.13+
```
tonidero: This is very easy to be outdated 😢 Not sure if it would be worth just including a reference to the location where the versions are defined currently...

revenuecat_examples/MagicWeather/android/app/build.gradle
```diff
@@ -81,3 +81,47 @@ flutter {
...
+        if (project.hasProperty('commonPath')) {
+            new File(rootProject.projectDir, ".composite-enable").text = commonPath
+            println "Enabled local purchases-hybrid-common at: $commonPath"
+        }
+        if (project.hasProperty('androidPath')) {
+            new File(rootProject.projectDir, ".android-enable").text = androidPath
+            println "Enabled local purchases-android at: $androidPath"
+        }
+        if (!project.hasProperty('commonPath') && !project.hasProperty('androidPath')) {
+            throw new GradleException("Please specify -PcommonPath or -PandroidPath parameter")
```
JayShortway: Seems like both is also supported?
```suggestion
            throw new GradleException("Please specify -PcommonPath and/or -PandroidPath parameter")
```

DEVELOPMENT.md
vegaro: Actually I agree with you. I think I added it to react native 🤔 

I don't think it hurts so yes let's add it (in another PR)

PR #1523, opened 2025-11-04
Support addOnSubscriptionOptions

fire-at-will:
This PR allows developers using the Flutter SDK to purchase a Play Store subscription with `SubscriptionOption` add-ons by adding `addOnSubscriptionOptions` to the `PurchaseParams` constructors.

It also:
- Stops passing the presented offering context for add-on StoreProducts to the PHC layer since the backend only considers presented offering contexts for the base item in a subscription with add-ons
- Fixes a bug where the add-on `StoreProduct`s weren't being passed from Flutter's Java layer to PHC when purchasing a `Package` or `SubscriptionOption`
- Updates the purchase tester's add-ons screen to allow you to select whether each option is purchased as a `StoreProduct` or as a `SubscriptionOption`. This is helpful to test the various combinations purchasable items that the base item and add-ons can be:
Screenshot 2025-11-04 at 6 16 26 PM

### DIFF
--- a/android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
+++ b/android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
@@ -154,8 +154,10 @@ public class PurchasesFlutterPlugin implements FlutterPlugin, MethodCallHandler,
                 type = call.argument("type");
                 Map<String, Object> presentedOfferingContext = call.argument("presentedOfferingContext");
                 List<Map<String, Object>> addOnStoreProducts = call.argument("addOnStoreProducts");
+                List<Map<String, Object>> addOnSubscriptionOptions = call.argument("addOnSubscriptionOptions");
                 purchaseProduct(productIdentifier, type, googleOldProductIdentifer, googleProrationMode,
-                        googleIsPersonalizedPrice, presentedOfferingContext, addOnStoreProducts, result);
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
```diff
@@ -154,17 +154,22 @@ public void onMethodCall(@NonNull MethodCall call, @NonNull Result result) {
...
+                        googleIsPersonalizedPrice, presentedOfferingContext, addOnStoreProducts,
+                        addOnSubscriptionOptions, result);
                 break;
             case "purchasePackage":
                 String packageIdentifier = call.argument("packageIdentifier");
                 presentedOfferingContext = call.argument("presentedOfferingContext");
                 googleOldProductIdentifer = call.argument("googleOldProductIdentifier");
                 googleProrationMode = call.argument("googleProrationMode");
                 googleIsPersonalizedPrice = call.argument("googleIsPersonalizedPrice");
+                addOnStoreProducts = call.argument("addOnStoreProducts");
```
fire-at-will, line 168: I'm not a fan of the pattern we're using here where the variables are declared in an earlier case in this switch...I might make a follow up PR to move away from this pattern 😅

tonidero: Indeed, it's very confusing... I think ideally we would use kotlin for this.

PR #1542, opened 2025-11-17
Add BC7 version of library when deploying

tonidero:
With this, when we deploy a new version of the library, we will also deploy versions with a `+bc7` metadata using the bc7 version of the Android native SDK.

Depends on https://github.com/RevenueCat/purchases-hybrid-common/pull/1371

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -3,7 +3,7 @@ version '9.9.7'
 
 buildscript {
     ext.kotlin_version = '1.8.22'
-    ext.common_version = '17.18.0'
+    ext.common_version = '17.18.1'
     repositories {
         google()
         mavenCentral()
--- a/ios/purchases_flutter.podspec
+++ b/ios/purchases_flutter.podspec
@@ -15,7 +15,7 @@ Pod::Spec.new do |s|
   s.source_files = 'purchases_flutter/Sources/purchases_flutter/**/*.{h,m}'
   s.public_header_files = 'purchases_flutter/Sources/purchases_flutter/include/**/*.h'
   s.dependency 'Flutter'
-  s.dependency 'PurchasesHybridCommon', '17.18.0'
+  s.dependency 'PurchasesHybridCommon', '17.18.1'
   s.ios.deployment_target = '13.0'
   s.swift_version         = '5.0'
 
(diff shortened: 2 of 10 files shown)


fastlane/Fastfile
```diff
@@ -60,6 +60,11 @@ lane :bump do |options|
     versions_file_path: versions_path,
     is_prerelease: options[:is_prerelease]
   )
+
+  # Update CHANGELOG.md to include BC7 version
+  update_changelog_with_bc7_version(changelog_path, current_version_number)
```
tonidero, line 65: Have to do this because pub.dev validates that the version number appears in the changelog. This seemed the simpler way to bypass this validation.

fastlane/Fastfile
```diff
@@ -114,6 +119,72 @@ lane :release_purchases_ui_flutter do |options|
...
+  # Modify files in-place for BC7 variant
+  UI.message("Creating BC7 variant of purchases_flutter in-place...")
+
+  # Apply BC7 modifications
+  apply_bc7_modifications('..', false)
+
+  # Temporarily commit changes to satisfy git clean check
+  Dir.chdir('..') do
+    sh('git', 'add', 'pubspec.yaml', 'android/build.gradle')
+    sh('git', 'commit', '-m', '[TEMPORARY] BC7 modifications for publishing')
```
tonidero, line 133: I need to commit the files with changes, otherwise it will fail pub.dev validation. This should be fine during CI, since we won't be pushing this commit. But we reset the changes at the end anyway... It might not be undone if the flutter pub publish command fails for any reason though, but this should be fine when run on CI as I mentioned.

tonidero:
Keeping this as draft until the PHC part is merged, but this is ready for review @RevenueCat/sdk

fastlane/Fastfile
```diff
@@ -114,6 +119,72 @@ lane :release_purchases_ui_flutter do |options|
...
+  end
+
+  # Dry run
+  Dir.chdir('..') do
```
rickvdl, line 137: I was just thinking it might be nice to be able to reuse existing code a bit more, for example calling the regular `release` lane here. But I also realized this would often require passing some new parameters and would probably not be worth the effort. Since it should be temporary in the grand scheme of things anyways I think this approach makes sense as-is.

tonidero: Yeah... I think we could try to refactor it to avoid repeating code, but as you said, this will go away in a few months at most, so might be easier to clean it up later if we keep it split.

tonidero:
Closing this for now, since we're going with a different approach, by adding an AIDL file to the native SDK.

PR #1561, opened 2025-12-04
[EXTERNAL] Add support for preferredUILocaleOverride to Purchases Flutter SDK

marcorizza:
## Description
This PR adds support for preferredUILocaleOverride to the Purchases Flutter SDK, allowing developers to explicitly override the UI locale used by RevenueCat paywalls and related interfaces. This feature is documented in other SDKs but was missing from the Flutter implementation: [https://www.revenuecat.com/docs/tools/paywalls/displaying-paywalls#setting-preferred-locale](https://www.revenuecat.com/docs/tools/paywalls/displaying-paywalls#setting-preferred-locale)

## Changes
- Added preferredUILocaleOverride optional parameter to the PurchasesConfiguration class
- Added native Android implementation in PurchasesFlutterPlugin.java
- Added native iOS implementation in PurchasesFlutterPlugin.m

## Testing
- Added dedicated test cases for the new locale-override functionality:
  - Verifying that `preferredUILocaleOverride` is included in the configuration payload during setup.
  - Ensuring that `overridePreferredLocale()` correctly sends the specified locale to the native layer.
  - Confirming that passing `null` resets the override and reverts to the system default.
- Updated existing tests to include `preferredUILocaleOverride: null` where required, ensuring consistency with the new configuration schema.
- All tests pass successfully (`flutter test` — exit code 0)

## Motivation
This feature enables developers to control the locale used by RevenueCat UI elements independently of the device's system locale. It allows better localization management and aligns the Flutter SDK with other platform implementations.

### DIFF
--- a/ios/purchases_flutter/Sources/purchases_flutter/PurchasesFlutterPlugin.m
+++ b/ios/purchases_flutter/Sources/purchases_flutter/PurchasesFlutterPlugin.m
@@ -292,9 +292,6 @@ automaticDeviceIdentifierCollectionEnabled:(BOOL)automaticDeviceIdentifierCollec
     if ([userDefaultsSuiteName isKindOfClass:NSNull.class]) {
         userDefaultsSuiteName = nil;
     }
-    if ([preferredUILocaleOverride isKindOfClass:NSNull.class]) {
-        preferredUILocaleOverride = nil;
-    }
 
     RCPurchases *purchases = [RCPurchases configureWithAPIKey:apiKey
                                                     appUserID:appUserID


android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
```diff
@@ -397,6 +403,11 @@ private void setupPurchases(String apiKey, String appUserID,
                     verificationMode,
                     pendingTransactionsForPrepaidPlansEnabled,
                     automaticDeviceIdentifierCollectionEnabled);
+            
+            if (preferredUILocaleOverride != null) {
+                CommonKt.overridePreferredLocale(preferredUILocaleOverride);
```
tonidero: The [CommonKt.configure](https://github.com/RevenueCat/purchases-hybrid-common/blob/94794c632fe87b6f4d88f74e781b2b8c64d775fa/android/hybridcommon/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt#L947) accepts a nullable locale. It's preferred to pass that parameter at configure time, than setting it up later. Could you change that? 🙏

ios/purchases_flutter/Sources/purchases_flutter/PurchasesFlutterPlugin.m
```diff
@@ -300,6 +308,11 @@ - (void)setupPurchases:(NSString *)apiKey
                                              verificationMode:verificationMode
                                            diagnosticsEnabled:NO
                    automaticDeviceIdentifierCollectionEnabled:automaticDeviceIdentifierCollectionEnabled];
+    
+    if (preferredUILocaleOverride != nil) {
+        [RCCommonFunctionality overridePreferredLocale:preferredUILocaleOverride];
```
tonidero: Same as in Android, I think we can just pass the preferred locale to the [configure](https://github.com/RevenueCat/purchases-hybrid-common/blob/94794c632fe87b6f4d88f74e781b2b8c64d775fa/ios/PurchasesHybridCommon/PurchasesHybridCommon/Purchases%2BHybridAdditions.swift#L29) call above.

lib/models/purchases_configuration.dart
```diff
@@ -5,7 +5,16 @@ class PurchasesConfiguration {
   /// RevenueCat API Key.
   final String apiKey;
 
-  PurchasesConfiguration(this.apiKey);
+  PurchasesConfiguration(
+    this.apiKey, {
+    this.preferredUILocaleOverride,
```
tonidero: Hmm so I do think this makes sense, but we probably should do it for all optional parameters of PurchasesConfiguration. It's a bit strange to only do it for this one. 

I would say, maybe we can leave this out from this PR, just leaving it to be set as all other parameters, and then we can fix the PurchasesConfiguration constructor with all the parameters in a separate PR?

tonidero (commented):
Amazing job!! Thank you so much for contributing this 🙇. I just left some minor comments. I think once those are in, we can merge this. Thanks again!

ios/purchases_flutter/Sources/purchases_flutter/PurchasesFlutterPlugin.m
ajpallares: Also, small nitpick. We can get rid of the `NSNull` check with the existing method `mappingNSNullToNil`. So, together with Toni's comment, we could just pass:
```objc
preferredLocale: preferredUILocaleOverride.mappingNSNullToNil
```

ios/purchases_flutter/Sources/purchases_flutter/PurchasesFlutterPlugin.m
```diff
@@ -416,6 +429,15 @@ - (void)setProxyURLString:(nullable NSString *)proxyURLString
     result(nil);
 }
 
+- (void)overridePreferredLocale:(nullable NSString *)locale
+                         result:(FlutterResult)result {
+    if ([locale isKindOfClass:NSNull.class]) {
+        locale = nil;
+    }
+    [RCCommonFunctionality overridePreferredLocale:locale];
```
ajpallares: This is a nice safety touch. But (the same as above), it can be simplified to 

```suggestion
    [RCCommonFunctionality overridePreferredLocale:locale.mappingNSNullToNil];
```

lib/purchases_flutter.dart
```diff
@@ -185,10 +185,29 @@ class Purchases {
...
     );
   }
 
+  /// Overrides the preferred UI locale used by RevenueCat UI components.
+  ///
+  /// When provided, the SDK will use the specified locale instead of the system default.
+  /// Both "es-ES" and "es_ES" formats are supported.
+  ///
+  /// Pass null to revert to the system default.
+  /// When the locale is changed, the SDK will automatically clear any cached offerings.
```
ajpallares: I _think_ I would probably remove this line from the documentation, since:
1. It's an implementation detail
2. It opens the door to abuse
```suggestion
```

ajpallares (commented):
This looks fantastic! Thank you so much for your contribution! Pretty useful indeed 👏 

I have one last ask 🙏. Since we're adding a new public API in this PR, could you please add some API tests in [purchases_flutter_api_test.dart](https://github.com/RevenueCat/purchases-flutter/blob/main/api_tester/lib/api_tests/purchases_flutter_api_test.dart)

This way we make sure we're not accidentally breaking this new API in the future.

I also added a couple of small comments about some tiny code cleanup. But, regardless, it looks amazing!

We can probably merge this as is once all those small details are addressed 😊

marcorizza:
Thanks so much for the helpful guidance 🙏. I have updated the PR with all the requested changes. Could you take a look at the latest commits to confirm everything is in good shape? If there is anything else I should adjust, please feel free to let me know

ios/purchases_flutter/Sources/purchases_flutter/PurchasesFlutterPlugin.m
```diff
@@ -280,13 +284,17 @@ - (void)setupPurchases:(NSString *)apiKey
...
                 result:(FlutterResult)result {
     if ([appUserID isKindOfClass:NSNull.class]) {
         appUserID = nil;
     }
     if ([userDefaultsSuiteName isKindOfClass:NSNull.class]) {
         userDefaultsSuiteName = nil;
     }
+    if ([preferredUILocaleOverride isKindOfClass:NSNull.class]) {
+        preferredUILocaleOverride = nil;
+    }
```
ajpallares: This wouldn't be needed anymore thanks to using
```objc
preferredUILocaleOverride.mappingNSNullToNil
```

lib/purchases_flutter.dart
ajpallares: In this comment I meant removing this part:
> /// When the locale is changed, the SDK will automatically clear any cached offerings.

However, I would leave this part that has been removed 
>   /// Both "es-ES" and "es_ES" formats are supported.
  ///
  /// Pass null to revert to the system default.

lib/purchases_flutter.dart
```diff
@@ -185,10 +185,26 @@ class Purchases {
...
     );
   }
 
+  /// Overrides the preferred UI locale used by RevenueCat UI components.
+  ///
+  /// When provided, the SDK will use the specified locale instead of the system default.
+  /// When the locale is changed, the SDK will automatically clear any cached offerings.
+  ///
+  /// [locale] The locale identifier (e.g., "de-DE", "es_ES") or null to use the system default.
+  static Future<void> overridePreferredLocale(String? locale) =>
```
ajpallares: I would change this API to `overridePreferredUILocale`, to match the API in both Android and iOS:

[Android's `overridePreferredUILocale`](https://github.com/RevenueCat/purchases-android/blob/b18ae87b437054d01a0868e213bcc804201d9f4b/purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt#L927)

[iOS' `overridePreferredUILocale`](https://github.com/RevenueCat/purchases-ios/blob/520dd3bc94f0f47b058b9c6bd16175976d6276a8/Sources/Purchasing/Purchases/Purchases.swift#L1504)

And remember to update the API test too! 🙏

ajpallares (commented):
Thank you again for the quick iteration! We're almost there. Just some more comments. The main one about having the public API match the APIs in Android and iOS 😊

PR #1567, opened 2025-12-10
Updated CircleCI macOS executor from m1 to m4 to avoid deprecation

rickvdl:
The M1 executor will be deprecated soon, so I've updated the config to use the M4 executor instead, and centralized the config for easier updates down the road.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -41,6 +41,17 @@ orbs:
   flutter: circleci/flutter@2.1.0
   ruby: circleci/ruby@2.5.3
 
+executors:
+  macos-executor:
+    parameters:
+      xcode_version:
+        type: string
+        default: "15.4"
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -41,6 +41,17 @@ orbs:
   flutter: circleci/flutter@2.1.0
   ruby: circleci/ruby@2.5.3
 
+executors:
+  macos-executor:
+    parameters:
+      xcode_version:
+        type: string
+        default: "15.4"
```
tonidero, line 49: I guess updating to the 16.4 xcode might be trickier? (just checking)

rickvdl: Should probably be fine, will take a look at Xcode versions in a follow up PR :)

PR #1584, opened 2026-01-08
Update `missingDimensionStrategy` for local development

vegaro:
We need ot update the `missingDimensionStrategy` since we now have `billingclient`

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -49,7 +49,8 @@ android {
         // Gradle doesn't know which variant of purchases-android to use when using
         // includeBuild and working with a local copy of purchases-android
         // https://developer.android.com/studio/build/build-variants#resolve_matching_errors
-        missingDimensionStrategy 'apis', 'billingclient'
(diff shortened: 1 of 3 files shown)


android/build.gradle
```diff
@@ -49,7 +49,7 @@ android {
         // Gradle doesn't know which variant of purchases-android to use when using
         // includeBuild and working with a local copy of purchases-android
         // https://developer.android.com/studio/build/build-variants#resolve_matching_errors
-        missingDimensionStrategy 'apis', 'defaults'
+        missingDimensionStrategy 'apis', 'billingclient'
```
tonidero: This is wrong, I think it should be
```suggestion
        missingDimensionStrategy 'apis', 'defaults'
        missingDimensionStrategy 'billingclient', 'bc8'
```
Same in the other places

vegaro: you are correct

tonidero (approved):

PR #1599, opened 2026-01-27
Fix iOS setPostHogUserID not working due to argument key typo

limdauto:
## Summary
- Fixes `setPostHogUserID()` always setting an empty value on iOS
- The Dart side sends the argument with key `postHogUserID` (uppercase `ID`), but the iOS side was looking for `postHogUserId` (lowercase `Id`)
- Android works correctly since it uses the correct key

## Root Cause
```dart
// Dart sends:
_channel.invokeMethod('setPostHogUserID', {'postHogUserID': postHogUserID});
```

```objc
// iOS was expecting (wrong):
NSString *postHogUserID = arguments[@"postHogUserId"];  // lowercase 'd'

// Should be:
NSString *postHogUserID = arguments[@"postHogUserID"];  // uppercase 'D'
```

## Test plan
- [ ] Verify `Purchases.setPostHogUserID("test-id")` correctly sets the `$posthogUserId` subscriber attribute on iOS
- [ ] Verify the attribute value is non-empty in RevenueCat dashboard
- [ ] Verify Android still works correctly


limdauto:
@tonidero @ajpallares thank you both for reviewing. Could you help me merge it please?

Issue #1606, opened 2026-02-01
Warning: the package uses or overrides a deprecated API


robot-den:
- [x] I have updated Purchases SDK to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-flutter/blob/main/CONTRIBUTING.md)
- [x] I have searched the [Community](https://community.revenuecat.com)
- [x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [x] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-flutter/issues)

## Environment ##
- [x] Output of `flutter doctor`
  [✓] Flutter (Channel stable, 3.38.9, on macOS 15.6.1 24G90 darwin-arm64, locale en-KZ)
  [✓] Android toolchain - develop for Android devices (Android SDK version 36.0.0)
  [✓] Xcode - develop for iOS and macOS (Xcode 26.2)
  [✓] Connected device (2 available)
  [✓] Network resources

- [x] Version of `purchases-flutter`
  9.10.8

- [x] Testing device version
  Android emulator, android version 16

- [x] How often the issue occurs:
  Each project compilation

- [x] [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue
  Original warning:
```
 Note: /Users/dn/.pub-cache/hosted/pub.dev/purchases_flutter-9.10.8/android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java uses or overrides a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
```  
  If you pass [Xlint flags](https://github.com/flutter/flutter/issues/147831#issuecomment-2094304106), you can see more details:
```
/Users/dn/.pub-cache/hosted/pub.dev/purchases_flutter-9.10.8/android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java:564: warning: [deprecation] syncAmazonPurchase(String,String,String,String,Double) in Purchases has been deprecated
        Purchases.getSharedInstance().syncAmazonPurchase(productID, receiptID,
                                     ^
```

- [x] Steps to reproduce, with a description of expected vs. actual behavior
I think it can be reproduced on fresh app when it builds via `flutter run` or via VS Code.
I target Java 17 and Java 17 is used for compilation.

## Describe the bug ##
When app compiling with purchases_flutter `9.10.8` it drops warning.
```
[...]

PR #1607, opened 2026-02-02
Update `syncAmazonPurchase` to support `purchaseTime`

tonidero:
This should help dealing with #1606 

We added a new parameter to `syncAmazonPurchase` in purchases-android, the purchase time and deprecated the existing method, which caused some deprecated warnings here.

This updates the public API by adding a new optional parameter to pass that new data. While the ideal is to deprecate the existing method so people pass us the purchase time, this is tricky in dart, since it doesn't allow for method overloads, which would force us to make the new signature something like `syncAmazonPurchaseWithPurchaseTime`, which is not great...

I think we can make this non breaking by making this new parameter optional for now, and then do a breaking change and make it required on the next major. Wdyt?

### DIFF
--- a/api_tester/lib/api_tests/models/store_product_wrapper_api_test.dart
+++ b/api_tester/lib/api_tests/models/store_product_wrapper_api_test.dart
@@ -19,7 +19,13 @@ class _StoreProductApiTest {
       List<StoreProductDiscount>? discounts,
       ProductCategory? productCategory,
       String? subscriptionPeriod,
-      PresentedOfferingContext? presentedOfferingContext) {
+      PresentedOfferingContext? presentedOfferingContext,
+      double? pricePerWeek,
+      double? pricePerMonth,
+      double? pricePerYear,
+      String? pricePerWeekString,
+      String? pricePerMonthString,
+      String? pricePerYearString) {
     StoreProduct product = StoreProduct(
         identifier, description, title, price, priceString, currencyCode);
     product = StoreProduct(
@@ -28,7 +34,13 @@ class _StoreProductApiTest {
         discounts: discounts,
         productCategory: productCategory,
         subscriptionPeriod: subscriptionPeriod,
-        presentedOfferingContext: presentedOfferingContext);
+        presentedOfferingContext: presentedOfferingContext,
+        pricePerWeek: pricePerWeek,
+        pricePerMonth: pricePerMonth,
+        pricePerYear: pricePerYear,
+        pricePerWeekString: pricePerWeekString,
+        pricePerMonthString: pricePerMonthString,
+        pricePerYearString: pricePerYearString);
   }
 
   void _checkProperties(StoreProduct product) {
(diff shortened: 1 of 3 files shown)


api_tester/lib/api_tests/purchases_flutter_api_test.dart
```diff
@@ -546,9 +546,13 @@ class _PurchasesFlutterApiTest {
...
 
   void _checkSyncAmazonPurchase(String productID, String receiptID,
-      String amazonUserID, String? isoCurrencyCode, double? price) async {
+      String amazonUserID, String? isoCurrencyCode, double? price,
+      int? purchaseTime) async {
     Future<void> future = Purchases.syncAmazonPurchase(
         productID, receiptID, amazonUserID, isoCurrencyCode, price);
+    Future<void> future2 = Purchases.syncAmazonPurchase(
+        productID, receiptID, amazonUserID, isoCurrencyCode, price,
+        purchaseTime: purchaseTime);
```
tonidero, line 555: TBH, I don't like this new API much, as I mention in the PR description... The alternative is to find a different name for this method, happy to hear thoughts, but couldn't find any that could be confusing later :(

android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
```diff
@@ -354,8 +354,13 @@ public void onMethodCall(@NonNull MethodCall call, @NonNull Result result) {
                 String amazonUserID = call.argument("amazonUserID");
                 String isoCurrencyCode = call.argument("isoCurrencyCode");
                 Double price = call.argument("price");
+                Integer purchaseTime = call.argument("purchaseTime");
+                Long purchaseTimeLong = null;
+                if (purchaseTime != null) {
+                    purchaseTimeLong = purchaseTime.longValue();
```
tonidero, line 360: Note this conversion I had to do, in dart, milliseconds since epoch is expressed as an integer, so we need to convert it to a long, which is what the native SDK expects here.

android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
```diff
@@ -555,14 +560,22 @@ private void setDebugLogsEnabled(boolean enabled, final Result result) {
         result.success(null);
     }
 
+    @SuppressWarnings("deprecation")
```
tonidero, line 563: Since we're still using the deprecated method, we do this to avoid propagating it to consumers.

api_tester/lib/api_tests/purchases_flutter_api_test.dart
ajpallares: I agree with your proposed solution in the PR description. I agree it would get confusing...

lib/purchases_flutter.dart
```diff
@@ -1250,13 +1250,16 @@ class Purchases {
   /// @param [amazonUserID] Amazon's userID. This parameter will be ignored when syncing a Google purchase.
   /// @param [isoCurrencyCode] Product's currency code in ISO 4217 format.
   /// @param [price] Product's price.
+  /// @param [purchaseTime] Purchase time in milliseconds since epoch. Usage of this parameter is
+  /// highly recommended and usages without it are deprecated and will be removed in future versions.
```
ajpallares, line 1254: Seems like this is the most we can do to lead developers to pass the `purchaseTime` parameter, yes

PR #1608, opened 2026-02-02
Added LeakCanary to the purchase tester example app on Android

rickvdl:


tonidero (approved):
Nice find about those integration tests! Just to check but are we sure that doing this, it will catch leaks during integration tests? Asking because we're not calling the `assertNoLeaks` method in leakcanary instrumentation... but looking at our integration tests, maybe it will catch those leaks, not sure.
In any case, I think adding it to the sample app should already be helpful 💪

rickvdl:
> Nice find about those integration tests! Just to check but are we sure that doing this, it will catch leaks during integration tests? Asking because we're not calling the `assertNoLeaks` method in leakcanary instrumentation... but looking at our integration tests, maybe it will catch those leaks, not sure. In any case, I think adding it to the sample app should already be helpful 💪

As discussed over Slack; we don't really have much integration tests setup right now, so the surface would be quite small. We'll leave them out of scope for now but will add LeakCanary to the E2E tests as soon as we have them setup for the hybrids.

PR #1617, opened 2026-02-06
Fixed offerings order in purchase tester app not being stable

rickvdl:
Makes it easier to find the same offering after relaunching during testing

### DIFF
--- a/revenuecat_examples/purchase_tester/lib/src/upsell.dart
+++ b/revenuecat_examples/purchase_tester/lib/src/upsell.dart
@@ -81,8 +81,15 @@ class _UpsellScreenState extends State<UpsellScreen> {
   Widget _buildUpsell(BuildContext context) {
     final currentOfferingId = _offerings?.current?.identifier;
     // Sort offerings by identifier (key) for stable ordering, case-insensitive
+    // Put current/default offering first
     final sortedOfferings = _offerings!.all.entries.toList()
-      ..sort((a, b) => a.key.toLowerCase().compareTo(b.key.toLowerCase()));
+      ..sort((a, b) {
+        // Current offering always comes first
+        if (a.key == currentOfferingId) return -1;
+        if (b.key == currentOfferingId) return 1;
+        // Otherwise sort alphabetically
+        return a.key.toLowerCase().compareTo(b.key.toLowerCase());
+      });
     
     return ListView(children: [
       if (_customerInfo != null)


revenuecat_examples/purchase_tester/lib/src/upsell.dart
```diff
@@ -80,6 +80,10 @@ class _UpsellScreenState extends State<UpsellScreen> {
 
   Widget _buildUpsell(BuildContext context) {
     final currentOfferingId = _offerings?.current?.identifier;
+    // Sort offerings by serverDescription (name) for stable ordering
+    final sortedOfferings = _offerings!.all.entries.toList()
+      ..sort((a, b) => a.value.serverDescription.compareTo(b.value.serverDescription));
```
ajpallares: Does it make sense to sort by `serverDescription` is that's not visible in the UI? Would it make more sense to sort by `key` instead?

rickvdl: I thought `serverDescription` was (or at least contained) the visible name, but let me double check

rickvdl: Right, it's indeed the 'display name' in the dashboard, but you're right that this is not something we display. So I'll change it to use the ID instead, thanks!

revenuecat_examples/purchase_tester/lib/src/upsell.dart
```diff
@@ -80,6 +80,10 @@ class _UpsellScreenState extends State<UpsellScreen> {
 
   Widget _buildUpsell(BuildContext context) {
     final currentOfferingId = _offerings?.current?.identifier;
+    // Sort offerings by identifier (key) for stable ordering, case-insensitive
+    final sortedOfferings = _offerings!.all.entries.toList()
+      ..sort((a, b) => a.key.toLowerCase().compareTo(b.key.toLowerCase()));
```
tonidero: Could be nice to show the default offering first always, then the rest sorted by id. Not a huge deal though.

tonidero (approved):
Just a possible improvement. Not a blocker though . Thanks for doing this!

revenuecat_examples/purchase_tester/lib/src/upsell.dart
rickvdl: Great idea! Added

PR #1632, opened 2026-02-12
Add CLAUDE.md and AGENTS.md for AI assistant guidance

facumenzella:
## Summary
- Adds `CLAUDE.md` and `AGENTS.md` files to help AI coding assistants work effectively in this repository
- Both files contain identical content (different AI tools look for different filenames)

## Key Content
- **Public API stability requirements** — emphasizes avoiding breaking changes at all costs
- **Related repositories** — links to purchases-ios, purchases-android, and purchases-hybrid-common for reference
- **Code structure overview** — explains what lives where
- **Testing commands** — all the commands needed to verify changes
- **Coding conventions** — patterns to follow for Dart and platform code
- **Guardrails** — what NOT to do

## Test plan
- [x] Files are valid markdown
- [x] Commands in the file are accurate (verified against DEVELOPMENT.md and CI config)

🤖 Generated with [Claude Code](https://claude.ai/code)


CLAUDE.md
```diff
@@ -0,0 +1,163 @@
+# purchases-flutter — Development Guidelines
```
tonidero: Should this just be a symlink of the other? (or the other way around 😅 )

AGENTS.md
```diff
@@ -0,0 +1,163 @@
+# purchases-flutter — Development Guidelines
+
+## Project Overview
+
+RevenueCat's official Flutter SDK for in-app purchases and subscriptions. This plugin wraps the native iOS and Android SDKs, providing a unified Dart API for Flutter apps.
+
+**Related repositories:**
+- **iOS SDK**: https://github.com/RevenueCat/purchases-ios
+- **Android SDK**: https://github.com/RevenueCat/purchases-android
+- **Hybrid Common**: https://github.com/RevenueCat/purchases-hybrid-common (shared layer for Flutter, React Native, etc.)
```
tonidero: Could be good to add `purchases-js` since we also support web here. Also, might be worth mentioning some more details about PHC, since that repo has multiple libraries, so it could benefit of more context.

facumenzella: I think we can link it to the new one right? the one with submodules

tonidero: I was thinking that each repo can still keep their own details in files like this, since many times we don't need to work cross-repo + that way we keep the context more scoped. So yeah, I was thinking just a small mention here would be enough, but NABD :)

PR #1657, opened 2026-02-27
Add PurchaseLogic support to PaywallView

rickvdl:
## Summary
- Adds `PaywallPurchaseLogic` interface allowing apps to handle purchase and restore operations when `purchasesAreCompletedBy` is set to `myApp`
- When `purchaseLogic` is provided to `PaywallView`, the paywall delegates purchase/restore to the app instead of using RevenueCat's default flow
- Uses `HybridPurchaseLogicBridge` from `purchases-hybrid-common` for async request/response coordination via `requestId`-based correlation

## Related PRs
- React Native: https://github.com/RevenueCat/react-native-purchases/pull/1607

## Test plan
- [x] Build and run on iOS simulator with `purchasesAreCompletedBy: myApp` — verify `performPurchase` is called on package tap
- [x] Verify returning `success` / `cancellation` / `error` correctly updates paywall state
- [x] Verify `performRestore` is called when user taps restore
- [x] Build and run on Android emulator with same configuration — verify same behavior
- [x] Verify paywall without `purchaseLogic` continues to work as before (no regression)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/revenuecat_examples/purchase_tester/lib/src/constant.dart
+++ b/revenuecat_examples/purchase_tester/lib/src/constant.dart
@@ -1,5 +1,5 @@
 //TO DO: add the Apple API key for your app from the RevenueCat dashboard: https://app.revenuecat.com
-const appleApiKey = 'appl_api_key';
+const appleApiKey = 'appl_SYkbbYptqjDadVkQeBUtXxVxCcw';
 
 //TO DO: add the Google API key for your app from the RevenueCat dashboard: https://app.revenuecat.com
 const googleApiKey = 'googl_api_key';
--- a/revenuecat_examples/purchase_tester/lib/src/upsell.dart
+++ b/revenuecat_examples/purchase_tester/lib/src/upsell.dart
@@ -350,13 +350,32 @@ class _UpsellScreenState extends State<UpsellScreen> {
                             builder: (context) => PaywallScreen(
                                   offering: offering,
                                   customVariables: _getCustomVariablesForPaywall(),
-                                  purchaseLogic: purchasesAreCompletedByMyApp
-                                      ? SamplePurchaseLogic()
-                                      : null,
                                 )),
                       );
                     },
                     child: const Text('Show paywall view'),
+                  ),
+                  ElevatedButton(
+                    onPressed: purchasesAreCompletedByMyApp
+                        ? () async {
+                            Navigator.push(
+                              context,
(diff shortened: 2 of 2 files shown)


api_tester/lib/api_tests/purchases_ui_flutter_api_test.dart
```diff
@@ -231,6 +231,25 @@ class _PurchasesFlutterApiTest {
...
+      case PurchaseLogicResult.error:
+        break;
+    }
+  }
+
+  Widget _checkPaywallViewWithPurchaseLogic(PaywallPurchaseLogic logic) {
+    return Scaffold(
+      body: Center(
+        child: PaywallView(
+          purchaseLogic: logic,
```
tonidero, line 247: Should we add some usages of this to the sample app as well for easier testing? I guess it won't be used most of the time while `PurchasesAreCompletedBy.REVENUECAT`... But we can probably show a button disabled in this scenario with an indicator to change that to test it.

rickvdl: Good idea! I've added this based on the `purchasesAreCompletedByMyApp` variable. I did not add a button per se, because I think we can only have the `Show paywall view` button as an example, because
- When PACB myApp it is required to pass the purchaselogic
- When PACB revenueCat it won't have any effect

No when `purchasesAreCompletedByMyApp` is set to true the custom purchase logic will be used. Please LMK what you think :)

tonidero: Indeed, the only reason to separate into 2 buttons is to test what would happen if not passing `PurchaseLogic` when `PurchasesAreCompletedBy.REVENUECAT`... The other advantage is to have an explicit UI that reminds us of this case, so it's easier to remember to test it... But TBH, not attached to the idea. I'm ok keeping it in one button.

rickvdl: Right, understand. I do indeed think it's good for discoverability during testing, added it like so :)

Simulator Screenshot - iPhone 17 Pro - 2026-03-02 at 13 34 27

tonidero: I think that looks good. Thank you!!

PR #1629, opened 2026-02-12
Enable pod update retry for bump-phc branches

facumenzella:
## Summary

Fixes the pod update retry logic from #1618 to also apply to `bump-phc/*` branches.

### Problem
The retry logic was only enabled for `release/*` branches in the deploy workflow, but PHC bump PRs run on `bump-phc/*` branches using the test workflow where retry was not enabled.

### Changes
- Added `bump-phc-branches` alias to match `bump-phc/*` branches
- Added `exclude-bump-phc-branches` alias to exclude them from regular test runs
- Test workflow now runs CocoaPods integration tests:
  - **Without retry** on regular branches (fast fail)
  - **With retry** on `bump-phc/*` branches (handles propagation delays)

### Related
- Fixes CI failures on #1628
- Follow-up to #1618

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #1618, opened 2026-02-06
Add pod update with retry for CocoaPods propagation issues

facumenzella:
## Summary

Adds retry logic for `pod update` in iOS/macOS integration tests to handle CocoaPods CDN propagation delays during PHC bumps.

### Changes
- New `pod-update-with-retry` command with configurable retries (default: 5 retries, 5 min wait)
- Added `enable_retry` parameter to `ios-integration-test-cocoapods` and `macos-integration-test-cocoapods` jobs
- Retry is only enabled in `deploy` workflow (release branches) where PHC bumps run

### How it works
- On regular CI runs: `pod update --repo-update` runs once (no retry)
- On release branches (PHC bumps): retries up to 5 times with 5-minute waits for CocoaPods propagation

Closes https://github.com/RevenueCat/sdks-circleci-orb/pull/29

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #1663, opened 2026-03-05
Fix CocoaPods retry job names breaking GitHub required checks

ajpallares:
## Summary

- Fixes CocoaPods integration test jobs reporting different names on `bump-phc/*` branches, which broke GitHub required checks.
- Instead of duplicating jobs with branch filters (introduced in #1629), the `pod-update-with-retry` command now auto-detects `bump-phc/*` branches at runtime and enables retry accordingly.
- Removes the now-unused `bump-phc-branches` / `not-bump-phc-branches` aliases and the `enable_retry` parameter from jobs and command.
- Removes unnecessary retry for `release/*` branches, since they don't run right after a new PHC version is published.

## Related

- Simplifies the approach from #1629
- Original retry mechanism added in #1618

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -233,7 +233,7 @@ commands:
           working_directory: << parameters.working_directory >>
           command: |
             enable_retry=false
-            if [[ "$CIRCLE_BRANCH" == bump-phc/* ]] || [[ "$CIRCLE_BRANCH" == release/* ]]; then
+            if [[ "$CIRCLE_BRANCH" == bump-phc/* ]]; then
               enable_retry=true
             fi
 


.circleci/config.yml
```diff
@@ -247,7 +232,11 @@ commands:
           name: Pod update
           working_directory: << parameters.working_directory >>
           command: |
-            enable_retry=<< parameters.enable_retry >>
+            enable_retry=false
+            if [[ "$CIRCLE_BRANCH" == bump-phc/* ]] || [[ "$CIRCLE_BRANCH" == release/* ]]; then
```
tonidero: I do wonder why this is needed for release branches? Don't think we should be updating pods there? Not sure if it refers to pushing new pods?

tonidero (approved):
Just a question but looks good to me. Thank you!!

.circleci/config.yml
ajpallares: Right. Good question 🤔 

Normally, release branches don't happen right after PHC gets released, so they wouldn't need the retries for CocoaPods, it would seem.

@facumenzella do you remember why it was originally enabled for release branches in https://github.com/RevenueCat/purchases-flutter/pull/1618?

facumenzella: I think my goal was to reuse it. And since flutter uses its own thing, this was the easiest. BUT, I agree it shouldn't be needed for releases

facumenzella: Let's take it out

ajpallares: Okay!

Good catch @tonidero, btw

ajpallares: Done in https://github.com/RevenueCat/purchases-flutter/pull/1663/commits/e162b7af58913fdbb18240cc2e51c8a582a341d3

Now only `bump-phc` branches get retries
```
            enable_retry=false
            if [[ "$CIRCLE_BRANCH" == bump-phc/* ]]; then
              enable_retry=true
            fi
```

PR #1679, opened 2026-03-12
Add trackCustomPaywallImpression API

rickvdl:
## Summary
- Expose `trackCustomPaywallImpression` as a public API in Flutter
- Support optional `paywallId` and `offeringId` parameters via `CustomPaywallImpressionParams`
- If `offeringId` is not provided, the SDK uses the current offering identifier from the cache
- Pass full arguments dictionary to PHC on both iOS and Android instead of unpacking individual fields
- Add purchase tester screen with text fields for paywall ID and offering ID

## Changes
- **Dart:** `CustomPaywallImpressionParams` with `paywallId` and `offeringId`
- **iOS bridge:** Pass arguments dictionary directly to `[RCCommonFunctionality trackCustomPaywallImpression:]`
- **Android bridge:** Pass arguments map directly to `CommonKt.trackCustomPaywallImpression()`
- **Tests:** Unit tests for all parameter combinations
- **API tester:** Compile-time verification of all API shapes
- **Purchase tester:** Screen with text fields for paywall ID and offering ID

### Related PRs
- purchases-hybrid-common: https://github.com/RevenueCat/purchases-hybrid-common/pull/1565
- react-native-purchases: https://github.com/RevenueCat/react-native-purchases/pull/1679
- purchases-unity: https://github.com/RevenueCat/purchases-unity/pull/869
- purchases-capacitor: https://github.com/RevenueCat/purchases-capacitor/pull/725

### DIFF
--- a/android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
+++ b/android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
@@ -385,8 +385,7 @@ public class PurchasesFlutterPlugin implements FlutterPlugin, MethodCallHandler,
                 getCachedVirtualCurrencies(result);
                 break;
             case "trackCustomPaywallImpression":
-                String paywallId = call.argument("paywallId");
-                trackCustomPaywallImpression(paywallId, result);
+                trackCustomPaywallImpression(call.arguments(), result);
                 break;
             default:
                 result.notImplemented();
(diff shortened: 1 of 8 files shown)


android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
```diff
@@ -384,6 +384,10 @@ public void onMethodCall(@NonNull MethodCall call, @NonNull Result result) {
             case "getCachedVirtualCurrencies":
                 getCachedVirtualCurrencies(result);
                 break;
+            case "trackCustomPaywallImpression":
+                String paywallId = call.argument("paywallId");
```
tonidero: Similar to capacitor, could be interesting to move the parsing of the parameters to PHC, so it's all unified in a single place (same in iOS). But I think this does work for now :)

lib/purchases_flutter.dart
```diff
@@ -1380,6 +1380,33 @@ class Purchases {
     return VirtualCurrencies.fromJson(Map<String, dynamic>.from(result));
   }
 
+  ///================================================================================
+  /// Custom Paywall Tracking
+  ///================================================================================
+
+  /// **Experimental**: This API is experimental and may be changed or removed
```
tonidero, line 1387: I guess this is not really enforced in any way... So my thinking is that we shouldn't consider this experimental and if we need to do a breaking change, deprecate instead... But yeah I don't think this should remain experimental for long in any case.

rickvdl: Removed the experimental annotation here as discussed :)

android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
rickvdl: Done!

PR #1750, opened 2026-05-11
feat: ad tracking API

peterporfy:
## Summary

Adds five new methods to track ad lifecycle events through RevenueCat, exposed via `Purchases.adTracker`:

- `Purchases.adTracker.trackAdDisplayed(AdDisplayedData)`
- `Purchases.adTracker.trackAdOpened(AdOpenedData)`
- `Purchases.adTracker.trackAdLoaded(AdLoadedData)`
- `Purchases.adTracker.trackAdRevenue(AdRevenueData)`
- `Purchases.adTracker.trackAdFailedToLoad(AdFailedToLoadData)`

All methods and types are annotated `@experimental`.

## Key decisions

**`PurchasesAdTracker` singleton** — methods are grouped on `Purchases.adTracker` (private constructor, defined in the same file as `Purchases` to share `_channel`) rather than as flat static methods on `Purchases`. Matches the pattern used in other SDKs.

**Flat separate data classes** — `AdDisplayedData`, `AdOpenedData`, `AdLoadedData`, `AdRevenueData`, `AdFailedToLoadData` are independent types rather than a shared base class. Mirror the iOS and Android SDK design: identical fields today, kept separate for semantic clarity.

**Typed wrappers** — `AdMediatorName`, `AdFormat`, `AdRevenuePrecision` are open-ended value types with predefined constants, not enums. Matches the native SDK pattern since ad networks and formats aren't a closed set.

**`trackAdFailedToLoad` has no `impressionId` or `networkName`** — consistent with PHC: a failed load produces neither an impression nor a serving network; `mediatorErrorCode` is added instead.

**iOS availability guard** — `@available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *)` matches PHC's own guard; logs a warning on older OS versions.

## Testing

- API surface covered by `api_tester` (compile-time shape checks)
- Method channel wiring covered by unit tests
- Manually tested with a companion Flutter sample app (google_mobile_ads, all 5 ad formats)

---

> [!NOTE]
> **Medium Risk**
> New experimental public API and ad revenue event forwarding to native SDKs; low direct purchase/auth impact but incorrect payloads could skew analytics.
> 
> **Overview**
> Adds an **experimental** ad lifecycle analytics surface on **`Purchases.adTracker`**, with five events: displayed, opened, loaded, revenue, and failed-to-load. Each call takes a dedicated payload type (`AdDisplayedData`, `AdRevenueData`, etc.) serialized via `toMap()` over the existing method channel to **Purchases Hybrid Common** on **Android** and **iOS** (iOS 15+; older OS versions log and no-op). **Web** treats these channel methods as no-ops.
> 
> New **open-ended** value types (`AdMediatorName`, `AdFormat`, `AdRevenuePrecision`) mirror native SDKs; `AdFailedToLoadData` omits impression/network fields and supports optional `mediatorErrorCode`. API shape is locked by **api_tester** compile checks and **unit tests** for channel arguments; `meta` is added for `@experimental`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 515a8a78f978228c6ab09e60911f176e9e0b2d9e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


api_tester/lib/api_tests/purchases_flutter_api_test.dart
```diff
@@ -634,6 +634,53 @@ class _PurchasesFlutterApiTest {
         await Purchases.getCachedVirtualCurrencies();
   }
 
+  void _checkTrackAdDisplayed() {
+    Future<void> future = Purchases.trackAdDisplayed(const AdDisplayedData(
```
tonidero: I know this is in line with what we're doing in other APIs, but to avoid bloating the `Purchases` API, I think it would be ideal to keep the ad methods within a separate object, like we're doing in the natives, even if they remain static methods, so these can be for example something like: 
- `Purchases.AdTracker.trackAdDisplayed` (in a namespace if possible) or 
- `PurchasesAdTracker` (if not possible to namespace). 
 
I understand this is a bit trickier... But I think it could help to avoid some bloating issues. I also think it's fine even if internally we need to reuse the same channel to communicate with the kotlin/swift layer. Lmk what you think! 🙏

lib/models/ad_displayed_data.dart
```diff
@@ -0,0 +1,29 @@
+import 'ad_format.dart';
+import 'ad_mediator_name.dart';
+
+class AdDisplayedData {
```
tonidero, line 7: Just to check, these are not experimental (we don't currently have a good way to mark these as experimental in dart). Just to confirm, but are we ok with launching these as stable? And if so, should we make the native APIs stable already?

lib/models/ad_displayed_data.dart
```diff
@@ -0,0 +1,29 @@
...
+    this.placement,
+    required this.adUnitId,
+    required this.impressionId,
+  });
+
+  Map<String, dynamic> toMap() => {
+        if (networkName != null) 'networkName': networkName,
+        'mediatorName': mediatorName.value,
+        'adFormat': adFormat.value,
+        if (placement != null) 'placement': placement,
```
tonidero: Not a huge deal... but shoudl we keep the properties but just with a null value in these cases? I find I prefer to have the properties there even as null rather than not having them at all.

tonidero (commented):
Looking great! Nothing too blocking, just some thoughts. Let me know what you think!

api_tester/lib/api_tests/purchases_flutter_api_test.dart
peterporfy: I agree that it would be more in line with existing SDKs. Although `Purchases.AdTracker` won't work I think, but we can do a singleton `Purchases.adTracker` on the static `Purchases` - or a separate static `PurchasesAdTracker` as you suggested. I am leaning towards the second. Wdyt?

lib/models/ad_displayed_data.dart
peterporfy: That's a good question, I will check with the team.

lib/models/ad_displayed_data.dart
peterporfy: I removed those null checks and also removed them from android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java - it works just fine with nulls so it looks simpler like this, thanks!

android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
```diff
@@ -878,6 +893,31 @@ private void trackCustomPaywallImpression(Map<String, Object> arguments, final R
...
+
+    private void trackAdRevenue(Map<String, Object> arguments, final Result result) {
+        CommonKt.trackAdRevenue(arguments);
+        result.success(null);
+    }
+
+    private void trackAdFailedToLoad(Map<String, Object> arguments, final Result result) {
+        CommonKt.trackAdFailedToLoad(arguments);
+        result.success(null);
+    }
```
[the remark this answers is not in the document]
peterporfy: I tested the android locally with null values for ad placement and it worked just fine. What do you think @tonidero ?

lib/models/ad_displayed_data.dart
peterporfy: What if we don't document these features in flutter/unity/reactnative etc until it is not marked as stable in the core sdks? Would that be good enough to signal they are experimental/not official yet? Or we must have it stable in ios/android before we add anything here? Wdyt?

api_tester/lib/api_tests/purchases_flutter_api_test.dart
tonidero: IMO, we should avoid having static APIs... I think we have this most for legacy reasons... So I kinda would lean towards the singleton approach if possible. Feels to me that it's easier to test and iterate on  🙏

android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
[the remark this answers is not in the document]
tonidero: Hmm right... I'm not sure why we added those null filters in those other methods... But IMO, we should keep the least possible logic in the hybrid layers, and I think passing null values explicitedly is cleaner, so as long as we have tested it, I think this is fine!

lib/models/ad_displayed_data.dart
tonidero: Even if we don't publicize them, if we release these APIs in a release as stable, I think we should consider them stable and basically, we should try very very hard to avoid breaking changes, or they would need to be a major.

So I think the proper way is to use meta's library and add an experimental annotation to these APIs: https://api.flutter.dev/flutter/meta/experimental-constant.html. We haven't used this before but I think it's probably worth using it here...

api_tester/lib/api_tests/purchases_flutter_api_test.dart
peterporfy: Updated to singleton - indeed looks better

lib/models/ad_displayed_data.dart
peterporfy: Oh that's nice - I've added the annotation, thanks!

ios/purchases_flutter/Sources/purchases_flutter/PurchasesFlutterPlugin.m
```diff
@@ -715,6 +725,51 @@ - (void)beginRefundRequestForEntitlement:(NSString *)entitlementIdentifier resul
...
+}
+
+- (void)trackAdFailedToLoad:(NSDictionary *)arguments result:(FlutterResult)result {
+    if (@available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *)) {
+        [RCCommonFunctionality trackAdFailedToLoad:[arguments mappingNSNullToNil]];
+    } else {
+        NSLog(@"[Purchases] Warning: tried to call trackAdFailedToLoad, but it's only available on iOS 15.0 or greater.");
+    }
+    result(nil);
+}
```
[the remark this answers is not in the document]
peterporfy: I think it is a different situation than android. The intention here is to map nsnull to nil - and a nil will probably be omitted, but that is a side effect and not an extra logic like what we did before (manually checking for nulls on optional params).

lib/models/ad_failed_to_load_data.dart
```diff
@@ -0,0 +1,29 @@
...
+  });
+
+  Map<String, dynamic> toMap() => {
+        'mediatorName': mediatorName.value,
+        'adFormat': adFormat.value,
+        'placement': placement,
+        'adUnitId': adUnitId,
+        'mediatorErrorCode': mediatorErrorCode,
+      };
+}
```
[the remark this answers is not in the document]
peterporfy: It is intentional. We don't know (and could be multiple) which network couldn't load.

api_tester/lib/api_tests/models/ad_tracking_api_test.dart
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  void _checkProperties(AdFailedToLoadData data) {
+    AdMediatorName mediatorName = data.mediatorName;
+    AdFormat adFormat = data.adFormat;
+    String? placement = data.placement;
+    String adUnitId = data.adUnitId;
+    int? mediatorErrorCode = data.mediatorErrorCode;
+    Map<String, dynamic> map = data.toMap();
+  }
```
facumenzella, line 154: I see we're not including `String? networkName = data.networkName` for this one. Is it intentional? AdDisplayedData, AdOpenedData, and AdLoadedData all have it.

api_tester/lib/api_tests/purchases_flutter_api_test.dart
```diff
@@ -634,6 +634,53 @@ class _PurchasesFlutterApiTest {
         await Purchases.getCachedVirtualCurrencies();
   }
 
+  void _checkTrackAdDisplayed() {
+    Future<void> future = Purchases.adTracker.trackAdDisplayed(const AdDisplayedData(
```
rickvdl, line 638: The PR description mentions the following
> All five methods are flat static methods on Purchases - following how it looks like on PHC.

But these appear to have been exposed through `adTracker` instead. Should we update the PR description?

api_tester/lib/api_tests/purchases_flutter_api_test.dart
```diff
@@ -634,6 +634,53 @@ class _PurchasesFlutterApiTest {
...
+      mediatorName: AdMediatorName.adMob,
+      adFormat: AdFormat.banner,
+      adUnitId: 'unit-1',
+      impressionId: 'imp-1',
+    ));
+  }
+
+  void _checkTrackAdOpened() {
+    Future<void> future = Purchases.adTracker.trackAdOpened(const AdOpenedData(
```
rickvdl, line 658: Should we also add (in a separate test case?) the optional fields here? `networkName` and `placement` IIRC

rickvdl: Same for `trackAdLoaded` and `trackAdRevenue`

rickvdl (commented):
Looks good! Two small comments, and one other thing;
I think we should add `pr:feat` instead of `pr:other` since we're adding new APIs

api_tester/lib/api_tests/purchases_flutter_api_test.dart
peterporfy: Indeed it was changed later from static to singletion, thanks! Updated

api_tester/lib/api_tests/models/ad_tracking_api_test.dart
peterporfy: Yes, intentional. We don't have a network when we fail to load (we only know the network when one was successfully selected by the process). I've updated the PR desc.

api_tester/lib/api_tests/purchases_flutter_api_test.dart
peterporfy: Added new cases where we include optionals 👍

rickvdl (approved):
Nice work!

PR #1781, opened 2026-06-12
Add presentedOfferingContext support to custom paywall impression events

rickvdl:
Based on PHC PR https://github.com/RevenueCat/purchases-hybrid-common/pull/1665.

## API

We will now send the `presentedOfferingContext` from the passed `Offering` along with the event. This PR adds the new API variant that takes an `Offering` instead of a plain string `offeringId`. The latter is now deprecated in favor of this new API variant. We will derive the `presentedOfferingContext` from this `Offering` and send it along with the request.

```
await Purchases.trackCustomPaywallImpression(
  params: CustomPaywallImpressionParams(paywallId: "", offering: offering),
)
```

---

> [!NOTE]
> **Low Risk**
> Analytics-only API extension with backward-compatible deprecation; no purchase, auth, or billing flow changes, though impression event shape changes may affect downstream analytics if callers relied on explicit null fields.
> 
> **Overview**
> **Custom paywall impression tracking** now accepts an `Offering` on `CustomPaywallImpressionParams` and forwards **placement/targeting** via `presentedOfferingContext` (taken from the offering’s first package when packages exist). String `offeringId` is **deprecated** but still supported; when both are set, the offering’s identifier wins.
> 
> `Purchases.trackCustomPaywallImpression` no longer always sends `paywallId`/`offeringId` as null—it builds a sparse map and only includes set fields. With no offering context, docs clarify that the native SDK picks the current offering.
> 
> **Android/iOS bridges** harden the channel path: null method arguments default to `{}`, and payloads strip null/`NSNull` values (Android recursively for nested maps/lists; iOS reuses `mappingNSNullToNil`).
> 
> Tests, API tester coverage, and the purchase_tester screen were updated to exercise tracking with/without an offering and the new payload shape.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3c3e4c22dbf12a99b6ec592d37cf1cf31e3f1886. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
+++ b/android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
@@ -384,9 +384,13 @@ public class PurchasesFlutterPlugin implements FlutterPlugin, MethodCallHandler,
             case "getCachedVirtualCurrencies":
                 getCachedVirtualCurrencies(result);
                 break;
-            case "trackCustomPaywallImpression":
-                trackCustomPaywallImpression(call.arguments(), result);
+            case "trackCustomPaywallImpression": {
+                Map<String, Object> arguments = call.arguments();
+                trackCustomPaywallImpression(
+                        arguments != null ? arguments : new HashMap<>(),
+                        result);
                 break;
+            }
             case "trackAdDisplayed":
                 trackAdDisplayed(call.arguments(), result);
                 break;
@@ -883,14 +887,46 @@ public class PurchasesFlutterPlugin implements FlutterPlugin, MethodCallHandler,
     }
 
     private void trackCustomPaywallImpression(Map<String, Object> arguments, final Result result) {
-        HashMap<String, Object> data = new HashMap<>();
-        for (Map.Entry<String, Object> entry : arguments.entrySet()) {
-            if (entry.getValue() != null) {
-                data.put(entry.getKey(), entry.getValue());
+        CommonKt.trackCustomPaywallImpression(mapWithoutNullValues(arguments));
+        result.success(null);
+    }
+
+    private static HashMap<String, Object> mapWithoutNullValues(@Nullable Map<String, Object> map) {
+        HashMap<String, Object> filteredMap = new HashMap<>();
+        if (map != null) {
+            for (Map.Entry<String, Object> entry : map.entrySet()) {
+                Object value = valueWithoutNullValues(entry.getValue());
+                if (value != null) {
+                    filteredMap.put(entry.getKey(), value);
+                }
             }
         }
-        CommonKt.trackCustomPaywallImpression(data);
-        result.success(null);
+        return filteredMap;
+    }
+
+    private static Object valueWithoutNullValues(@Nullable Object value) {
+        if (value instanceof Map) {
+            HashMap<String, Object> filteredMap = new HashMap<>();
+            Map<?, ?> originalMap = (Map<?, ?>) value;
+            for (Map.Entry<?, ?> entry : originalMap.entrySet()) {
+                Object filteredValue = valueWithoutNullValues(entry.getValue());
+                if (entry.getKey() instanceof String && filteredValue != null) {
+                    filteredMap.put((String) entry.getKey(), filteredValue);
+                }
+            }
+            return filteredMap;
+        }
+        if (value instanceof List) {
+            ArrayList<Object> filteredList = new ArrayList<>();
+            for (Object item : (List<?>) value) {
+                Object filteredItem = valueWithoutNullValues(item);
+                if (filteredItem != null) {
+                    filteredList.add(filteredItem);
+                }
+            }
+            return filteredList;
+        }
+        return value;
     }
 
     private void trackAdDisplayed(Map<String, Object> arguments, final Result result) {
--- a/android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
+++ b/android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
@@ -887,17 +887,17 @@ public class PurchasesFlutterPlugin implements FlutterPlugin, MethodCallHandler,
     }
 
     private void trackCustomPaywallImpression(Map<String, Object> arguments, final Result result) {
-        CommonKt.trackCustomPaywallImpression(mapWithoutNullValues(arguments));
+        CommonKt.trackCustomPaywallImpression(stringKeyedMapWithoutNullValues(arguments));
         result.success(null);
     }
 
-    private static HashMap<String, Object> mapWithoutNullValues(@Nullable Map<String, Object> map) {
+    private static HashMap<String, Object> stringKeyedMapWithoutNullValues(@Nullable Map<?, ?> map) {
         HashMap<String, Object> filteredMap = new HashMap<>();
         if (map != null) {
-            for (Map.Entry<String, Object> entry : map.entrySet()) {
-                Object value = valueWithoutNullValues(entry.getValue());
-                if (value != null) {
-                    filteredMap.put(entry.getKey(), value);
+            for (Map.Entry<?, ?> entry : map.entrySet()) {
+                Object filteredValue = valueWithoutNullValues(entry.getValue());
+                if (entry.getKey() instanceof String && filteredValue != null) {
+                    filteredMap.put((String) entry.getKey(), filteredValue);
                 }
             }
         }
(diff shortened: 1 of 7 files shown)


android/build.gradle
```diff
@@ -3,7 +3,7 @@ version '10.2.3'
 
 buildscript {
     ext.kotlin_version = '1.8.22'
-    ext.common_version = '18.14.1'
+    ext.common_version = '18.15.0'
```
tonidero: Hmm should we leave the PHC bump to a separate PR?

lib/models/map_helpers.dart
```diff
@@ -74,3 +74,19 @@ PeriodUnit periodUnitFromJson(dynamic value) {
...
+    );
+
+dynamic _valueWithoutNullValues(dynamic value) {
+  if (value is Map) {
+    return mapWithoutNullValues(Map<String, dynamic>.from(value));
+  }
+  if (value is List) {
+    return value.map(_valueWithoutNullValues).toList();
+  }
+  return value;
```
tonidero: Should we also filter for `value` being null? Though I guess that means dynamic might not catch it not support it not sure...

revenuecat_examples/purchase_tester/android/app/build.gradle
```diff
@@ -45,7 +45,7 @@ android {
     }
 
     defaultConfig {
-        applicationId "com.revenuecat.purchases_sample"
+        applicationId "com.revenuecat.paywall_tester"
```
tonidero: Hmm we might not won't to push this?

revenuecat_examples/purchase_tester/ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -431,7 +431,7 @@
 					"$(inherited)",
 					"$(PROJECT_DIR)/Flutter",
 				);
-				PRODUCT_BUNDLE_IDENTIFIER = com.revenuecat.sampleapp;
+				PRODUCT_BUNDLE_IDENTIFIER = com.revenuecat.PaywallsTester;
```
tonidero: Same here and some below

revenuecat_examples/purchase_tester/lib/src/constant.dart
```diff
@@ -1,8 +1,8 @@
 //TO DO: add the Apple API key for your app from the RevenueCat dashboard: https://app.revenuecat.com
-const appleApiKey = 'appl_api_key';
+const appleApiKey = 'appl_SYkbbYptqjDadVkQeBUtXxVxCcw';
```
tonidero: Same

CHANGELOG-LATEST.md
```diff
@@ -1,4 +1,7 @@
 ## RevenueCat SDK
+### ✨ New Features
+* Add `offering` support to `trackCustomPaywallImpression` so custom paywall events include presented offering context when available.
```
tonidero: This shouldn't be added here :)

tonidero (commented):
Some comments, but the main code looks good

lib/models/map_helpers.dart
rickvdl: Right, good catch. We're checking it in `mapWithoutNullValues` but indeed that assuming it's only called from there. I've moved to code to `mapWithoutNullValues` to make it explicit.

android/build.gradle
rickvdl: Sorry 🤦 accidentally committed unrelated changes. Should've checked this before pushing. Fixed it and the PR should be clean now.

lib/models/map_helpers.dart
rickvdl: Actually changed my mind on this and moved the filtering to the native layer like we did for RN, using existing helpers

android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
```diff
@@ -485,15 +489,7 @@ private void syncAttributesAndOfferingsIfNeeded(final Result result) {
...
-        HashMap<String, Object> filteredData = new HashMap<>();
-        if (data != null) {
-            for (Map.Entry<String, Object> entry : data.entrySet()) {
-                if (entry.getValue() != null) {
-                    filteredData.put(entry.getKey(), entry.getValue());
-                }
-            }
-        }
-        CommonKt.setAppstackAttributionParams(filteredData, getOnResult(result));
+        CommonKt.setAppstackAttributionParams(mapWithoutNullValues(data), getOnResult(result));
```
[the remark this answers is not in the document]
rickvdl: This should now be resolved

android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
```diff
@@ -883,14 +887,46 @@ private void getCachedVirtualCurrencies(final Result result) {
...
+                Object value = valueWithoutNullValues(entry.getValue());
+                if (value != null) {
+                    filteredMap.put(entry.getKey(), value);
+                }
             }
         }
-        CommonKt.trackCustomPaywallImpression(data);
-        result.success(null);
+        return filteredMap;
+    }
```
ajpallares, line 905: Total nit, but I wonder if we could allow receiving any map in this method.
```suggestion
    private static HashMap<String, Object> mapWithoutNullValues(@Nullable Map<?, ?> map) {
        HashMap<String, Object> filteredMap = new HashMap<>();
        if (map != null) {
            for (Map.Entry<?, ?> entry : map.entrySet()) {
                Object filteredValue = valueWithoutNullValues(entry.getValue());
                if (entry.getKey() instanceof String && filteredValue != null) {
                    filteredMap.put((String) entry.getKey(), filteredValue);
                }
            }
        }
        return filteredMap;
    }
```

Then, below, we would just do

```
    private static Object valueWithoutNullValues(@Nullable Object value) {
        if (value instanceof Map) {
            return mapWithoutNullValues((Map<?, ?>) value);
        }
        ...
```

Maybe changing the method name as we're filtering out non-String keys as well as null values. `stringKeyedMapWithoutNullValues`?

api_tester/lib/api_tests/purchases_flutter_api_test.dart
```diff
@@ -763,16 +776,26 @@ class _PurchasesFlutterApiTest {
...
+    CustomPaywallImpressionParams paramsWithNullIds =
+        const CustomPaywallImpressionParams(
+      paywallId: null,
+      offeringId: null,
+    );
+
     CustomPaywallImpressionParams paramsWithOffering =
+        CustomPaywallImpressionParams(offering: offering);
+
+    CustomPaywallImpressionParams paramsWithOfferingId =
```
ajpallares, line 804: I think we're missing the constructor containing `paywallId` and `offering` parameters

ajpallares: And, in theory, now we'd be supporting the constructors with the two `offeringId` and `offering` parameters, but I wouldn't add an API test for those, since they clearly are never supposed to be sent together

lib/purchases_flutter.dart
```diff
@@ -1447,20 +1447,45 @@ class Purchases {
...
-          'paywallId': params?.paywallId,
-          'offeringId': params?.offeringId,
-        },
+        _customPaywallImpressionData(params),
       );
 
+  static Map<String, dynamic> _customPaywallImpressionData(
+    CustomPaywallImpressionParams? params,
+  ) {
+    final offering = params?.offering;
```
ajpallares, line 1467: Should we add a warning if both `offering` and `offeringId` are passed saying that the passed `offeringId` will be ignored? (I'd probably add it even if `offering.identifier == offeringId`)

ajpallares (approved):
Looks good! Just some non-blocking comments

android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
rickvdl: Thanks for the suggestion! Updated it with just that :)

api_tester/lib/api_tests/purchases_flutter_api_test.dart
rickvdl: Good catch! added the missing tests

lib/purchases_flutter.dart
rickvdl: Hm right, that's an interesting one, however I think since the inits that take the `offeringId` are already deprecated that should already warn the developer. Not sure if the log line (which has to be going through `debugPrint` instead of our logging infrastructure from the Dart-side) would add much value 🤔 Wdyt?

ajpallares: Ohh true. The deprecation message should be enough then 👍

PR #1785, opened 2026-06-15
Fix web support under strict CSP (Chrome Extension Manifest V3)

### LINKED ISSUE #1784: `purchases_flutter` web breaks Chrome Extension Manifest V3

# ‼️ Required data ‼️ #
**Do not remove any of the steps from the template below. If a step is not applicable to your issue, please leave that step empty.**

There are a lot of things that can contribute to things not working. Having a very basic understanding of your environment will help us understand your issue faster!

## Environment ##
[✓] Flutter (Channel stable, 3.44.1, on macOS 26.4.1 25E253 darwin-arm64, locale en-TW)
    • Flutter version 3.44.1 on channel stable at /Users/yayo/fvm/versions/3.44.1
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision 924134a44c (2 weeks ago), 2026-05-29 12:13:22 -0400
    • Engine revision c416acfeb8
    • Dart version 3.12.1
    • DevTools version 2.57.0

[✓] Android toolchain - develop for Android devices (Android SDK version 36.1.0)
[✓] Xcode - develop for iOS and macOS (Xcode 26.5)
[✓] Chrome - develop for the web
    • Chrome at /Applications/Google Chrome.app/Contents/MacOS/Google Chrome

[✓] Connected device (3 available)
    • sdk gphone64 arm64 (mobile) • emulator-5554 • android-arm64  • Android 16 (API 36) (emulator)
    • macOS (desktop)             • macos         • darwin-arm64   • macOS 26.4.1 25E253 darwin-arm64
    • Chrome (web)                • chrome        • web-javascript • Google Chrome 149.0.7827.104

[✓] Network resources
    • All expected network resources are available.

• No issues found!

purchases_flutter: ^10.2.3

Chrome Extension Manifest V3 popup running in Google Chrome 149.0.7827.104 on macOS 26.4.1.

This happens every time the Chrome Extension MV3 popup opens when purchases_flutter is included in the Flutter web build. This affects the extension at startup

- [x] [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue
Executing inline script violates the following Content Security Policy directive:
"script-src 'self' 'wasm-unsafe-eval'".
Either the 'unsafe-inline' keyword, a hash ('sha256-LypY5zNcO7N91nG7Ylv6EDF9Q/VBpPwt9bhDrAKmGfE='), or a nonce ('nonce-...') is required to enable inline execution.
The action has been blocked.

1. Create a Flutter app that depends on: purchases_flutter: ^10.2.3
2. Build chrome extension `flutter build web --wasm --no-web-resources-cdn --pwa-strategy=none`
3. Use following manifest:
```
{
  "content_security_policy": {
    "extension_pages": "script-src 'self' 'wasm-unsafe-eval'; object-src 'self';"
  },
  "action": {
    "default_popup": "index.html"
  },
  "manifest_version": 3
}
```
4. Open the extension and it will crash.

## Describe the bug ##
`purchases_flutter` web currently injects an inline script and dynamically imports RevenueCat JavaScript from jsDelivr during web plugin registration. This breaks Flutter web apps packaged as Chrome Extension Manifest V3 popups because MV3 extension pages cannot run inline scripts and cannot rely on remote hosted executable code.

## Additional context ##
The same app works when `purchases_flutter` is excluded from the web extension build. After excluding it, the generated main.dart.js no longer contains `after_rc_load_callback`, `purchases-js-hybrid-mappings`, or `PurchasesFlutterPlugin`, and the Chrome extension popup loads correctly.

ajpallares:
### Motivation
On web, the plugin injected an inline script that dynamically imported the RevenueCat web SDK from a CDN. Chrome Extension Manifest V3 pages forbid both inline scripts and remotely-hosted code, so a Flutter web app packaged as an MV3 extension crashes on startup. Resolves #1784.

### Summary
- Vendor the `purchases-js-hybrid-mappings` UMD bundle and load it via an external, same-origin `<script src>` instead of an inline script + CDN import.
- Keep the vendored bundle in sync with the pinned `purchases-hybrid-common` version via the `update_hybrid_common` fastlane lane.
- Restrict the asset to web builds with `platforms: [web]`.
- Add a CI check that fails if the vendored bundle doesn't byte-match the published bundle for the pinned version (tested the failure [here](https://app.circleci.com/pipelines/github/RevenueCat/purchases-flutter/5883/workflows/c43ae154-5ad7-4f3a-b9e1-d0a4de055d5d/jobs/47734))

### Notes
- `platforms` requires Flutter 3.41+. On Flutter 3.22–3.40 the key is ignored, so the bundle is included in mobile/desktop builds (~700 KB) but nothing breaks; on 3.41+ it is web-only.


fastlane/Fastfile
```diff
@@ -241,6 +251,37 @@ def get_root_folder
...
 
+# Downloads the purchases-js-hybrid-mappings UMD bundle matching the given
+# purchases-hybrid-common version and stores it as the vendored Flutter web
+# asset. The npm package is versioned in lockstep with purchases-hybrid-common.
+# When `commit_and_push` is true, commits the updated asset to the current
+# branch (the bump branch created by bump_phc_version) and pushes it.
+def update_web_hybrid_mappings_bundle(version:, commit_and_push:)
+  relative_path = 'assets/web/purchases_js_hybrid_mappings.js'
+  asset_path = File.join(get_root_folder, relative_path)
+  download_url = "https://cdn.jsdelivr.net/npm/@revenuecat/purchases-js-hybrid-mappings@#{version}/dist/index.umd.js"
```
ajpallares: I think it's safe to assume that this file is available after PHC was released, right?

fastlane/Fastfile
```diff
@@ -241,6 +250,36 @@ def get_root_folder
...
+
+  UI.message("ℹ️  Downloading purchases-js-hybrid-mappings #{version} bundle from #{download_url}")
+  Actions.sh("curl", "-fsSL", download_url, "-o", asset_path)
```
rickvdl, line 312: Could we somehow add a checksum or other check here to ensure that we've included the right thing? Preventing possible supply chain attacks or similar? Alternatively, could we get the same file from our own Git repo instead of the CDN?

Although tbf we previously linked to the CDN as well, but we're now bundling it.

ajpallares: > Alternatively, could we get the same file from our own Git repo instead of the CDN?

I wish this was possible. But the `index.umd.js` is a build artifact produced by `yarn build` at release and published only to npm. So pulling it from the PHC repo isn't directly possible today 😞

ajpallares: > Could we somehow add a checksum or other check here to ensure that we've included the right thing? Preventing possible supply chain attacks or similar?

Good point. I would say there's virtually no regression here since we're now downloading it once right after PHC is released (less likely to be supply-chain attacked I would say). But it's probably a good improvement to do.

rickvdl: That's true as well, let's keep it as is and we can always follow up.

fastlane/Fastfile
tonidero: Yeah it should be

fastlane/Fastfile
tonidero: It could be good to add a test that verifies that the `assets/web/purchases_js_hybrid_mappings.js` file matches what is online for the current version. This way we make sure it's not modified through other methods.

lib/web/purchases_flutter_web.dart
```diff
@@ -12,45 +12,62 @@ import '../purchases_flutter.dart';
...
-        globalContext.delete('after_rc_load_callback'.toJS);
-        _initCompleter?.complete();
-      }).toJS,
-    );
+    // Load the bundle through an external <script src> pointing at a
+    // same-origin asset. Shipping the code inside the app and loading it as an
+    // external (non-inline) script avoids both inline-script execution and
+    // remotely-hosted code, which are disallowed by strict Content Security
+    // Policies such as the one enforced on Chrome Extension Manifest V3 pages.
+    // The UMD bundle registers `window.PurchasesHybridMappings` when it runs.
```
tonidero, line 48: Probably not needed for this PR, but we should revisit how we integrate with the web sdk now that we're embedding the JS sdk... I think there are cleaner ways to use JS code than what we currently have 🙏

pubspec.yaml
```diff
@@ -39,3 +39,12 @@ flutter:
       web:
         pluginClass: PurchasesFlutterPlugin
         fileName: web/purchases_flutter_web.dart
+  assets:
+    # Vendored RevenueCat web SDK (purchases-js-hybrid-mappings UMD bundle).
+    # Bundled locally so it can be loaded same-origin, which is required by
+    # strict CSP environments such as Chrome Extension Manifest V3 pages that
+    # disallow remotely-hosted code. Kept in sync with the bumped
+    # purchases-hybrid-common version by the fastlane `update_hybrid_common` lane.
+    - path: assets/web/purchases_js_hybrid_mappings.js
```
tonidero: Should we filter the entire `web` folder?

tonidero (commented):
Nice one! I love this change, though we should test this thoroughly to make sure it works. Also we can probably make this much nicer after this, but that can be a separate PR.

pubspec.yaml
ajpallares: Good point. Done in https://github.com/RevenueCat/purchases-flutter/pull/1785/commits/dc880926ab3371119e0c5f9b064b477ca778fcea

fastlane/Fastfile
ajpallares: Great idea! Done in https://github.com/RevenueCat/purchases-flutter/pull/1785/commits/d22854fa6822e67c5c34fc64884c5dba26ba425b

rickvdl (approved):
Nice work!

PR #1801, opened 2026-06-29
feat(ads): reward-verification primitives for Flutter

peterporfy:
## Description

Adds the Flutter API for ad reward verification, with the two-primitives SSV flow.

We do not provide AdMob adapters for Flutter and these two primitives makes it possible to still use rewarded ads with a Flutter admob sdk. An example app will come in a later PR.

Verified locally e2e with both ios and android simulators.

---

> [!NOTE]
> **Medium Risk**
> New experimental path that polls RevenueCat and interprets granted virtual currency or entitlements; incorrect integration could mishandle rewards, though it does not change core purchase or auth flows.
> 
> **Overview**
> Adds an **experimental** Flutter API for rewarded-ad **server-side verification (SSV)** without a bundled AdMob adapter: apps call **`Purchases.generateRewardVerificationToken`** after load (using the ad **`impressionId`**) and **`Purchases.pollRewardVerification`** when the network’s reward callback fires (using **`clientTransactionId`**).
> 
> New **`@experimental`** models describe the token (`customData`, `clientTransactionId`, `appUserID`), poll results (`failed`, primary `reward`, `moreRewards`), and typed **`VerifiedReward`** outcomes (virtual currency, entitlement, no reward, unsupported). Android and iOS plugins forward new method-channel handlers to hybrid common; **web** rejects both methods. API tester hooks and unit tests cover parsing and public API surface.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f00425abdb13f183def538e122107035b84381d4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


peterporfy:
This change is part of the following stack:

- #1801 ◀
    - #1802

<sub>Change managed by [git-spice](https://abhinav.github.io/git-spice/).</sub>

ios/purchases_flutter/Sources/purchases_flutter/PurchasesFlutterPlugin.m
```diff
@@ -292,6 +292,11 @@ - (void)handleMethodCall:(FlutterMethodCall *)call
         [self trackAdLoaded:arguments result:result];
     } else if ([@"trackAdFailedToLoad" isEqualToString:call.method]) {
         [self trackAdFailedToLoad:arguments result:result];
+    } else if ([@"generateRewardVerificationToken" isEqualToString:call.method]) {
+        result([RCCommonFunctionality generateRewardVerificationTokenWithImpressionId:arguments[@"impressionId"]]);
+    } else if ([@"pollRewardVerification" isEqualToString:call.method]) {
+        [RCCommonFunctionality pollRewardVerificationWithClientTransactionId:arguments[@"clientTransactionId"]
+                                                                  completion:[self getResponseCompletionBlock:result]];
```
tonidero: Hmm I guess until now we've always extracted these into separate functions, mostly so we can keep the arguments parsing outside of this huge if-else I guess 😅... Not a blocker though.

lib/models/verified_reward.dart
```diff
@@ -0,0 +1,44 @@
...
+  final String? code;
+
+  /// Virtual-currency amount (`virtual_currency` only).
+  final int? amount;
+
+  /// Entitlement identifier (`entitlement` only).
+  final String? identifier;
+
+  /// Entitlement expiration (`entitlement` only).
+  final DateTime? expiresAt;
```
tonidero, line 62: I do think, most of these properties should be encapsulated somehow... Maybe we need to define 2 subtypes for each of these, and create one or the other, depending on the `type`? Feels like a nicer API...

tonidero (commented):
Looking great! I think I would only change the API for `RewardVerificationResult`... feels to me it should change so it properly encapsulates the data for each type. We should just make sure not to use sealed classes or the sort, since that's not great API practice, but probably an abstract class with multiple implementations would work.

ios/purchases_flutter/Sources/purchases_flutter/PurchasesFlutterPlugin.m
peterporfy: Indeed - moved!

lib/models/verified_reward.dart
peterporfy: I was hesitant to replicate a union type in dart - but looking at the result I agree it looks better 👍

android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
```diff
@@ -407,6 +407,12 @@ public void onMethodCall(@NonNull MethodCall call, @NonNull Result result) {
             case "trackAdFailedToLoad":
                 trackAdFailedToLoad(call.arguments(), result);
                 break;
+            case "generateRewardVerificationToken":
+                result.success(CommonKt.generateRewardVerificationToken(call.argument("impressionId")));
+                break;
+            case "pollRewardVerification":
+                CommonKt.pollRewardVerification(call.argument("clientTransactionId"), getOnResult(result));
```
tonidero: Forgot to mention it, but I think we need to do the same change as in iOS (extract to methods to keep this small)

lib/models/verified_reward.dart
```diff
@@ -0,0 +1,80 @@
...
+/// final reward = result.reward;
+/// if (reward is VerifiedVirtualCurrencyReward) {
+///   grantCurrency(reward.code, reward.amount);
+/// } else if (reward is VerifiedEntitlementReward) {
+///   unlock(reward.identifier);
+/// }
+/// ```
+@experimental
+abstract class VerifiedReward {
+  const VerifiedReward();
```
tonidero: Hmm should we make this constructor private? I think it should be
```suggestion
 VerifiedReward._();
```
This way, devs can't create their own types.

lib/models/verified_reward.dart
```diff
@@ -0,0 +1,80 @@
...
+  }
+}
+
+/// A reward of [amount] units of the virtual currency [code].
+@experimental
+class VerifiedVirtualCurrencyReward extends VerifiedReward {
+  final String code;
+  final int amount;
+
+  const VerifiedVirtualCurrencyReward({
```
tonidero, line 50: I also wonder if we want to make these constructors internal/private... but could be useful for testing, so I'm ok leaving them

tonidero (approved):
I think this makes sense!

android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
peterporfy: done!

lib/models/verified_reward.dart
peterporfy: Good catch - it should be. Done!

lib/models/verified_reward.dart
peterporfy: I would keep it as is, we are never consuming it anyway so there is no potential harm. A user might indeed would like to use it to test some app code

PR #1829, opened 2026-07-23
Add internal support for enabling workflows

vegaro:
## What changed

Adds an internal-only `dangerousSettings.useWorkflows` option to `PurchasesConfiguration`,
defaulting to `false`. The Flutter bridge passes it through to native configuration:

- Android calls `DangerousSettings.forWorkflows(true)` when enabled, and keeps the existing
  `new DangerousSettings()` otherwise.
- iOS passes `nil` dangerous settings unless workflows are enabled, exactly as before. When they
  are enabled it calls `[RCDangerousSettings createDangerousSettingsWithAutoSyncPurchases:useWorkflows:]`,
  the Objective-C factory added in purchases-hybrid-common 18.24.0. The underlying
  `DangerousSettings(autoSyncPurchases:useWorkflows:)` initializer is `@_spi(Internal)` in
  purchases-ios, so hybrid-common bridges it into the Objective-C-visible surface on our behalf;
  no Swift shim is needed in this repo.

With the option omitted, neither platform's configuration changes at all: iOS still passes `nil`
and lets the native defaults apply, and Android's default branch is untouched.

This mirrors purchases-unity#996 and react-native-purchases#1847.

## Verification
- [x] `flutter test` (158 passing)
- [x] `flutter analyze lib`
- [x] `flutter build ios --no-codesign`
- [x] `bundle exec fastlane run_api_tests`

---

> [!NOTE]
> **Low Risk**
> Default-off internal configuration with no change to the standard configure path; only affects RevenueCat-internal workflows when explicitly enabled.
> 
> **Overview**
> Adds an **internal-only** `dangerousSettings.useWorkflows` flag on `PurchasesConfiguration` (defaults to `false`). `Purchases.configure` forwards it to native `setupPurchases` as `useWorkflows`.
> 
> On **Android**, workflows enabled uses `DangerousSettings.forWorkflows(true)` during configure; otherwise the plugin still passes a plain `DangerousSettings()` as before.
> 
> On **iOS**, `dangerousSettings` stays `nil` unless workflows are enabled, preserving existing native defaults. When enabled, configure uses the hybrid-common factory to turn on workflows.
> 
> Tests assert `useWorkflows: false` for existing setup paths and cover the enabled case.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d5a9f4d91827130b5ffbf1e5dee9f9e8388513ec. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


rickvdl (approved):

ios/purchases_flutter/Sources/purchases_flutter/PurchasesFlutterPlugin.m
```diff
@@ -306,22 +307,26 @@ - (void)setupPurchases:(NSString *)apiKey
...
     if ([appUserID isKindOfClass:NSNull.class]) {
         appUserID = nil;
     }
     if ([userDefaultsSuiteName isKindOfClass:NSNull.class]) {
         userDefaultsSuiteName = nil;
     }
 
+    RCDangerousSettings *dangerousSettings =
+        [RCDangerousSettings createDangerousSettingsWithAutoSyncPurchases:YES
+                                                             useWorkflows:useWorkflows];
```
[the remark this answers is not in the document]
vegaro: I think this is from a stale PR description

vegaro:
@cursoragent review

lib/models/purchases_configuration.dart
```diff
@@ -80,6 +80,17 @@ class PurchasesConfiguration {
...
+  /// Dangerous settings for the SDK. Internal RevenueCat use only.
+  DangerousSettings dangerousSettings = DangerousSettings();
+}
+
+/// Internal RevenueCat-only settings that may change without warning.
+class DangerousSettings {
+  /// Enables RevenueCat Workflows (multipage paywalls).
+  bool useWorkflows;
+
+  DangerousSettings({this.useWorkflows = false});
```
tonidero, line 93: I guess the bad thing about this is that we override the default from the natives... Maybe we could make the dangerous settings nullable, and in case it's null, we just don't pass it and rely on the natives defaults. But TBH, NABD. I'm ok merging this :)

android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
```diff
@@ -427,13 +429,16 @@ private void setupPurchases(String apiKey, String appUserID,
                 store = Store.AMAZON;
             }
 
+            DangerousSettings dangerousSettings = Boolean.TRUE.equals(useWorkflows)
```
tonidero: This is interesting 😅 Couldn't we just do `useWorkflows == true`? It still should work though, so NABD :)

tonidero (approved):
:shipit:

android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java
vegaro: yeah lol

vegaro: oh you know why? it's to prevent NPEs! `Boolean == boolean` triggers unboxing, so `useWorkflows == true` throws on null

tonidero: Ohh interesting haha. Sounds good then!

lib/models/purchases_configuration.dart
vegaro: What I've done is only pass it if `useWorkflows` is true. That way if not we rely on the defaults

tonidero: Hmm this means it can't be disabled from the hybrids... probably ok if we remove it quickly, but something we should be aware of.

PR #1854, opened 2026-08-06
Wire mise into CI for Ruby and Flutter

AlvaroBrey:
`mise.toml` already pinned Java + Flutter, but only 3 jobs used it, via bare `install-mise-tools` calls that silently installed everything declared. This finishes the wiring and adds Ruby.

- Flutter: pinned `mise.toml` to `3.44.2` (was `"latest"`). Replaced every remaining unpinned `setup-flutter` git-clone with mise's pinned Flutter (4 cocoapods/spm jobs, both maestro jobs, `update-hybrid-common-versions`, `android-integration-test`); removed the now-dead `setup-flutter` command.
- Converted the 3 pre-existing bare `install-mise-tools` calls to explicit `tools:` lists, dropping the parts each job didn't actually need.
- Ruby: was ambient everywhere; wired into every job using fastlane/bundler, bumped to 3.3.0 + `rb-readline` (fastlane#21794), orb `ruby_version` params bumped to match.
- Replaced a hand-rolled `gem install bundler && bundle install` in `run-maestro-e2e-tests-android` with the orb's `install-gem-unix-dependencies` command.

---

> [!NOTE]
> **Medium Risk**
> Touches release, integration, and E2E CI paths and bumps Ruby; misconfigured mise tool lists or lockfile drift could break pipelines until fixed.
> 
> **Overview**
> **CI** standardizes tool installs on **mise** instead of ad hoc Flutter git clones and ambient Ruby. The custom `setup-flutter` command is removed; integration, Maestro, and hybrid-common jobs now call `revenuecat/install-mise-tools` with explicit `tools:` lists (e.g. `flutter`, `java`, `ruby`).
> 
> **mise.toml** pins Flutter to **3.44.2** (was `latest`) and adds **Ruby 3.3.0** with precompiled installs. **mise.lock** is refreshed (Ruby entries, fewer musl platform rows). CircleCI orb jobs that use Ruby bump `ruby_version` to **3.3.0**; **Gemfile** adds `rb-readline` for a fastlane/macOS Ruby 3.3 issue.
> 
> **Android Maestro** drops hand-rolled `gem install bundler` and uses the orb’s gem install step. A new **GitHub Actions** workflow on Renovate PRs runs `fastlane sync_mise_lock` to regenerate **mise.lock** when Renovate can’t run `mise lock`. **renovate.json** enables grouped minor/major Flutter bumps via the mise manager.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1ec50b78e4ffa8d2c0598bbfe562ffeebe51f1ea. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


.circleci/config.yml
```diff
@@ -335,15 +322,15 @@ jobs:
     steps:
       - checkout
       - replace-api-key
-      - revenuecat/install-mise-tools
+      - revenuecat/install-mise-tools:
+          tools: java,flutter
```
tonidero, line 326: Hmm so I was thinking Flutter has been known to do breaking changes in minors in the past :( That's why we wanted to have some tests running against the latest stable version, so we at least were aware of things breaking. That's why I'm unsure we want to lock to a specific version of flutter for all tests...

I think we should have tests against the latest stable + min flutter version we support, which I think we lose with these changes?

tonidero (commented):
Looks good but I'm unsure about locking the flutter version... We can discuss though of course!

.circleci/config.yml
AlvaroBrey: good to know, I'll give it some thought. First thing that comes to mind is that we could have Dependabot/Renovate bump mise Flutter versions

tonidero: Yup! That would also work indeed and be more "appropriate" 👍

renovate.json
```diff
@@ -1,4 +1,13 @@
...
+  "extends": ["github>RevenueCat/renovate"],
+  "packageRules": [
+    {
+      "matchManagers": ["mise"],
+      "matchDepNames": ["flutter"],
+      "matchUpdateTypes": ["minor", "major"],
+      "enabled": true,
+      "groupName": "flutter (mise)"
+    }
+  ]
```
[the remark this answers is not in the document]
AlvaroBrey: accepted. mise can update mise.lock too but it requires global renovate permissions which I don't want to grant, so this is the workaround

.circleci/config.yml
AlvaroBrey: I've added it to [Renovate](https://github.com/RevenueCat/purchases-flutter/pull/1854/commits/3eb3e1ee50855d3f7182117a75b1800f91e4adcb)

tonidero (approved):
Amazing! Thank you so much!!

PR #1765, opened 2026-05-29
Prepare both Android packages for AGP 9 built-in Kotlin

ajpallares:
### Motivation

Partially addresses #1759. AGP 9 turns Flutter's KGP warning into a build failure. Follows [Flutter's plugin author guide](https://docs.flutter.dev/release/breaking-changes/migrate-to-built-in-kotlin/for-plugin-authors).

### Summary

- `purchases_ui_flutter` applies `kotlin-android` only when no `kotlin` extension is already registered, so it works whether or not AGP supplies Kotlin itself.
- Main package drops KGP entirely; its Android sources are Java-only, so the #1759 warning no longer names it.
- New `agp9-smoke-test` job covers both `android.builtInKotlin` values, at configuration time only.

### Notes

`builtInKotlin=true` still fails, unchanged by this PR: Flutter force-applies `kotlin-android` to modules that don't declare it, and AGP 9 rejects that. First-party plugins like `shared_preferences` hit it too, so it needs an upstream fix. The default `builtInKotlin=false` works today, and AGP 8 behaviour is unchanged.

---

> [!NOTE]
> **Medium Risk**
> Changes Android Gradle/Kotlin wiring for a published plugin module; risk is mitigated by a dedicated AGP 9 smoke test, but consumer builds on AGP 9 still depend on correct conditional plugin application.
> 
> **Overview**
> Prepares both Flutter Android packages for **AGP 9** and Flutter’s built-in Kotlin migration, following the plugin-author guidance for avoiding duplicate `kotlin` extensions.
> 
> The **core** `purchases_flutter` Android library **drops Kotlin Gradle Plugin usage entirely** (no `kotlin-android`, KGP classpath, or `kotlin-stdlib`), since its native code is Java-only—removing the AGP 9 warning path for that module.
> 
> **`purchases_ui_flutter`** now **applies `kotlin-android` only when no `kotlin` extension exists** (after `com.android.library`), and sets **JVM 1.8** via either KGP `compilerOptions` or legacy `android.kotlinOptions`, depending on which Kotlin stack is active.
> 
> Adds a **Gradle-only AGP 9 smoke fixture** (`e2e-tests/agp9_smoke`) plus a CircleCI **`agp9-smoke-test`** job that runs both `android.builtInKotlin=true/false` at configuration time (AGP version, extension registration, whether KGP is applied, `jvmTarget`, and `compileDebugKotlin` dry-run). **Mise pins Gradle 9.1.0** for that job only.
> 
[...]

Issue #1759, opened 2026-05-22
Please migrate your plugin to Built-in Kotlin (for AGP9)


under3415:
> WARNING: Your app uses the following plugins that apply Kotlin Gradle Plugin (KGP): purchases_flutter
> Future versions of Flutter will fail to build if your app uses plugins that apply KGP.
> 
> If you are a plugin author, please migrate your plugin to Built-in Kotlin using this guide: https://docs.flutter.dev/release/breaking-changes/migrate-to-built-in-kotlin/for-plugin-authors

ajpallares:
Hi @under3415, thanks a lot for the report. It prompted a thorough investigation on our side!

The warning you're seeing is an informational one and **your builds will continue to work normally** on Flutter 3.44's default (`android.builtInKotlin=false`). We've decided to hold off on the migration for the moment, for two reasons:

1. **It's a wider Flutter ecosystem issue, not specific to `purchases_flutter`.** We tested the full Built-in Kotlin migration and found that under `android.builtInKotlin=true`, builds currently fail. Not just for our plugin, but also for Flutter's own first-party plugins like [`shared_preferences`](https://github.com/flutter/packages/blob/shared_preferences_android-v2.4.23/packages/shared_preferences/shared_preferences_android/android/build.gradle.kts#L32-L35). The Flutter Gradle plugin force-applies `kotlin-android` to every plugin subproject, and AGP 9 then rejects it. I suspect that's also why `flutter create` itself currently writes `android.builtInKotlin=false` in newly generated projects: the migration isn't ready end-to-end yet.

2. **The warning would persist anyway.** `purchases_ui_flutter` legitimately uses Kotlin and must still apply KGP under Flutter's current default, so this message is emitted as long as either of our packages is in your dependency graph, regardless of any change on our end.

We've prepared the migration locally (see [#1765](https://github.com/RevenueCat/purchases-flutter/pull/1765)) and will land it as soon as Flutter's upstream force-apply behavior is fixed and first-party plugins build cleanly under Built-in Kotlin. We'll update this issue when something changes on either front.

Thanks again for flagging it, and apologies for the noisy warning in the meantime!

CharlieMacnamara:
Hi team, I'm another user affected by this. Just noting that PR #1765 exists but appears to be marked DO NOT MERGE. It would be great to get this unblocked — my app is currently pinned to AGP 9.x and `android.builtInKotlin=false` because of this.

[...]

PR #1875, opened 2026-08-26
build: update Kotlin to 2.2.20 in purchases_ui_flutter

AlvaroBrey:
- Bumps `ext.kotlin_version` in `purchases_ui_flutter/android/build.gradle` from 1.9.20 to 2.2.20. One line.
- **Required before this plugin can take the next purchases-hybrid-common release** (RevenueCat/purchases-hybrid-common#1844), which publishes `kotlin-stdlib` with metadata 2.2.0. A Kotlin compiler reads metadata at most one minor ahead, so 1.9.20 cannot consume it.
- **`purchases_flutter` is deliberately untouched.** Its Android sources are Java only, so its `compileKotlin` task has no sources and never reads that metadata. Its stale `kotlin_version` is harmless.
- **App developers do not need to change their Kotlin version.** The plugin pins its own KGP in `buildscript`, it does not read the app's. They do need Gradle 7.6.3 or newer, which is KGP 2.2.20's minimum.
- Unrelated to the Built-in Kotlin work in #1765, which is blocked upstream on Flutter. This PR touches the same lines, so expect a conflict, but the two changes are compatible in substance.

### Checklist
- [x] A description about what and why you are contributing
- [x] The issue number(s) or PR number(s) in the description
- [ ] If applicable, unit tests

<details><summary>Agent description</summary>

### Motivation

purchases-android v11 moves to Kotlin 2.2.21, and purchases-hybrid-common follows in
RevenueCat/purchases-hybrid-common#1844. Both publish `kotlin-stdlib` at `compile` scope, so Gradle
resolves consumers onto it, and a compiler reads metadata from at most one minor version ahead of
itself. The published metadata becomes 2.2.0, which a 1.9.20 compiler rejects.

### Scope

Only `purchases_ui_flutter` needs the bump. Checked directly:

| Module | Kotlin sources | Java sources | Needs bump |
|---|---|---|---|
| `purchases_flutter/android` | 0 | 1 | no |
| `purchases_ui_flutter/android` | 9 | 0 | yes |

`javac` does not read `.kotlin_module` metadata, so a Java-only module is unaffected. Its
`kotlin-stdlib-jdk7` dependency resolves up to the newer version at runtime, which is fine.

### Consumer impact

This plugin pins its Kotlin Gradle Plugin version in its own `buildscript` block rather than reading
`rootProject.ext`, so an app's Kotlin version is irrelevant here and app developers are not forced to
change it. The one real requirement is Gradle 7.6.3 or newer, KGP 2.2.20's stated minimum.

### Relationship to #1765

#1765 restructures these same lines, moving the KGP classpath and `kotlinOptions` behind an
`applyLegacyKgp` conditional for AGP 9 built-in Kotlin. That work is held on a Flutter upstream
issue (#1759) and is a different problem from this one: it is about apps that want to run AGP 9,
whereas this is about the Kotlin metadata floor. `ext.kotlin_version` survives #1765's
restructuring, so whichever lands second just needs a textual conflict resolved.

### Testing

`flutter build apk --debug` on `revenuecat_examples/purchase_tester` builds successfully with the
bump applied, with `:purchases_ui_flutter:compileDebugKotlin` running under 2.2.20. Note the plugin's
`android/` directory has no Gradle wrapper of its own, so it can only be built through a host app.

### Related

- RevenueCat/purchases-hybrid-common#1844 is the upstream Kotlin bump that brings this floor here.
  It in turn follows the purchases-android v11 major.

</details>

### DIFF
--- a/purchases_ui_flutter/android/build.gradle
+++ b/purchases_ui_flutter/android/build.gradle
@@ -2,7 +2,7 @@ group 'com.revenuecat.purchases_ui_flutter'
 version '10.10.0'
 
 buildscript {
-    ext.kotlin_version = '1.9.20'
+    ext.kotlin_version = '2.2.21'
     ext.common_version = '18.32.1'
     repositories {
         google()
--- a/revenuecat_examples/purchase_tester/android/gradle.properties
+++ b/revenuecat_examples/purchase_tester/android/gradle.properties
@@ -4,3 +4,7 @@ android.enableJetifier=true
 android.defaults.buildfeatures.buildconfig=true
 android.nonTransitiveRClass=false
 android.nonFinalResIds=false
+# This builtInKotlin flag was added automatically by Flutter migrator
+android.builtInKotlin=false
+# This newDsl flag was added automatically by Flutter migrator
+android.newDsl=false


AlvaroBrey:
Closing: this change is a no-op.

Verified against a purchases-hybrid-common built with Kotlin 2.2.21 and published locally under a unique version, so resolution was provable:

- app declares Kotlin 1.8.22, plugin at 2.2.21 -> FAILS (`:purchases_ui_flutter:compileDebugKotlin`, metadata 2.2.0 unreadable)
- app declares Kotlin 2.1.21, plugin left at its original 1.9.20 -> BUILDS

So `ext.kotlin_version` in `purchases_ui_flutter/android/build.gradle` does not control the compiler in an app context. Flutter's plugin loader applies `org.jetbrains.kotlin.android` from the app's own `settings.gradle` to plugin subprojects, and that is what decides it.

The real requirement is on the app side: a Flutter app consuming the next PHC release needs Kotlin 2.1.0 or newer in its `android/settings.gradle`. That is a release-notes and migration-guide item, not a code change here. Flutter's own app template already ships 2.3.20, so only older apps are affected, and they can bump Kotlin without changing their Flutter version.

PR #1876, opened 2026-08-27
docs: state the minimum Kotlin version

AlvaroBrey:
Docs only. Adds the minimum Kotlin version that arrives with the next purchases-hybrid-common major.

- purchases-android v11 publishes `kotlin-stdlib` 2.2.21 at compile scope, so its metadata is 2.2.0. A Kotlin compiler reads metadata at most one minor ahead, so an app on a compiler older than 2.1 fails to build.
- Flutter applies the version declared in the **app's** `android/settings.gradle` to plugin subprojects, so the requirement lands on the consuming app. Verified: app at 1.8.22 fails, app at 2.1.21 builds, with the plugin's own `ext.kotlin_version` unchanged in both.
- Flutter's own app template already ships Kotlin 2.3.20, so only existing projects need the bump.

Depends on RevenueCat/purchases-hybrid-common#1844, which propagates the floor. Merge after the purchases-hybrid-common major that carries purchases-android v11.

### Checklist
- [ ] If applicable, unit tests were added
- [ ] If applicable, followup issues for other SDKs were created

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@
 To use this plugin, add `purchases_flutter` as a [dependency in your pubspec.yaml file](https://flutter.io/platform-plugins/).
 
 ### Requirements
-*purchases_flutter* requires Xcode 14.0+ and minimum targets iOS 13.0+/Android SDK 21+ (Android 5.0+).
+*purchases_flutter* requires Xcode 14.0+, Kotlin 2.1.0+, and minimum targets iOS 13.0+/Android SDK 21+ (Android 5.0+).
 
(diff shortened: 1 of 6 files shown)


tonidero (approved):
Looks like there are still some tests failing, but in any case this PR looks good! we can tackle those failures in follow-ups

AlvaroBrey:
> Looks like there are still some tests failing, but in any case this PR looks good! we can tackle those failures in follow-ups

fixed! raised compileSdk

PR #1897, opened 2026-09-14
feat(paywalls): Add onInteraction callback to PaywallView

AlvaroBrey:
- Adds `onInteraction` to `PaywallView`, called with the `paywall_component_interacted` event as a `Map<String, dynamic>` of snake_case keys; keys that do not apply are absent.
- The native views forward the event from `PurchasesHybridCommonUI` 19.0.0: the Kotlin `PaywallListenerWrapper` override and the Swift `didTrackInteraction` delegate method.
- `PaywallFooterView` gets the now mandatory Kotlin override as a no-op: the footer never exposed this callback in Dart or Swift, and interactions originate in V2 paywalls, which the footer does not render.
- Stacked on the 19.0.0 bump, #1896, which does not compile on its own because the wrapper method is abstract.
- The API tester also gains `onWebCheckoutOpened` and `onUrlOpened`, which had been missing.
- Part of [SDK-4459](https://linear.app/revenuecat/issue/SDK-4459).

### Checklist
- [x] A description about what and why you are contributing, even if it's trivial.
- [x] The issue number(s) or PR number(s) in the description if you are contributing in response to those.
- [x] If applicable, unit tests.
- [ ] Follow-up for `presentPaywall*()` listener support (separate PR).

<details><summary>Agent description</summary>

### Motivation

The native SDKs expose paywall interactions so apps can route them to their own analytics tools. `purchases-hybrid-common` 19.0.0 bridges the callback; this PR surfaces it on the Flutter `PaywallView`, following the shape of `onUrlOpened`.

### Description

- `purchases_ui_flutter/android/.../views/PaywallView.kt`: `onInteraction(event: Map<String, Any>)` invokes `onInteraction` on the view's method channel with the map as arguments.
- `purchases_ui_flutter/android/.../views/PaywallFooterView.kt`: empty override, required by the abstract method.
- `purchases_ui_flutter/ios/.../PurchasesUiPaywallView.swift`: `paywallViewController(_:didTrackInteraction:)` invokes `onInteraction` on the channel with the dictionary.
- `lib/views/paywall_view.dart`: `onInteraction` named optional, documented next to its siblings, passed to `PaywallViewMethodHandler`.
- `lib/views/paywall_view_method_handler.dart`: `onInteraction` named optional and a `case 'onInteraction'` that copies the arguments into a `Map<String, dynamic>`.
[...]

PR #1898, opened 2026-09-15
feat(paywalls): Add callbacks to presentPaywall and presentPaywallIfNeeded

AlvaroBrey:
> Parked pending RevenueCat/purchases-hybrid-common#1911, which adds a per-presentation delegate to `PaywallProxy`. That replaces the shared-delegate bookkeeping described below and lets each presentation's callbacks receive only its own events.

- Adds the `PaywallView` callbacks as named optional parameters of `presentPaywall()` and `presentPaywallIfNeeded()`, so modally presented paywalls report the same events as the embedded view. No `onDismiss`: the returned `Future<PaywallResult>` completes on dismissal.
- Callbacks go on a dedicated `com.revenuecat.purchasesui/PresentedPaywall` channel handled by the existing `PaywallViewMethodHandler`. The plugin channel's `onRestore*` and `onDismiss` method names already belong to the customer center handler.
- Native attaches a listener only when a callback is passed (`hasListener`).
- Also fixes `onPurchaseCompleted` never firing on iOS when a purchase completes with no StoreKit transaction: the handler parsed the absent transaction and threw. It affected `PaywallView` and the footer view already, since all four share `PaywallViewMethodHandler`; the callback now receives an empty `StoreTransaction`.
- Part of [SDK-4498](https://linear.app/revenuecat/issue/SDK-4498).

### Checklist
- [x] A description about what and why you are contributing, even if it's trivial.
- [x] The issue number(s) or PR number(s) in the description if you are contributing in response to those.
- [x] If applicable, unit tests.

<details><summary>Agent description</summary>

### Motivation

`PaywallView` gained `onInteraction` in #1897, but apps that present paywalls modally still only get the final `PaywallResult`. They cannot route interactions to analytics or react to purchase events while the paywall is up. Both natives already have the hook unused: `PresentPaywallOptions.paywallListener` on Android and `PaywallProxy.delegate` on iOS.

### Description

- `purchases_ui_flutter.dart`: both present methods build a `PaywallViewMethodHandler` when any callback is given, add it to the presented-paywall channel's active handlers, send `'hasListener': true`, and remove it in a `finally` once the result arrives. The channel fans each call out to every active handler, so a presentation that resolves early (`presentPaywallIfNeeded` returning `NOT_PRESENTED`) does not silence one still on screen. Without callbacks the argument map is unchanged.
- Android `PurchasesUiFlutterPlugin.kt`: a second `MethodChannel` and a `PaywallListenerWrapper` passed as `paywallListener` when `hasListener`. The wrapper bodies moved to a shared `forwardingPaywallListener(channel)` that `views/PaywallView.kt` now uses too, so both surfaces send the same method names. `onPurchasePackageInitiated` is left to PHC's default, which resumes the purchase.
- iOS `PurchasesUiFlutterPlugin.swift`: `PresentedPaywallDelegateForwarder` mirrors the `PurchasesUiPaywallView` delegate extension on a second `FlutterMethodChannel`. One instance is set as `paywallProxy.delegate` while any presentation with callbacks is active, and dropped when the last one finishes. It does not implement the purchase-initiated delegate method, so `PaywallProxy` resumes automatically. macOS is untouched (same file, `#if os(iOS)`).
- `api_tester`: `_checkPresentPaywallWithListeners` covers every callback on both methods.

### Regression gates
- `paywall_view_method_handler_test.dart`: `onPurchaseCompleted fires when there is no store transaction`, which throws `type 'Null' is not a subtype of type 'Map<dynamic, dynamic>'` without the handler guard.

- `presentPaywall callbacks › delivers callbacks while presented and stops after the result`: a call on the presented-paywall channel reaches the callback while the native result is pending, and no longer after it resolves.
- `presentPaywall callbacks › sends hasListener only when a callback is given`; the existing exact-match argument tests prove the key is absent otherwise.

**Rejected:**
- Reusing the `purchases_ui_flutter` channel with prefixed method names: duplicates the parsing already in `PaywallViewMethodHandler` and keeps the customer center switch growing.
- A `PaywallListener` object parameter: diverges from both `presentCustomerCenter()` and `PaywallView`, which take individual callbacks.
- Always attaching the native listener: changes process-death behavior for every caller, not only those that opted into callbacks.

</details>

---

> [!NOTE]
> **Medium Risk**
> Touches purchase/restore event plumbing on both platforms and concurrent modal presentation; behavior is opt-in via callbacks but incorrect listener lifetime could mis-route or drop events.
> 
> **Overview**
> **Modal paywall presentation** now accepts the same optional lifecycle callbacks as embedded `PaywallView` (purchase start/complete/cancel/error, restore, web checkout, URL opens, and `onInteraction`). Dismissal is still only via the returned `Future<PaywallResult>`—there is no `onDismiss`.
> 
> Dart registers handlers on a dedicated `com.revenuecat.purchasesui/PresentedPaywall` channel for the duration of the native call, sends `hasListener` only when any callback is set, and fans events to all active presentations so an early `NOT_PRESENTED` does not block callbacks for a paywall still on screen. **Android** wires `PresentPaywallOptions.paywallListener` through a shared `forwardingPaywallListener`; **iOS** adds `PresentedPaywallDelegateForwarder` with reference counting when multiple modal presentations use callbacks.
> 
> **Bugfix:** `PaywallViewMethodHandler` no longer crashes on `onPurchaseCompleted` when `storeTransaction` is missing (e.g. some iOS completions); callers get an empty `StoreTransaction` instead.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 84db1411c2cfb8a452ed20eb58fce1d04bbcdaa4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


AlvaroBrey:
@cursor review

